PR #6, opened 2026-05-24
fix: support arbitrary version segments in openai-compat base URLs

lucasliet:
## Motivation

The `openai-compat` provider normalizes the configured base URL by stripping
a trailing `/v1` and then always re-appending `/v1/chat/completions` (or
`/v1/embeddings`). This works for the common cases (OpenAI, OpenRouter,
Ollama) but breaks providers that use a different version segment in their
path, such as **Z.AI** (`/v4`) or any provider that already includes the
full endpoint URL in the configured base.

## Before

`normalize_openai_base` stripped `/v1` and the caller unconditionally
re-inserted it:

```rust
pub fn normalize_openai_base(url: &str) -> String {
    let s = url.trim_end_matches('/');
    s.strip_suffix("/v1").unwrap_or(s).to_string()
}
```

With `AI_MEMORY_LLM_BASE_URL=https://api.z.ai/api/coding/paas/v4` the
resulting request URL was:

```
https://api.z.ai/api/coding/paas/v4/v1/chat/completions  ← 404
```

## After

`normalize_openai_base` now receives the target endpoint and builds the
full URL in one step, covering three cases in priority order:

1. URL already ends with `/<endpoint>` (full endpoint URL supplied) → used as-is
2. Last path segment is a version token `v\d{1,2}` (e.g. `v1`, `v4`) → append `/<endpoint>` directly
3. Bare host with no version segment → append `/v1/<endpoint>` (OpenAI default)

```rust
pub fn normalize_openai_base(base: &str, endpoint: &str) -> String { … }
```

| Configured base URL | Before | After |
|---|---|---|
| `https://api.openai.com` | `.../v1/chat/completions` ✓ | `.../v1/chat/completions` ✓ |
| `https://openrouter.ai/api/v1` | `.../v1/chat/completions` ✓ | `.../v1/chat/completions` ✓ |
| `http://localhost:11434/v1` | `.../v1/chat/completions` ✓ | `.../v1/chat/completions` ✓ |
| `https://api.z.ai/api/coding/paas/v4` | `.../v4/v1/chat/completions` ✗ | `.../v4/chat/completions` ✓ |
| `https://api.githubcopilot.com/chat/completions` | `.../chat/completions/v1/chat/completions` ✗ | `.../chat/completions` ✓ |

The same fix applies to the embedding providers (`OpenAiEmbedder`,
`VoyageEmbedder`) which shared the same pattern.

## Test plan

- Existing unit tests for `normalize_openai_base` updated to assert the
  full URL (not just the normalized base), covering all three branches
- New cases added for Z.AI-style `/v4` path and full endpoint URLs
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` all pass

### DIFF
--- a/crates/ai-memory-llm/src/embedding.rs
+++ b/crates/ai-memory-llm/src/embedding.rs
@@ -114,7 +114,7 @@ impl Embedder for OpenAiEmbedder {
     }
 
     async fn embed(&self, text: &str) -> LlmResult<Vec<f32>> {
-        let url = format!("{}/v1/embeddings", normalize_openai_base(&self.base_url));
+        let url = normalize_openai_base(&self.base_url, "embeddings");
         debug!(url, model = %self.model, "POST openai/embeddings");
         let req = OpenAiEmbeddingRequest {
             input: text,
@@ -222,7 +222,7 @@ impl Embedder for VoyageEmbedder {
     }
 
     async fn embed(&self, text: &str) -> LlmResult<Vec<f32>> {
-        let url = format!("{}/v1/embeddings", normalize_openai_base(&self.base_url));
+        let url = normalize_openai_base(&self.base_url, "embeddings");
         let req = VoyageRequest {
             input: [text],
             model: &self.model,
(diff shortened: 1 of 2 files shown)


akitaonrails:
Merged in 87714da and shipped in **v0.1.3** (`akitaonrails/ai-memory:0.1.3`). Clean fix with thorough tests — the version-segment detection + full-endpoint passthrough is exactly right, and it unblocks Z.AI/`/v4` and any provider that hands over a complete endpoint URL. Thanks for the contribution, @lucasliet! 🙏

PR #7, opened 2026-05-24
Add frontend API and optional custom web UI directory

djalmajr:
## Summary

Adds a read-only HTTP/JSON surface and an optional custom web UI so external frontends can browse an `ai-memory` instance, on the same origin and behind the existing auth/allowed-host middleware. It also extends the MCP read tools and the store with the queries those surfaces need, plus a few self-contained fixes. No write/admin operations are exposed under `/api/v1`.

## What's included

**Read-only `/api/v1`** (mounted before the bearer layer, so it inherits auth):
- workspaces, projects, pages, page (markdown + frontmatter), recent, briefing
- search: global, a single workspace+project, or multi-scope (POST); a partial scope (`workspace` without `project`, or vice versa) returns 400 instead of silently broadening to global
- per-workspace and per-project `overview`: open handoff + briefing + memory-health in one call
- enriched page reads: resolved outgoing links and incoming back-links
- memory-health drill-down: the actual stale / duplicate / orphan pages, not just counts
- project-scoped routes return 404 when the workspace/project is missing

**Custom web UI**:
- `--web-ui-dir` serves a static SPA at `/web` with SPA fallback to `index.html`; without it, the built-in server-side browser stays the default
- `--web-ui-dir` requires `--enable-web`, and the directory must exist and contain `index.html`

**MCP read tools**:
- optional `workspace` and multi-`scopes` on the read tools (query/recent/status/briefing/explore); existing single-project behavior is unchanged (`#[serde(default)]`)
- `memory_query` uses hybrid search when an embedder is configured and degrades to BM25 if the embedder fails

**Store / CLI / misc** (self-contained, could land separately):
- page `kind` derived from path (`_rules/`→rule, `decisions/`→decision, `gotchas/`→gotcha) when frontmatter `kind` is absent; `write-page --kind` to set it explicitly
- `openai-compat` `base_url` falls back to `LLM_BASE_URL`, mirroring how the API key is sourced from env

## Validation

- `TAILWIND_SKIP=1 cargo test --workspace` — `ai-memory-cli` (153), `ai-memory-store` (37), `ai-memory-web` routes (27)

Builds use `TAILWIND_SKIP=1` to skip the `build.rs` Tailwind step in headless/CI environments. Two pre-existing items, both present on `main` and green in CI: `cargo deny` may flag transitive advisories (`syntect` → `bincode`/`yaml-rust`), and the three `commands::reset::tests` are environment-specific (they touch real data dirs and fail in some local sandboxes).

## Reference UI

A working frontend built against the `--web-ui-dir` + `/api/v1` contract: **[djalmajr/ai-memory-ui](https://github.com/djalmajr/ai-memory-ui)** — a SolidJS SPA with workspace/project overview (handoff + briefing + memory-health), a document reader with references/back-links, command-palette multi-scope search, and i18n (en/pt-BR/es). It ships Playwright e2e tests and screenshots, and exercises this PR end-to-end against `ai-memory serve --enable-web --web-ui-dir <dist>`.

## Deploying

A working self-host packaging of this branch lives at **[djalmajr/ai-memory-ops](https://github.com/djalmajr/ai-memory-ops)** — a Helm chart plus the supporting images (the engine with the custom web UI baked in, an OIDC `mcp-auth` gateway, and a git→wiki ETL). It builds the `ai-memory` image from this branch and the SPA from `ai-memory-ui`, and runs them behind Keycloak/OIDC on Kubernetes.


akitaonrails:
Linking the design discussion from #21 here. I am closing #21 to avoid splitting the API-shape conversation across an issue and this draft PR; please keep review/design feedback on this PR while it remains draft.

djalmajr:
Synced `feature/custom-web-ui` with `upstream/main` (now at v0.5.1) — merge commit b4b3cb2. Zero conflicts; `cargo build --workspace --all-features` clean; `cargo test --workspace` passes (559/559, modulo 4 environment-dependent failures in `commands::reset::tests` + `uninstall_purge_data_apply_wipes` that abort when `ai-memory` processes are running on the host — pre-existing defensive behavior, not regressions).

Ready for fresh review when you have cycles 🙏

akitaonrails:
Merging 🎉 — thank you for the careful, well-scoped work, @djalmajr. This is a substantial contribution and I verified the high-stakes parts against the actual head before merging, not just the description:

- **Auth ordering** — `/api/v1` and `/web` are registered before `apply_http_layers`, so axum 0.8's layer semantics correctly route them through `require_bearer` + host allowlist. Anonymous → 401, exactly as it should.
- **Read-only enforcement** — zero `writer.` / `wiki.write` / `insert_*` / `delete_*` / `consolidate` calls in `api.rs`. Pure read surface.
- **Backward compatibility** — `workspace` + `scopes` args have `#[serde(default)]`; existing single-`project` callers unchanged. The new docstrings ("Omit to use the current/default workspace resolution chain") align cleanly with v0.5.2's "default to current project" rule.
- **Limits properly bounded** — every endpoint clamps `limit` to `(1, 100)`, `MAX_SEARCH_SCOPES = 25` — no `limit=1000000` abuse vector.
- **Static-files security** — `tower-http::ServeDir` rejects `..`, and `validate_web_ui_args` checks `is_dir()` + `index.html` exists *before* the server starts.
- **Build + gate** (verified myself in a detached worktree at 097bba5): fmt clean, clippy clean, **577 tests pass** (+31 net new, 1 ignored — pre-existing live-process guard, unrelated).
- **31 new tests** are real, not hollow — auth gating, 404s on missing scope, 400 on partial scope, multi-scope correctness, SPA fallback, page back-links, overview aggregation.

Two **non-blocking** follow-ups I'd open as separate issues (not asking you to do them, just so we don't lose them):

1. **`Cache-Control` + ETags on idempotent reads.** Right now a polling SPA hits the DB on every `/api/v1/overview` request. Adding `Cache-Control: private, max-age=N` (different N per endpoint) and ETags on page bodies is a cheap iteration once we see real polling patterns.
2. **CORS layer (+ write surface) for off-origin frontends.** Same-origin is the right v1 stance — host the SPA via `--web-ui-dir`. If/when someone wants a separately-hosted frontend, a `tower-http::cors::CorsLayer` gated by an `--cors-allow-origin` flag (and eventually a thin write surface so the UI can do more than browse) is the natural v2.

Thanks again — this is exactly the shape we needed: clean read-only contract, zero write-path risk, opt-in custom UI, no scope creep. 🙏

PR #8, opened 2026-05-24
fix(openai): unblock LLM-driven features (additionalProperties, max_completion_tokens, dry-run)

pablowinck:
Three independent bugs surfaced while trying to `ai-memory bootstrap` against the OpenAI provider on `:latest` (v0.1.3). All three blocked every LLM-driven feature (bootstrap, consolidate, lint, explore) until worked around. This PR fixes them. Each commit is self-contained and reviewable on its own.

### 1. `response_format` schema missing `additionalProperties: false` → OpenAI 400

```
Invalid schema for response_format 'Result': In context=(), 'additionalProperties'
is required to be supplied and to be false.
```

`complete_structured_raw` passed the caller's schema verbatim to OpenAI Structured Outputs with `strict: true`, but OpenAI requires every object node to set `additionalProperties: false`. Every consolidation/lint/bootstrap call hit this.

**Fix:** walk the schema before sending and inject `additionalProperties: false` on every object node that doesn't already set one. Caller-provided values (e.g. someone deliberately allowing extras) are preserved.

### 2. `max_tokens` rejected by gpt-5 / o-series → OpenAI 400

```
Unsupported parameter: 'max_tokens' is not supported with this model.
Use 'max_completion_tokens' instead.
```

The rename ships with gpt-5 and the o-series reasoning models (`o1*`, `o3*`, `o4*`). Sending the legacy field fails closed.

**Fix:** detect by model-id prefix and emit `max_completion_tokens` instead. gpt-4* / gpt-3.5* / openai-compat providers continue to receive `max_tokens` unchanged.

### 3. `ai-memory bootstrap --dry-run` still POSTs the full bundle → 413 on large repos

```
Error: server returned 413 Payload Too Large: Failed to buffer the request body
```

Previously, `--dry-run` collected sources locally and POSTed the entire bundle (often many MB) just so the server could short-circuit before the LLM call. Two problems:

1. Defeated the purpose of "dry run" as a no-network-side-effect rehearsal — a server still had to be running and reachable for the flag to do anything.
2. Repos large enough to exceed `MAX_BODY_BYTES` (10 MB, hardcoded in `serve.rs:37`) failed with 413 even when the user explicitly asked for a preview.

**Fix:** when `--dry-run` is set, compute the summary client-side by reusing the same `prune_to_budget` + `SourceCounts::from_sources` calls the server would have made, then print the human + machine-readable report without any HTTP request. Promotes `prune_to_budget` from crate-private to `pub` (re-exported from the consolidate crate root) so the CLI can reach it. Server behaviour is unchanged: `POST /admin/bootstrap` still honours `dry_run` in the payload, so older CLIs continue to work exactly as before.

## Testing

All 9 new unit tests pass (6 in `openai.rs`, 3 in `commands/bootstrap.rs`):

```
test openai::tests::enforce_strict_injects_additional_properties_false_on_root ... ok
test openai::tests::enforce_strict_recurses_into_nested_objects ... ok
test openai::tests::enforce_strict_preserves_existing_additional_properties ... ok
test openai::tests::enforce_strict_ignores_non_object_nodes ... ok
test openai::tests::model_requires_max_completion_tokens_matches_gpt5_and_o_series ... ok
test openai::tests::model_requires_max_completion_tokens_passes_gpt4_through ... ok
test openai::tests::build_request_uses_max_tokens_for_gpt4 ... ok
test openai::tests::build_request_uses_max_completion_tokens_for_gpt5 ... ok
test commands::bootstrap::tests::local_dry_run_marks_outcome_as_dry_run ... ok
test commands::bootstrap::tests::local_dry_run_tallies_kinds ... ok
test commands::bootstrap::tests::local_dry_run_drops_to_fit_tight_budget ... ok
```

Gates locally:

- `cargo clippy --workspace --all-targets -- -D warnings` ✓ (with `TAILWIND_SKIP=1` because of the `linux/arm64` SHA-pinning issue on my Apple Silicon — unrelated to this PR)
- `cargo test --workspace` ✓ (one pre-existing flaky `picks_up_externally_created_file` in `ai-memory-wiki::watcher`, also unrelated)

Real-call OpenAI smoke test of bootstrap is left for CI / your local validation — I can verify against `gpt-4o-mini` if useful.

## Compatibility

- No schema changes.
- No CLI flag changes.
- No server endpoint changes.
- `prune_to_budget` is the only previously-private item promoted to `pub` (additive).

Semver-wise this is a **patch** release: pure bug fixes, additive surface.


pablowinck:
Superseded by #10 — that one was rebased on top of the latest `main` and extended with bugs 4 and 5 surfaced during end-to-end smoke testing.

PR #10, opened 2026-05-24
fix(openai): unblock LLM features (5 bugs in schema, max_tokens, dry-run)

pablowinck:
Five independent bugs surfaced while running `ai-memory bootstrap` against the OpenAI provider on `:latest` (v0.1.3). All five blocked LLM-driven features (bootstrap, consolidate, lint, explore). This PR fixes them and was validated end-to-end against `gpt-4o-mini`.

(Supersedes the closed #8 — that branch was rebased + extended with bugs 4 and 5 surfaced during the smoke-test loop.)

Each commit is self-contained and reviewable on its own.

### 1. `response_format` schema missing `additionalProperties: false` → OpenAI 400

```
Invalid schema for response_format 'Result': In context=(), 'additionalProperties'
is required to be supplied and to be false.
```

`complete_structured_raw` passed the caller's schema verbatim to OpenAI Structured Outputs with `strict: true`, but OpenAI requires every object node to set `additionalProperties: false`.

**Fix:** walk the schema before sending and inject `additionalProperties: false` on every object node that doesn't already set one.

### 2. `required` must list every property key — partial arrays rejected

```
Invalid schema for response_format 'Result': In context=(), 'required' is required
to be supplied and to be an array including every key in properties. Missing 'tags'.
```

`schemars` emits an explicit but partial `required` for structs with `#[serde(default)]` fields (e.g. `BootstrapPage.tags`). OpenAI strict mode rejects partial arrays — optionality must be expressed at the value site (nullable union types like `["string", "null"]`).

**Fix:** overwrite `required` unconditionally with every key from `properties` whenever `properties` is present.

### 3. `max_tokens` rejected by gpt-5 / o-series → OpenAI 400

```
Unsupported parameter: 'max_tokens' is not supported with this model.
Use 'max_completion_tokens' instead.
```

The rename ships with gpt-5 and the o-series reasoning models (`o1*`, `o3*`, `o4*`).

**Fix:** detect by model-id prefix and emit `max_completion_tokens`. gpt-4* / gpt-3.5* / openai-compat providers keep `max_tokens`.

### 4. `max_tokens=64000` (bootstrap default) exceeds gpt-4o family cap

```
max_tokens is too large: 64000. This model supports at most 16384 completion tokens,
whereas you provided 64000.
```

Bootstrap deliberately asks for 64K so Anthropic Haiku-class models don't truncate mid-JSON. gpt-4o-mini caps at 16384 and rejects above.

**Fix:** client-side per-family cap. gpt-4o / gpt-4-turbo / gpt-3.5 → 16384 (their published cap). Reasoning models (gpt-5 / o-series) → uncapped (documented at 128K+). Caller continues to express its intent (generous budget); provider quietly honours the model-specific limit.

### 5. `ai-memory bootstrap --dry-run` still POSTs the full bundle → 413 on large repos

```
Error: server returned 413 Payload Too Large: Failed to buffer the request body
```

The CLI collected sources locally and POSTed the entire bundle just so the server could short-circuit before the LLM call. Defeated the purpose of `--dry-run` and exploded on repos exceeding `MAX_BODY_BYTES = 10 MB` (hardcoded in `serve.rs:37`).

**Fix:** when `--dry-run` is set, compute the summary client-side by reusing the same `prune_sources_to_budget` + `SourceCounts::from_sources` calls the server would have made, then print the report without any HTTP request. Server behaviour is unchanged — older CLIs continue to work exactly as before.

## End-to-end validation

Built the image locally, ran the full bootstrap pipeline against `gpt-4o-mini`:

```
$ cd <small-repo> && ai-memory bootstrap --since "60 days ago"
…
Generated 3 wiki pages:
  - decisions/0001-revert-hml-build-to-single-arch.md
  - decisions/0002-change-image-pull-policy.md
  - bootstrap.md

Rationale: Processed two significant decisions from the git log regarding CI
configurations and Kubernetes image policies, creating structured wiki pages
to document these changes.
```

Page bodies and frontmatter are well-formed (LLM-tagged, ADR-shaped, semantic tier).

## Test coverage

13 new unit tests across the affected modules:

- `openai.rs`: schema enforcement (4 cases: root injection, recursion, preserve existing, ignore non-objects), `required` completion (2 cases: fill when absent, overwrite when partial), model family routing (2 cases: gpt-5 family, gpt-4 family), max_tokens cap (2 cases: gpt-4o capped, gpt-5 uncapped), build_request integration (2 cases: max_tokens for gpt-4, max_completion_tokens for gpt-5).
- `commands/bootstrap.rs`: dry-run output (3 cases: empty, kind tally, budget pruning).

## Gates

Locally green:

- `cargo clippy --workspace --all-targets -- -D warnings` ✓ (with `TAILWIND_SKIP=1` because `linux/arm64` lacks a pinned Tailwind SHA in `build.rs` — unrelated to this PR)
- `cargo test --workspace` ✓ (one pre-existing flaky `picks_up_externally_created_file` in `ai-memory-wiki::watcher`, also unrelated)

## Compatibility

- No schema migration.
- No CLI flag changes.
- No server endpoint changes.
- `prune_sources_to_budget` is the only previously-private item promoted to `pub` (additive).
- Hooks / installed agents unaffected.

Semver: **patch** — pure bug fixes, additive surface only.

### DIFF
--- a/crates/ai-memory-cli/src/commands/bootstrap.rs
+++ b/crates/ai-memory-cli/src/commands/bootstrap.rs
@@ -93,7 +93,13 @@ pub async fn run(config: &Config, args: BootstrapArgs) -> Result<()> {
     // local preview and (b) exploded with 413 on repos large enough
     // to exceed the server's 10 MB body limit. We now compute the
     // dry-run summary entirely client-side and skip the round-trip.
+    //
+    // Server parity: `process_sources` returns `NoSources` when the
+    // collected bundle is empty, so the operator catches a wrong
+    // repo path or over-broad `--exclude-*` flag before they think
+    // they bootstrapped something. Mirror that here.
     if args.dry_run {
+        ensure_sources_for_dry_run(&sources)?;
         let outcome = local_dry_run(sources, args.max_input_tokens);
         print_human_report(&outcome, &args.workspace, &project);
         let report = serde_json::to_string_pretty(&outcome)?;
@@ -115,6 +121,23 @@ pub async fn run(config: &Config, args: BootstrapArgs) -> Result<()> {
     print_human_report(&outcome, &args.workspace, &project);
     let report = serde_json::to_string_pretty(&outcome)?;
     println!("\n--- machine-readable ---\n{report}");
+    Ok(())
+}
+
+/// Mirror the server's `process_sources` `NoSources` check so a local
+/// dry-run on an empty bundle fails the same way a real bootstrap
+/// would. Catches the operator who pointed at a wrong path, a folder
+/// without a git repo / README / docs, or who combined `--exclude-*`
+/// flags so aggressively that every source dropped — before they
+/// think they bootstrapped something and waste tokens against a real
+/// run later.
+fn ensure_sources_for_dry_run(sources: &[BootstrapSource]) -> Result<()> {
+    anyhow::ensure!(
+        !sources.is_empty(),
+        "no input sources collected (matches server `NoSources`): wrong repo path, or every \
+         --exclude-* flag combined to drop everything. Remove at least one --exclude-* flag \
+         or point --repo-path at a directory with a README, docs/, or commits.",
+    );
     Ok(())
 }
 
@@ -234,7 +257,7 @@ fn print_human_report(outcome: &BootstrapOutcome, workspace: &str, project: &str
 
 #[cfg(test)]
 mod tests {
-    use super::local_dry_run;
+    use super::{ensure_sources_for_dry_run, local_dry_run};
     use ai_memory_consolidate::{BootstrapSource, SourceKind};
 
     fn source(kind: SourceKind, label: &str, body_len: usize) -> BootstrapSource {
@@ -247,10 +270,13 @@ mod tests {
 
     #[test]
     fn local_dry_run_marks_outcome_as_dry_run() {
-        let outcome = local_dry_run(vec![], 150_000);
+        // The CLI handler refuses to call `local_dry_run` with zero
+        // sources (server parity — see `anyhow::ensure!` in `run`),
+        // so the lowest input we can hand the helper is one source.
+        let outcome = local_dry_run(vec![source(SourceKind::Readme, "README", 50)], 150_000);
         assert!(outcome.dry_run);
-        assert_eq!(outcome.sources_collected, 0);
-        assert_eq!(outcome.sources_sent, 0);
+        assert_eq!(outcome.sources_collected, 1);
+        assert_eq!(outcome.sources_sent, 1);
         assert!(outcome.pages_written.is_empty());
         assert!(outcome.rationale.contains("dry-run"));
     }
@@ -271,6 +297,29 @@ mod tests {
         assert_eq!(outcome.sources_by_kind.git_commits, 2);
         assert_eq!(outcome.sources_by_kind.doc_files, 1);
         assert!(outcome.estimated_input_tokens > 0);
+    }
+
+    #[test]
+    fn ensure_sources_for_dry_run_rejects_empty() {
+        // Server's `process_sources` returns `BootstrapError::NoSources`
+        // on an empty bundle (status 422). The client-side dry-run path
+        // must do the same so wrong repo paths / over-broad excludes
+        // don't get silently glossed over.
+        let err = ensure_sources_for_dry_run(&[])
+            .expect_err("empty bundle must fail in dry-run, matching server NoSources");
+        let msg = format!("{err}");
+        assert!(
+            msg.contains("no input sources collected") && msg.contains("NoSources"),
+            "error message should reference both the user-facing cause and the server parity: {msg}",
+        );
+    }
+
+    #[test]
+    fn ensure_sources_for_dry_run_accepts_non_empty() {
+        // A single source is enough — paridade with the server's check,
+        // which gates on emptiness alone (budget pruning happens later).
+        ensure_sources_for_dry_run(&[source(SourceKind::Readme, "README", 50)])
+            .expect("non-empty bundle must pass");
     }
 
     #[test]
--- a/crates/ai-memory-llm/src/openai.rs
+++ b/crates/ai-memory-llm/src/openai.rs
@@ -47,16 +47,44 @@ fn last_segment_is_version(url: &str) -> bool {
     })
 }
 
+/// Request dialect — picks which OpenAI quirks the provider applies.
+///
+/// `Official` targets `api.openai.com` and honours the model-family
+/// rules that the real OpenAI Chat Completions endpoint enforces:
+/// `max_completion_tokens` for gpt-5 / o-series, model-family output
+/// caps, omitted `temperature` for reasoning models, strict-mode JSON
+/// schema normalisation.
+///
+/// `Compat` targets the OpenAI-compatible wire format spoken by
+/// Ollama, vLLM, LM Studio, llama.cpp, and the long tail of local /
+/// proxy backends. Those backends almost universally implement the
+/// legacy `max_tokens` dialect, ignore OpenAI-specific output caps,
+/// and accept any temperature value — so we keep the request shape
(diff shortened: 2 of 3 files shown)


akitaonrails (changes requested):
Thanks for the thorough work here. The direction is good and the PR fixes real OpenAI Structured Outputs bugs, but I do not think we should merge it as-is because a few compatibility regressions are either introduced or locked in by tests.

Blocking changes requested:

1. Force `additionalProperties: false` for OpenAI strict schemas.

`enforce_strict_object_schemas` currently preserves an explicit `additionalProperties: true`, and the new test locks that in. That conflicts with the PR rationale and OpenAI strict-mode requirements. For this OpenAI-specific normalization path, please always force object nodes to `additionalProperties: false` and update the test/comment accordingly.

2. Split official OpenAI request policy from `openai-compat`.

`OpenAiCompatProvider` wraps `OpenAiProvider`, so the new model-family logic for `max_completion_tokens` and OpenAI-specific output caps leaks into local/proxy OpenAI-compatible providers. That contradicts the compatibility note in the PR and can break providers that still expect the legacy `max_tokens` dialect. Please add an explicit request dialect/policy: official OpenAI can choose `max_tokens` vs `max_completion_tokens` and apply OpenAI caps; `openai-compat` should keep the legacy `max_tokens` behavior and avoid OpenAI-specific caps unless we intentionally opt into them.

3. Preserve server behavior for `bootstrap --dry-run` with zero sources.

The new client-side dry-run returns success with an empty report, while the server dry-run path returns `NoSources` / 422. That can hide wrong repo paths or overly broad excludes. Please make local dry-run return an error when source collection is empty, matching the server behavior.

4. Handle gpt-5/o-series temperature policy.

The PR switches those models to `max_completion_tokens`, but still forwards caller temperatures such as `0.1` or `0.2` from lint/bootstrap/consolidation. Those model families commonly reject non-default temperature values, so gpt-5/o-series can still 400 after this fix. Please omit or normalize `temperature` for the same official OpenAI model family, with a regression test.

Non-blocking cleanup:

- The `required` comment says caller-supplied trimmed arrays are preserved, but the code correctly overwrites them. Please update the comment.
- The gpt-4-turbo cap is lower than 16384, so it may still produce a clear 400 there. That is acceptable if intentional, but the comment should not imply the cap fully covers gpt-4-turbo.

Validation on my side was otherwise green:

- `git diff --check origin/main...HEAD`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-llm openai`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-cli commands::bootstrap::tests`
- `TAILWIND_SKIP=1 cargo test --workspace --all-targets -- --skip watcher::tests::picks_up_externally_created_file`

This is close. I would be happy to merge after the focused compatibility fixes above.

pablowinck:
Thanks for the careful review. All four blocking items + both non-blocking cleanups landed in 1a0c95e. End-to-end validation included a real bootstrap against `gpt-4o-mini` (Official) **and** `llama3:latest` via openai-compat pointing at Ollama (Compat) — both generated three wiki pages without 4xx.

| # | Item | Where |
|---|------|-------|
| 1 | Force `additionalProperties: false` (drop preservation) | `enforce_strict_object_schemas` in `openai.rs`. Test `enforce_strict_overwrites_caller_additional_properties_true` flipped to assert overwrite. |
| 2 | Split policy: official vs compat | New `RequestDialect { Official, Compat }`. `OpenAiCompatProvider::new` now selects `Compat`, which keeps the legacy `max_tokens` wire format, skips OpenAI-specific caps, doesn't omit temperature, and bypasses schema normalisation entirely. Test `build_request_compat_dialect_keeps_max_tokens_and_temperature` locks the contract end-to-end. |
| 3 | dry-run errors on empty bundle | Extracted `ensure_sources_for_dry_run` in `commands/bootstrap.rs` for unit testability. Tests `ensure_sources_for_dry_run_rejects_empty` / `_accepts_non_empty`. Functional run on an empty `--exclude-*`-everything bundle now errors with the `NoSources`-shaped message. |
| 4 | Omit `temperature` for gpt-5 / o-series under `Official` | New `model_requires_default_temperature` (aliased to the same family check), wired into the `Official` arm of `build_request`. Tests `build_request_omits_temperature_for_gpt5` and `build_request_keeps_temperature_for_gpt4`. |
| A | `required` doc reworded | Docstring of `enforce_strict_object_schemas` now says "Both rules are unconditional here" and explicitly notes the function overwrites a caller-supplied partial `required`. |
| B | gpt-4-turbo cap note | `max_output_tokens_for` doc now states explicitly that gpt-4-turbo's real cap is 4096 and that a generous bootstrap call to it will still 400 — that model is outside the strict-mode target audience. |

Local gates:

```
cargo fmt --all -- --check                                                   ✓
cargo clippy --workspace --all-targets -- -D warnings  (TAILWIND_SKIP=1)     ✓
cargo test --workspace --all-targets                                          ✓
  -- --skip watcher::tests::picks_up_externally_created_file
```

Branch is pushed to `pablowinck/ai-memory:fix/openai-provider-bugs` and the PR head was updated.

pablowinck:
Noticed you merged the first three commits into main via \`pr10-test\` — thanks. The PR is rebased onto the new main so it now contains **only the follow-up commit** (\`db99d2d\`) that addresses the review feedback. The diff is no longer a mix of already-merged work and review responses.

Single commit on top of latest main covers, all six review items:

| # | Item | Where |
|---|------|-------|
| 1 | Force \`additionalProperties: false\` | \`enforce_strict_object_schemas\`: \`map.insert(\"additionalProperties\", false)\` unconditional. Test \`enforce_strict_overwrites_caller_additional_properties_true\`. |
| 2 | Split policy: official vs compat | New \`RequestDialect { Official, Compat }\` on \`OpenAiProvider\`. \`OpenAiCompatProvider::new\` switches to \`Compat\`, which keeps legacy \`max_tokens\` only, no OpenAI caps, no temperature massaging, no strict-mode schema normalisation. Test \`build_request_compat_dialect_keeps_max_tokens_and_temperature\`. |
| 3 | dry-run errors on empty bundle | Extracted \`ensure_sources_for_dry_run\` for unit testability. Functional run on empty bundle now errors with \`NoSources\`-shaped message, matching the server. Tests \`ensure_sources_for_dry_run_rejects_empty\` / \`_accepts_non_empty\`. |
| 4 | Omit \`temperature\` for gpt-5 / o-series under \`Official\` | New \`model_requires_default_temperature\` (aliased to the same family check), wired into the \`Official\` arm of \`build_request\`. Tests \`build_request_omits_temperature_for_gpt5\` + \`build_request_keeps_temperature_for_gpt4\`. |
| A | \`required\` doc reworded | Docstring of \`enforce_strict_object_schemas\` says "Both rules are unconditional here" and explicitly notes the function overwrites a caller-supplied partial \`required\`. |
| B | gpt-4-turbo cap note | \`max_output_tokens_for\` doc now states explicitly that gpt-4-turbo's real cap is 4096 — that model is outside the strict-mode target audience. |

Your six commands, all green:

\`\`\`
cargo fmt --all -- --check                                                   ✓
git diff --check origin/main...HEAD                                          ✓
TAILWIND_SKIP=1 cargo test -p ai-memory-llm openai                            (23 passed)
TAILWIND_SKIP=1 cargo test -p ai-memory-cli commands::bootstrap::tests        (5 passed)
TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings        ✓
TAILWIND_SKIP=1 cargo test --workspace --all-targets \\
  -- --skip watcher::tests::picks_up_externally_created_file                  (36 passed, 1 skipped)
\`\`\`

Functional E2E ran on both dialects after the rebase:

- \`Official\` against \`gpt-4o-mini\`: bootstrap of a small repo generated 3 wiki pages (manifest + 2 ADR-shaped decisions), no 4xx.
- \`Compat\` against \`llama3:latest\` via openai-compat pointing at Ollama: bootstrap of the same repo generated 3 pages, no 4xx — proves the dialect split unblocks local backends.

Issue #11, opened 2026-05-25
Windows support: build, test, and runtime fixes


mrpaiva:
## Problem this would solve

The project currently targets Linux (and presumably macOS) but does not build
or test cleanly on Windows. As a Windows user I'd like to contribute the
fixes needed to make `cargo build --workspace && cargo test --workspace` pass
on Windows without workarounds, and ensure the runtime generates correct
hook payloads for Windows shells.

### Environment
- OS: Windows 11 Pro 10.0.26200
- Rust: 1.95.0 (via rust-toolchain.toml)
- MSVC: Visual Studio 2022 Professional
- ai-memory version: 0.1.3 (commit cc115f9)

### What breaks today

**Build (P0)** — `cargo build` panics in `ai-memory-web/build.rs` unless
`TAILWIND_SKIP=1` is set:

```
thread 'main' panicked at crates/ai-memory-web/build.rs:134:18:
No pinned Tailwind SHA-256 for tailwindcss-windows-x64.exe.
Set TAILWIND_SKIP=1 and provide static/tailwind.css manually.
```

`expected_tailwind_sha256()` only has a hash for `tailwindcss-linux-x64`.
The download fallback also relies on `curl`/`wget`, missing a native
Windows alternative.

**Tests (P1)** — 1 test fails, others are fragile on Windows:

- `render_shared::tests::claude_code_payload_emits_absolute_paths` —
  hardcodes `/home/user/.ai-memory/hooks/claude-code/session-start.sh`
- Several test fixtures across `admin_phase3.rs`, `payload.rs`,
  `server.rs`, and `watcher.rs` use Unix paths like `/tmp/x`,
  `/home/user/alpha-repo`, `/data/wiki/...`

**Runtime (P2)** — hook payloads are POSIX-only:

- `render_shared.rs` generates `VAR=val /path/script.sh` syntax (doesn't
  work on Windows CMD or PowerShell)
- `shell_quote()` uses POSIX single-quote escaping
- `install_hooks.rs` uses `/dev/null` as a sentinel path for agents that
  don't need hook scripts (Omp, Openclaw)
- Hook script discovery only looks for `.sh` files

**Degraded functionality (P3):**

- `atomic.rs` `inode_of()` returns `0` on non-Unix, so the file watcher's
  own-write detection is disabled — may cause unnecessary re-indexing

## Proposed solution

I'm willing to submit a PR (or a series of PRs if you prefer) addressing
these in priority order:

[...]

PR #12, opened 2026-05-25
feat: Windows support — build, test, and runtime fixes

mrpaiva:
## What changed

- **Build (P0):** `build.rs` now has pinned SHA-256 hashes for all 5 Tailwind
  platform binaries (Linux x64/arm64, macOS x64/arm64, Windows x64) and a
  PowerShell `Invoke-WebRequest` download fallback. `cargo build` no longer
  panics on Windows without `TAILWIND_SKIP=1`.
- **Tests (P1):** Path fixtures across 5 files use `cfg!(windows)` to provide
  platform-appropriate absolute paths. All 74+ tests now pass on Windows.
- **Runtime (P2):** Hook payloads on Windows use PowerShell syntax
  (`$env:VAR='val'; & powershell -NoProfile -File 'script.ps1'`) instead of
  POSIX (`VAR=val /path/script.sh`). Added `.ps1` equivalents of all 35 hook
  scripts. `install_hooks` discovers `.ps1` on Windows and uses `NUL` instead
  of `/dev/null` as sentinel.
- **Inode (P3):** `atomic.rs` now uses `winapi-util`'s
  `GetFileInformationByHandle` to get the NTFS file index, enabling the
  watcher's own-write skip optimization on Windows (previously returned 0).

## Why

The project does not build or test cleanly on Windows (see #11). These changes
make `cargo build --workspace && cargo test --workspace` pass on Windows 11
with no workarounds, and ensure the hook scripts actually work when invoked by
agent CLIs on Windows.

## Test plan

- [x] `cargo test --workspace` passes (all tests, no `TAILWIND_SKIP`)
- [x] Manual test: `cargo build --workspace` succeeds on Windows 11 Pro,
  Tailwind binary downloads and verifies via SHA-256. New
  `inode_is_nonzero_on_all_platforms` test confirms nonzero file index on NTFS.
  New `claude_code_payload_emits_powershell_commands_on_windows` test confirms
  PowerShell-syntax hook commands.

## Notes for reviewers

- **`winapi-util` over `windows-sys`:** The workspace has `unsafe_code = "forbid"`,
  so the inode implementation uses `winapi-util` (already a transitive dependency
  via `walkdir`) instead of raw `windows-sys` FFI calls.
- **No Linux regression:** All `cfg!(windows)` branches are compile-time; the
  Unix code paths are unchanged. Existing tests continue to use Unix paths on
  non-Windows platforms.
- **35 `.ps1` hook scripts** mirror the existing `.sh` scripts exactly, using
  `Invoke-RestMethod` instead of `curl`. The `session-start.ps1` variant
  includes the handoff GET, matching the `.sh` behavior.

Ref: #11

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -250,6 +250,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tracing",
+ "winapi-util",
 ]
 
 [[package]]
--- a/crates/ai-memory-cli/src/commands/install_hooks.rs
+++ b/crates/ai-memory-cli/src/commands/install_hooks.rs
@@ -27,6 +27,12 @@ use crate::commands::render_shared::{
 };
 use crate::config::Config;
 
+/// Platform-appropriate extension for hook scripts: `.ps1` on
+/// Windows, `.sh` everywhere else.
+fn hook_script_ext() -> &'static str {
+    if cfg!(windows) { "ps1" } else { "sh" }
+}
+
 /// Run the `install-hooks` subcommand.
 ///
 /// # Errors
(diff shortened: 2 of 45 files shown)


mrpaiva:
Closing this PR — will wait for feedback on #11 before resubmitting.

PR #13, opened 2026-05-25
feat(llm): add Gemini provider with gemini-2.5-flash default

rpaggi:
Adds Google Gemini as a fourth LlmProvider alongside Anthropic, OpenAI, and openai-compat. Uses native responseSchema for structured output, inlining schemars $refs and stripping Draft-2020-12 keywords Gemini rejects. Reads GEMINI_API_KEY (or GOOGLE_API_KEY) and selects via AI_MEMORY_LLM_PROVIDER=gemini; defaults to gemini-2.5-flash when no model is set. CLI llm-test gains a matching --provider gemini choice.

## What changed

- New `GeminiProvider` (`crates/ai-memory-llm/src/gemini.rs`) implementing `LlmProvider` against Google's Generative Language API, with native `responseSchema` for structured output.
- `ProviderChoice::Gemini` added to the factory, with `GEMINI_API_KEY` (falling back to `GOOGLE_API_KEY`) plumbed through `Config` → `build_provider`.
- `AI_MEMORY_LLM_PROVIDER` now accepts `gemini` (and `google` as a synonym); model defaults to `gemini-2.5-flash` when unset.
- `ai-memory llm-test --provider gemini` wired through the CLI.
- README + ARCHITECTURE + deploy/install docs updated to list the fourth provider.

## Why

Gemini 2.5 Flash is a strong price/perf option for consolidation and lint workloads, and a generous free tier makes it especially appealing for self-hosted ai-memory users who don't want to pay Anthropic/OpenAI for background memory maintenance. Adding it as a first-class provider — rather than going through openai-compat — preserves the project rule of typed native clients per provider (CLAUDE.md §Stack; cognee #2840 lesson on LiteLLM/Instructor drift).

## Test plan

- [x] Manual test: `AI_MEMORY_LLM_PROVIDER=gemini GEMINI_API_KEY=… ./target/debug/ai-memory llm-test --provider gemini` returned a well-formed JSON object matching the schema.

## Notes for reviewers

- **Schema transform is the trickiest part.** Gemini's `responseSchema` only accepts an OpenAPI 3 subset, so `gemini.rs` inlines schemars `$ref`s and strips Draft-2020-12 keywords (`$schema`, `$defs`, certain `additionalProperties` shapes) before sending. The `compile_schema` helper + its tests are worth a careful look — that's where drift will hurt first.
- **No embeddings.** Gemini's embed endpoint isn't wired here; only generation. Embedders remain OpenAI/Voyage.
- **API key aliasing.** `GEMINI_API_KEY` is preferred, but `GOOGLE_API_KEY` is accepted as a fallback because Google's own docs still use the older name in several places.
- **Provider name aliasing.** `AI_MEMORY_LLM_PROVIDER=google` is treated as `gemini` for the same reason.
- Stays faithful to invariant #13 (zero-LLM default): the provider only constructs when explicitly selected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -595,7 +595,25 @@ AI_MEMORY_LLM_MODEL=openai/gpt-5.4-mini
 LLM_API_KEY=sk-or-v1-…
 ```
 
-### Option 3 - Local Ollama qwen3:32b *(free / self-hosted)*
+### Option 3 - Google Gemini 2.5 Flash *(generous free tier)*
+
+Hosted by Google, with a free tier that's enough for most
+personal consolidation loads. Structured-output is native
+(`responseSchema` over OpenAPI 3 subset); the client inlines
+`$ref`s and strips Draft-2020-12 keywords Gemini rejects, so
+schemars-derived schemas just work. Use `gemini-2.5-flash`
+for the speed/cost balance; avoid the *-thinking* variants
+(see "What we don't recommend" below).
+
+Grab a key from https://aistudio.google.com/apikey, then:
+
+```bash
+AI_MEMORY_LLM_PROVIDER=gemini
+AI_MEMORY_LLM_MODEL=gemini-2.5-flash      # optional; this is the default
+GEMINI_API_KEY=AIza…                      # GOOGLE_API_KEY also accepted
+```
+
+### Option 4 - Local Ollama qwen3:32b *(free / self-hosted)*
 
 $0 per consolidation. Requires a machine with at least ~24 GB
 of unified or VRAM memory to keep the Q4_K_M weights warm
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -407,6 +407,8 @@ pub enum LlmProviderChoice {
     Anthropic,
     /// OpenAI Chat Completions.
     Openai,
+    /// Google Gemini (Generative Language API).
+    Gemini,
     /// OpenAI-compatible local (Ollama, vLLM, LM Studio).
     OpenaiCompat,
 }
--- a/crates/ai-memory-cli/src/commands/llm_test.rs
+++ b/crates/ai-memory-cli/src/commands/llm_test.rs
@@ -52,6 +52,7 @@ impl From<LlmProviderChoice> for ProviderChoice {
         match value {
             LlmProviderChoice::Anthropic => Self::Anthropic,
             LlmProviderChoice::Openai => Self::OpenAi,
+            LlmProviderChoice::Gemini => Self::Gemini,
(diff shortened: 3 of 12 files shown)


akitaonrails (changes requested):
Thanks for the Gemini provider work. The branch is close, but I found one blocker before this should merge.

## Blocking

`gemini-2.5-flash` enables dynamic thinking by default, but the provider never sends `generationConfig.thinkingConfig.thinkingBudget = 0`.

Relevant code:

- `crates/ai-memory-cli/src/config.rs:316-320` defaults Gemini to `gemini-2.5-flash`.
- `crates/ai-memory-llm/src/gemini.rs:87-96` defines the generation config without `thinkingConfig`.
- `crates/ai-memory-llm/src/gemini.rs:190-195` sends only max tokens, temperature, mime type, and schema.

Why this matters: Gemini 2.5 Flash uses dynamic thinking when `thinkingBudget` is omitted. Those thinking tokens count against `maxOutputTokens`, so structured JSON can be truncated or absent even when the visible output is tiny. I reproduced this with a direct `generateContent` call using `maxOutputTokens: 64`: without `thinkingConfig`, the response spent 47 tokens on thoughts and returned only `{"ok`; with `"thinkingConfig":{"thinkingBudget":0}`, the same request returned `{"ok":true}` with `finishReason: STOP`.

ai-memory consolidation/lint is a background structured-output workload, and the project already avoids reasoning-mode defaults for exactly this reason: hidden reasoning burns token budget, increases latency/cost, and can produce empty/truncated strict JSON.

Please either:

1. Add `thinkingConfig: { thinkingBudget: 0 }` by default for Gemini models that support disabling thinking, especially the default `gemini-2.5-flash`, or
2. Choose a non-thinking Gemini default and document how to opt into thinking with a larger token budget.

A regression test should assert the serialized Gemini request includes `generationConfig.thinkingConfig.thinkingBudget = 0` for the default model.

## Verified

I ran:

- `TAILWIND_SKIP=1 cargo test -p ai-memory-llm`
- `docker build -t ai-memory:pr13 -f docker/Dockerfile .`

Those passed, so the issue is runtime behavior against Gemini 2.5 Flash defaults rather than compile/test breakage.

akitaonrails:
Maintainer follow-up after merge: I applied the Gemini 2.5 Flash thinking-budget fix in bf55fdb. Requests for gemini-2.5-flash now serialize generationConfig.thinkingConfig.thinkingBudget = 0, with unit coverage to assert the field is present for Flash and omitted for non-Flash models.

Verification:
- cargo fmt --check
- git diff --check
- docker build -t ai-memory:gemini -f docker/Dockerfile .

Live API check before the fix showed default Gemini 2.5 Flash spent hidden thought tokens and hit MAX_TOKENS; the same prompt with thinkingBudget 0 returned strict JSON with STOP.

PR #16, opened 2026-05-25
feat(hooks): `.ai-memory.toml` marker file for cwd→workspace mapping

pablowinck:
Adds support for a `.ai-memory.toml` marker file that lets a directory tree declare which `(workspace, project)` pair its `cwd` belongs to, without depending on `basename($cwd)` or post-hoc `rename-project`.

## Motivation

Today the hot path — `POST /hook` — hardcodes `workspace = DEFAULT_WORKSPACE_NAME` and derives `project = basename(cwd)`. That works for the solo dev in `~/projects/<repo>`, but breaks down for three concrete workflows:

- **Multi-client consultancies** (`~/projects/<client>/<repo>`): every client should land in a dedicated workspace; `basename` collisions silently merge memory across clients (e.g. `cnl/cnl-api` and `pe/pe-cnl-api` both bucket as `pe-cnl-api`).
- **Work / personal / open-source separation** for solo devs who want context isolation.
- **Mono-repos** with a single conceptual project across N packages, where the basename-per-package default scatters memory across N buckets.

`--workspace foo` already exists for `bootstrap` / `purge-project` / `rename-project` — administrative one-shots. It does not reach `/hook`. Without a marker file, the only mechanism is `rename-project` **after** the fact: you discover you mixed memory only after mixing it.

## Solution

A `.ai-memory.toml` marker file anywhere above the cwd:

```toml
workspace = "movvia"        # required
project   = "pe-portais"    # optional; forces project for the subtree
```

Lifecycle hook scripts walk up from `cwd` to find the closest marker (stops at `$HOME` to avoid leaking declarations on multi-user boxes), append `&workspace=X&project=Y` to the `POST /hook` and `GET /handoff` URLs. The server treats both as optional overrides; absent marker means the previous behaviour, **byte-identical**.

## Design notes

- **Hook-side walk-up, not server-side.** The server runs in Docker; host paths aren't visible in the container. Walk-up has to live where the filesystem does.
- **Query params, not headers.** Debuggable with `curl --trace` or server access logs. Akita's review on #10 implicitly validated this style.
- **No glob patterns, no ancestor merging, no env-var fallback.** Explicit anti-features in [`docs/marker-file.md`](docs/marker-file.md); easier to add later than to remove.
- **`$HOME` upper bound for walk-up, not `/`.** Multi-user systems shouldn't leak a root-level marker.
- **Single `_lib.sh` source-of-truth in `hooks/`** copied to per-agent staging dirs by `stage_hook_scripts`. No drift across `claude-code/`, `codex/`, `cursor/`, `gemini-cli/`, `opencode/` to maintain.

## Backward compatibility

| Scenario | Behaviour |
|---|---|
| Server new + hooks new + no marker | byte-identical to today (no extra params, no extra request) |
| Server new + hooks new + marker | new path — workspace declared by the marker |
| Server new + hooks old | hooks send no params → server falls through to defaults → identical to today |
| Server old + hooks new | server ignores unknown query params (axum default) → identical to today |
| Schema | no migration — no new tables, no column changes |
| MCP tool schemas | unchanged |
| CLI flags | unchanged |

Semver-wise this is a **minor** release: pure additive change, every existing flow keeps working.

## Test coverage

- **Rust unit tests (5 new):** `crates/ai-memory-hooks/src/router.rs`
  - `workspace_override_yields_distinct_workspace` — proves the same `cwd` lands in a different `WorkspaceId` when a marker declares one.
  - `project_override_replaces_basename` — proves an explicit `project` overrides `basename(cwd)`.
  - `matching_overrides_collapse_to_same_pair` — two different `cwd`s with the same override pair resolve to the same `(WorkspaceId, ProjectId)`.
  - `cache_does_not_poison_across_override_variants` — composite cache key keeps with-override and without-override variants of the same `cwd` isolated.
  - `overrides_resolve_without_cwd` — handoff fetch can resolve via overrides alone (no cwd cache hit needed).
- **Rust unit tests (2 new):** `crates/ai-memory-cli/src/commands/install_hooks.rs`
  - `stage_hook_scripts_copies_shared_lib_sh` — verifies the staging step picks up `_lib.sh` from the parent dir alongside the event scripts.
  - `stage_hook_scripts_tolerates_missing_lib_sh` — older source bundles without the helper still install cleanly.
- **Shell suite (14 cases):** `tests/hooks/test_lib.sh`
  - `find_marker` walks up correctly + stops at `$HOME` (pinned in the test fixture).
  - `parse_toml_key` returns first match, ignores comments/blanks, tolerates absent files.
  - `extract_cwd` pulls `cwd` from common payload shapes.
  - `url_encode` escapes the meaningful chars (`& = ? # %` and space).
  - `marker_qs` builds the right suffix for one or both keys; empty when no marker.

## Local validation

All four `CONTRIBUTING.md` gates run on my Apple Silicon laptop:

```
cargo fmt --all -- --check                                                ✓
cargo clippy --workspace --all-targets -- -D warnings                     ✓
cargo test --workspace --all-targets \
  -- --skip watcher::tests::picks_up_externally_created_file              ✓ (36 passed)
cargo deny check                                                          ✗ (pre-existing on main)
```

`cargo deny` fails identically to the last several `main` runs — `webpki-roots-1.0.7` (CDLA-Permissive-2.0 license) and `yaml-rust 0.4.5 / bincode 1.3.3` (unmaintained advisories). Both come from dependencies introduced by `graph retrieval` and `OMP integration` on `main` — none added by this PR. Happy to bundle a `deny.toml` allowlist update in this PR or open a separate one, your call.

Plus the shell suite:

```
sh tests/hooks/test_lib.sh                                                ✓ (14 passed)
```

And the `git diff --check origin/main...HEAD` whitespace pass:

```
git diff --check origin/main...HEAD                                       ✓ (clean)
```

## End-to-end smoke

Built `docker build -t ai-memory:marker-final -f docker/Dockerfile .` and ran a real `POST /hook` against the container with a marker file declaring `workspace = "movvia"` / `project = "monorepo-x"` above the cwd. Result:

```sql
SELECT w.name, p.name FROM workspaces w JOIN projects p ON p.workspace_id = w.id;
default | scratch
movvia  | monorepo-x
```

Exactly the declared pair lands in the store, not the basename-of-cwd fallback. Paridade reversa (cwd with no marker → `default / basename(cwd)`) also exercised in unit tests.

## Anti-scope explicit in the docs

- ❌ No glob patterns (walk-up by literal ancestry).
- ❌ No merge of ancestor markers (closest wins).
- ❌ No automatic migration of existing `default`-workspace projects (CLI `rename-project` already covers it).
- ❌ No `AI_MEMORY_WORKSPACE` env override.

Open to discussion on any of those if you'd take them; deliberately left them out to keep the PR focused and the contract obvious.

## Files changed

```
crates/ai-memory-hooks/src/payload.rs              +40   -2
crates/ai-memory-hooks/src/router.rs              +205  -54
crates/ai-memory-cli/src/commands/install_hooks.rs +98   -16
hooks/_lib.sh                                     +83  new
tests/hooks/test_lib.sh                          +106  new
hooks/claude-code/*.sh                            +56  -71
docs/marker-file.md                              +136  new
README.md                                          +6   -2
CHANGELOG.md                                      +10   -0
```

Total: +571 / -86 across 12 files.


akitaonrails:
Merged #16 locally as `0fa892b` and added follow-up commit `e577e4b` on `main`.

Adjustment in `e577e4b`:
- Resolved the rebase conflict with hook-ingest backpressure by keeping both `DEFAULT_HOOK_INGEST_MAX_IN_FLIGHT` and the new `ProjectCache` type.
- Fixed workspace-only marker handoff lookup by forwarding `cwd` with marker query params, so `workspace = "acme"` plus basename-derived project resolves on `GET /handoff`.
- Applied the marker helper path to POSIX Codex, Cursor, Gemini CLI, and OpenCode hook scripts, not just Claude Code. PowerShell already used the shared helper.
- Moved staging tests off the real `dirs::data_local_dir()` path by injecting a temp data dir, so readonly home environments pass.
- Added a router regression test for workspace-only marker handoff lookup and updated the shell helper test expectations.

Verification after the full integration stack:
- `cargo fmt --check`
- `git diff --check`
- `sh tests/hooks/test_lib.sh`

PR #23, opened 2026-05-26
fix(hooks): resolve worktree cwd to main repo for project identity

mrpaiva:
## What changed

The hook router's `resolve_project_ids` now resolves git worktree
directories back to the main repository root before falling back to
`basename(cwd)`. Previously, each worktree created a separate project
entry because the router only used `basename(cwd)`, and worktrees
have their own directory names.

**Before:** Working in `MultiClubes/21258-mover-mi...` (a worktree)
created a project called `21258-mover-mi...` instead of `MultiClubes`.

**After:** All worktrees of the same repo share one project, matching
the CLI's existing behavior.

## Why

When using git worktrees, knowledge was fragmented across duplicate
projects — handoffs, queries, and recent activity were invisible
across worktrees of the same repository.

The CLI already handled this correctly via `discover_repo_root()`,
but the hook router (the hot path for interactive sessions) did not.

## How

- Added `discover_main_repo_root()` to `ai-memory-consolidate` using
  `git2::Repository::commondir().parent()` — unlike `workdir()`,
  `commondir()` always points to the main repo's `.git/` even from
  a linked worktree.
- Bare repos are guarded (`is_bare()` → fall through to basename).
- The router tries git discovery first, falls back to `basename(cwd)`
  for non-git directories, preserving existing behavior.
- Cost: one `git2::Repository::discover()` call per unique `cwd`,
  then cached.

## Test plan

- [x] `cargo test --workspace` passes (4 pre-existing failures in
  `render_shared` on Windows, unrelated to this change)
- [x] Manual test: `worktree_resolves_to_same_project_as_main_repo` —
  creates a real git repo + worktree via `git2`, asserts both cwd
  paths resolve to the same `(workspace_id, project_id)`
- [x] `non_git_dir_falls_back_to_basename` — plain directories still
  resolve via basename
- [x] `bare_repo_falls_back_to_basename` — bare repos fall back
  safely to basename instead of returning the grandparent directory

## Notes for reviewers

- `discover_main_repo_root` vs existing `discover_repo_root`: the
  existing function uses `workdir()` which returns the worktree's own
  directory, not the main repo. The new function uses
  `commondir().parent()` which always resolves to the main repo root.
- The new function lives in `ai-memory-consolidate::bootstrap` next
  to the existing `discover_repo_root`. Long-term it could move to a
  shared utility module, but the dependency already exists.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -132,6 +132,7 @@ dependencies = [
  "ai-memory-wiki",
  "anyhow",
  "axum",
+ "git2",
  "jiff",
  "regex",
  "serde",
--- a/crates/ai-memory-consolidate/src/bootstrap.rs
+++ b/crates/ai-memory-consolidate/src/bootstrap.rs
@@ -536,6 +536,38 @@ pub fn discover_repo_root(start: &Path) -> Result<PathBuf, BootstrapError> {
         .ok_or_else(|| BootstrapError::NotARepo(start.to_path_buf()))
 }
 
+/// Like [`discover_repo_root`], but when `start` is inside a git
+/// worktree the function follows the `.git` commondir pointer back to
+/// the **main** repository root rather than returning the worktree's
+/// own working directory.
+///
+/// This is the right choice when the caller needs a stable identity
+/// for the repository (e.g. deriving a project name) — all worktrees
+/// of the same repo should resolve to the same root.
+///
+/// # Errors
+/// Returns `BootstrapError::NotARepo` when no repository is found
+/// at or above `start`.
+pub fn discover_main_repo_root(start: &Path) -> Result<PathBuf, BootstrapError> {
+    let repo = git2::Repository::discover(start)
+        .map_err(|_| BootstrapError::NotARepo(start.to_path_buf()))?;
+    // Bare repos have no working directory; `commondir()` IS the repo root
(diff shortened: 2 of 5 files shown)


akitaonrails:
Merged to main and added follow-up hardening in fd63f5c. I kept the previous default routing intact (project = basename(cwd)) and made repo-root/worktree collapsing opt-in via .ai-memory.toml with project_strategy = repo-root; explicit project overrides still win. I also forwarded the new marker key through POSIX, PowerShell, OpenCode, and OMP hooks, updated docs/changelog, and added regression coverage for default basename behavior, opt-in worktree collapse, override precedence, unknown-strategy fallback, and cache isolation. Validation passed: cargo fmt --check, git diff --check, TAILWIND_SKIP=1 cargo test --workspace, TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings, plus sh tests/hooks/test_lib.sh.

akitaonrails:
Thanks for the PR. It was merged in `b93f01a` and hardened in `fd63f5c`: I preserved the existing basename-by-default routing, made repo-root/worktree collapsing opt-in via `project_strategy = "repo-root"`, kept explicit project overrides winning, forwarded the marker through all hook integrations, and added regression coverage/docs.

PR #28, opened 2026-05-26
fix(hooks): self-heal when a cached project is deleted (e.g. purge-project)

### LINKED ISSUE #27: Hook capture breaks with FOREIGN KEY constraint failed after purge-project (stale project cache)
## Summary

After `POST /admin/purge-project` on a **running** server, hook capture stops working: every event logs `hook processing failed error=sqlite: FOREIGN KEY constraint failed` and no session/observation is persisted — until the process is restarted.

## Root cause

The hook router keeps an in-memory `project_cache` (`crates/ai-memory-hooks/src/router.rs`) mapping `(cwd, workspace_override, project_override)` → `(WorkspaceId, ProjectId)`. `resolve_project_ids` returns cached ids without touching the DB.

`purge-project` deletes the project row (cascade), but **does not invalidate this cache**. So the next event with the same key gets the **deleted** `project_id` from the cache, skips `get_or_create_project` (which would recreate it), and `begin_session` inserts a session referencing a non-existent project → `sessions.project_id` foreign key fails. It persists until a restart clears the cache.

The admin purge handler (`AdminState`) and the hook router (`HookState`, which owns the cache) are separate states, so the purge path can't easily evict the entry.

## Repro

1. Capture an event for some cwd → project created + cached.
2. `POST /admin/purge-project` for that workspace/project.
3. Capture another event for the same cwd → `FOREIGN KEY constraint failed`; nothing ingested until restart.

## Proposed fix

Make the router **self-healing**: on a `begin_session` failure, evict the stale cache slot, re-resolve (which recreates the project via `get_or_create_project`), and retry once. This is robust to any cause of an externally-deleted project, not just `purge-project`, and stays local to the hook router (no admin↔hooks coupling).

PR with the fix + a regression test follows.

djalmajr:
Fixes #27.

## Problem

The hook router caches `(workspace, project)` ids per `(cwd, overrides)`. When a project is deleted out from under it — e.g. `purge-project` on a live server — the in-memory cache still points at the dropped row, so `begin_session` trips the `sessions.project_id` foreign key on **every** subsequent event, and capture stays broken until the process restarts (which clears the cache).

## Fix

On a `begin_session` failure, evict the stale cache slot, re-resolve (recreating the project via `get_or_create_project`), and retry once. Robust to any cause of an externally-deleted project, not just `purge-project`, and stays local to the hook router (the admin purge handler and the router hold separate state, so the router self-healing avoids coupling them).

- Extracts `cache_key_for(..)` so `resolve_project_ids` (insert/lookup) and `process` (eviction) agree on the slot.
- One-shot retry in `process`; `begin_session` is idempotent (`ON CONFLICT DO NOTHING`) and `get_or_create_project` is safe, so the retry is harmless even if the first failure wasn't FK-related.

## Test

`process_self_heals_when_cached_project_purged`: create+cache a project, `purge_project` it (cache still holds the dead id), then process another event for the same cwd — asserts it recreates instead of FK-failing, and the data lands. Full suite green (42 passed).

### DIFF
--- a/crates/ai-memory-hooks/src/router.rs
+++ b/crates/ai-memory-hooks/src/router.rs
@@ -247,6 +247,22 @@ fn render_handoff_markdown(h: &Handoff) -> String {
 /// 2. `project_override` OR `basename(cwd)` OR fallback to
 ///    `state.project_id` (when `cwd` is also unavailable).
 ///
+/// Build the `project_cache` key from the (normalised cwd, workspace
+/// override, project override) triple. Shared by `resolve_project_ids`
+/// (insert/lookup) and `process` (eviction on the stale-cache retry) so
+/// the two always agree on the slot.
+fn cache_key_for(
+    cwd_norm: Option<&str>,
+    workspace_override: Option<&str>,
+    project_override: Option<&str>,
+) -> (String, String, String) {
+    (
+        cwd_norm.unwrap_or_default().to_string(),
+        workspace_override.unwrap_or_default().to_string(),
+        project_override.unwrap_or_default().to_string(),
+    )
+}
+
 /// Cache key is `(cwd, workspace_override, project_override)` so the
 /// same `cwd` resolved with and without an override (e.g. during a
 /// hook-script upgrade window) doesn't poison each other's slot.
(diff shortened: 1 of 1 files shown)


akitaonrails:
Merged and pushed to `main` in `3ee11ab`.

I resolved the merge conflict against the newer `project_strategy` cache-key work and kept the self-heal eviction on the full `(cwd, workspace_override, project_override, project_strategy)` key. I also added a repo-root strategy regression so this does not silently fall back to the old three-part cache key.

Additional changes:
- added a changelog entry for the live-server stale project-cache recovery
- kept the one-shot retry local to hook routing

Validation passed:
- `cargo fmt --check`
- `git diff --check`

akitaonrails:
Thanks for the PR. It was merged in `3ee11ab`; during the merge I adapted the self-heal path to the newer four-part `(cwd, workspace_override, project_override, project_strategy)` cache key and added a repo-root strategy regression plus changelog entry so the fix does not regress marker-file routing.

PR #31, opened 2026-05-26
fix(openai): unblock memory_consolidate multi_page on OpenAI strict mode

### LINKED ISSUE #30: memory_consolidate multi_page=true returns 400 from OpenAI strict mode (schemars $ref siblings + oneOf)
**Version**
ai-memory 0.2.0 (built from main, also reproduces on `akitaonrails/ai-memory:latest` running 0.1.3).

**What happened**
`memory_consolidate(multi_page=true)` against any session returns:

```
provider error 400: Invalid schema for response_format 'Result':
context=('properties', 'tier'), $ref cannot have keywords {'description'}.
```

After patching the `$ref` sibling, a second 400 surfaces:

```
Invalid schema for response_format 'Result': In context=(), 'oneOf' is not permitted.
```

**What I expected**
Multi-page consolidation succeeds against the OpenAI provider and persists `_rules/`/`decisions/`/`gotchas/` pages alongside the session summary.

**Steps to reproduce**
1. `docker run -e AI_MEMORY_LLM_PROVIDER=openai -e AI_MEMORY_LLM_MODEL=gpt-4o-mini -e OPENAI_API_KEY=... akitaonrails/ai-memory:latest`
2. Let any agent populate a session via the lifecycle hooks (a handful of prompts).
3. Call `memory_consolidate(session_id=<that session>, multi_page=true)` via MCP.

**Environment**
- OS: any (reproduced on CachyOS host + Debian 12 inside docker)
- Docker: yes — `akitaonrails/ai-memory:latest`
- Agent: Claude Code 2.1.x
- Transport: http

**Root cause**
Two issues in the JSON Schema emitted to OpenAI:

1. `schemars 1.x` emits a field-level `description` next to `$ref` for every doc-commented field typed as an external enum. Example from `ConsolidatedPageUpdate`:
   ```json
   "tier": { "$ref": "#/$defs/Tier", "description": "..." }
   ```
   OpenAI strict mode rejects this (JSON Schema 2020-12 §8.2.3.1 — `$ref` siblings are ignored during validation; OpenAI hardens that to a 400).

2. `schemars 1.x` emits Rust enums as `oneOf` with `{type: "string", const: "x"}` branches. OpenAI strict mode rejects `oneOf` outright (accepted equivalent is `anyOf`).

Both apply to `ConsolidatedBatch` (the multi-page path). `ConsolidatedPage` (single-page) only contains primitives, so it never hits either.

I have a fix ready — PR to follow.

joaosouzacoder:
## What changed

`enforce_strict_object_schemas` in `crates/ai-memory-llm/src/openai.rs` now performs two additional normalisations during its walk:

1. **Strip every sibling of `$ref`.** When an object node carries `$ref`, retain only `$ref` and short-circuit. Per JSON Schema 2020-12 §8.2.3.1, `$ref` siblings are ignored during validation, so the rewrite is semantically a no-op — only the documentation-only field-level annotation is lost; the type's own description on the referenced `$defs` entry still reaches the model.

2. **Rewrite `oneOf` → `anyOf`.** For closed enum sets the branches are non-overlapping by construction (exactly one matches per value), so `oneOf` ≡ `anyOf` in this codebase's emission. Lossless.

Together these unblock `memory_consolidate multi_page=true` end-to-end against `gpt-4o-mini` and `gpt-4o`.

## Why

Without the fix, every call to `memory_consolidate multi_page=true` (and `bootstrap` against OpenAI) returns a 400 — see #30 for the two distinct error messages.

`schemars 1.x` emits both shapes from existing types: a field-level `description` next to `$ref` for every doc-commented field typed as an external enum (e.g. `tier: Tier`, `kind: PageKind` on `ConsolidatedPageUpdate`), and `oneOf` with `{const: "x"}` branches for every Rust enum (`Tier`, `PageKind`).

Closes #30.

## Test plan

- [x] `cargo test --workspace` passes (454 passing, 1 ignored, 0 failing — including 3 new tests in `enforce_strict_*` that go RED on `main` and GREEN with this patch)
- [x] Manual test: built image from this branch via `docker/Dockerfile`, ran with `AI_MEMORY_LLM_PROVIDER=openai`, `AI_MEMORY_LLM_MODEL=gpt-4o`, called `memory_consolidate session_id=<…> multi_page=true` via Claude Code MCP — outcome array returned, session page + `_rules/<slug>.md` persisted on disk.

## Notes for reviewers

- The `oneOf` → `anyOf` rewrite is safe only because every `schemars` emission from a Rust enum has disjoint branches by construction. If a future contributor wants a true exclusive-or across non-disjoint branches, they'd need a different normalisation — none exist in the current codebase.
- Field-level descriptions are documentation-only in 2020-12, but the loss could surprise a reader inspecting a saved schema. I judged it acceptable; the type-level description on the `$defs` entry still reaches the model. Happy to switch to inlining `$ref`s instead if you'd prefer to keep field-level descriptions visible.
- Two commits, one per normalisation — chronological order of discovery (the `oneOf` bug only surfaces after the `$ref` one is fixed). Squash welcome at merge time if the maintainer prefers.

### DIFF
--- a/crates/ai-memory-llm/src/openai.rs
+++ b/crates/ai-memory-llm/src/openai.rs
@@ -346,6 +346,22 @@ impl OpenAiProvider {
 fn enforce_strict_object_schemas(value: &mut serde_json::Value) {
     match value {
         serde_json::Value::Object(map) => {
+            // OpenAI's structured-output validator rejects any sibling
+            // keyword next to `$ref` with a 400 (per JSON Schema 2020-12
+            // §8.2.3.1, $ref siblings are ignored during validation —
+            // OpenAI hardens that into a hard error). schemars 1.x emits
+            // a field-level `description` next to `$ref` for every
+            // doc-commented field typed as an external enum (e.g.
+            // `tier: Tier` on `ConsolidatedPageUpdate`); without this
+            // strip, `memory_consolidate multi_page=true` 400s on every
+            // call. Dropping the siblings is semantically a no-op —
+            // only the documentation-only annotation is lost; the
+            // type's own description on the referenced `$defs` entry
+            // is what reaches the model.
+            if map.contains_key("$ref") {
+                map.retain(|k, _| k == "$ref");
+                return;
+            }
             let is_object = map
                 .get("type")
                 .and_then(|t| t.as_str())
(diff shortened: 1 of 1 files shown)


akitaonrails:
Thanks for the fix and the detailed repro in #30. I independently audited the OpenAI strict-schema path, checked the current OpenAI Structured Outputs subset, merged the PR, and added a maintainer follow-up in 1e7f10d to tighten the comments, add a schemars-generated enum-ref regression test, and update changelog/deploy docs.\n\nValidation passed locally with fmt, diff whitespace, targeted ai-memory-llm tests, full workspace tests, and clippy. The final main CI is green: https://github.com/akitaonrails/ai-memory/actions/runs/26468911515

PR #35, opened 2026-05-26
fix(test): make Claude Code hook render tests portable across platforms

mrpaiva:
## Summary

- Two tests in `render_shared::tests` (`claude_code_payload_embeds_auth_token_when_provided` and `claude_code_payload_emits_absolute_paths`) called `build_claude_code_payload()` which delegates to `HookCommandPlatform::current()`, then asserted POSIX-style inline env-var syntax. On Windows, `current()` returns `Windows` and produces PowerShell output, breaking the assertions.
- Fix: call `build_hook_payload_for_platform()` with an explicit `HookCommandPlatform::Posix`, since these tests verify the POSIX rendering path. The Windows path is already covered by `windows_payload_uses_powershell_and_ps1_hooks`.
- Switch `ends_with` to `contains` in the absolute-path assertion, since Windows paths contain backslashes that trigger `shell_quote` wrapping.

**Scope:** Claude Code tests only — per issue #11 feedback, other agents (Cursor, Antigravity) are left as-is until their Windows hook runner behavior is verified.

## Test plan

- [x] `cargo fmt --all -- --check` — pass
- [x] `cargo clippy -p ai-memory-cli -- -D warnings` — pass
- [x] `cargo test -p ai-memory-cli -- render_shared::tests::claude_code` — 5/5 pass on Windows
- [x] Existing `windows_payload_uses_powershell_and_ps1_hooks` and `posix_hook_command_quotes_script_path_and_shell_metachars` still pass

Ref: #11 (PR 1 of 3)

### DIFF
--- a/crates/ai-memory-cli/src/commands/render_shared.rs
+++ b/crates/ai-memory-cli/src/commands/render_shared.rs
@@ -496,7 +496,14 @@ mod tests {
     #[test]
     fn claude_code_payload_embeds_auth_token_when_provided() {
         let root = PathBuf::from("/host/hooks/claude-code");
-        let v = build_claude_code_payload(&root, "http://localhost:49374", Some("tok"));
+        let v = build_hook_payload_for_platform(
+            &CLAUDE_CODE_EVENTS,
+            &root,
+            "http://localhost:49374",
+            Some("tok"),
+            HookShape::Nested,
+            HookCommandPlatform::Posix,
+        );
         // Env vars are inlined into the command string so CC's
         // hook runner sees them regardless of whether it honours
         // a separate `env` field. Assert the token landed in the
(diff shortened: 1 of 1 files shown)


akitaonrails:
Merged this first, then added follow-up commit 40f2f37: https://github.com/akitaonrails/ai-memory/commit/40f2f37

Your PR fixed the two Claude Code assertions, but the same Windows-simulated suite still failed for Cursor and Antigravity:

- `cursor_payload_uses_flat_shape` still asserted POSIX inline env syntax while rendering via the current platform.
- `antigravity_payload_uses_named_groups_with_mixed_shape` still asserted POSIX inline env syntax and `stop.sh` while rendering via the current platform.

The follow-up keeps the tests clean by adding one small POSIX hook-payload helper for POSIX-specific assertions, using it for Claude/Cursor, and explicitly rendering the Antigravity test with `HookCommandPlatform::Posix`. The dedicated Windows behavior remains covered by `windows_payload_uses_powershell_and_ps1_hooks`.

Verified with:

- `AI_MEMORY_HOOK_PLATFORM=windows TAILWIND_SKIP=1 cargo test -p ai-memory-cli render_shared`
- `cargo fmt --check`
- `git diff --check`

PR #36, opened 2026-05-27
fix(hooks): resolve Windows backslash paths to basename on Linux server

mrpaiva:
## Summary

- **Bug:** `std::path::Path::file_name()` on Linux treats `\` as a literal character, so Windows `cwd` values like `E:\source\ai-memory` sent via hooks were stored as the full path string instead of resolving to `ai-memory`. This caused duplicate projects in the database — one from bootstrap (correct basename) and one from hooks (full Windows path).
- **Fix:** Replace `Path::file_name()` with a manual `rsplit(['/', '\\'])` in the `basename()` helper inside `resolve_project_ids`, so both Unix and Windows path separators are handled regardless of the server's platform.
- **Test:** Added `windows_backslash_path_resolves_to_basename` that verifies two different Windows paths with the same final component resolve to the same project.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo test -p ai-memory-hooks -- windows_backslash` — new test passes
- [x] All 52 existing router tests still pass
- [ ] After merge + server redeploy: verify Windows hooks no longer create duplicate full-path projects

### DIFF
--- a/crates/ai-memory-hooks/src/router.rs
+++ b/crates/ai-memory-hooks/src/router.rs
@@ -350,10 +350,16 @@ async fn resolve_project_ids(
     }
 
     fn basename(path: &std::path::Path) -> Option<String> {
-        path.file_name()
-            .and_then(|s| s.to_str())
-            .map(str::to_string)
-            .filter(|s| !s.is_empty())
+        // Split on both `/` and `\` so Windows paths sent to a Linux
+        // server still resolve to the final component.
+        let s = path.to_str()?;
+        let name = s
+            .rsplit(['/', '\\'])
+            .find(|seg| !seg.is_empty())?;
+        if name.is_empty() {
+            return None;
+        }
+        Some(name.to_string())
     }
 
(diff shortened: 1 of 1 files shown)


akitaonrails:
Merged this, then added follow-up commit 434e2df: https://github.com/akitaonrails/ai-memory/commit/434e2df

The fix itself was worth keeping: Windows `cwd` strings arriving at a Linux server really do bypass `Path::file_name()` splitting, which can create duplicate projects named with the full backslash path.

The follow-up covers two review nits:

- Added the missing `CHANGELOG.md` entry because this changes supported Windows/Linux hook routing behavior.
- Removed the redundant `if name.is_empty()` branch after `.find(|seg| !seg.is_empty())?`.

Verified locally with:

- `TAILWIND_SKIP=1 cargo test -p ai-memory-hooks windows_backslash`
- `cargo fmt --check`
- `git diff --check`

PR #37, opened 2026-05-27
fix(wiki): implement inode tracking on Windows via NTFS file index

mrpaiva:
## Summary

- The watcher's own-write detection relies on `inode_of()` to skip files it just wrote. The `#[cfg(not(unix))]` fallback returned `0`, disabling this on Windows — every atomic write triggered a spurious reindex.
- Add a `#[cfg(windows)]` implementation using `winapi-util::file::information().file_index()` (the NTFS equivalent of a Unix inode). The crate is a lightweight, pure-Rust wrapper around `GetFileInformationByHandle`.
- `winapi-util` is a `cfg(windows)`-only dependency — zero cost on Linux/macOS builds.
- The `#[cfg(not(any(unix, windows)))]` fallback still returns `0` for other platforms.
- Test assertion updated to expect a non-zero inode on Windows too.

## Test plan

- [x] `cargo fmt --all -- --check` — pass
- [x] `cargo clippy -p ai-memory-wiki -- -D warnings` — pass
- [x] `cargo test -p ai-memory-wiki -- atomic` — 4/4 pass on Windows (including `writes_atomically_and_creates_parents` now asserting `ino != 0`)
- [x] Verified `file_index()` returns distinct non-zero values for different files on NTFS

Ref: #11 (PR 2 of 3)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,6 +254,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tracing",
+ "winapi-util",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,9 @@ notify = "8"
 notify-debouncer-full = "0.6"
 sysinfo = "0.32"
 
+# Windows NTFS file-index (inode equivalent) for own-write detection.
+winapi-util = "0.1"
+
 # MCP server SDK.
 rmcp = { version = "1.7", features = ["server", "macros", "transport-io", "transport-streamable-http-server", "schemars"] }
 schemars = "1"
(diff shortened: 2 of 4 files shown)


akitaonrails:
Merged in efa1c94 🎉 — verified it doesn't affect the Linux/macOS build (`winapi-util` is correctly scoped under `[target.'cfg(windows)']`) and the full suite stays green (520 tests).

This fixes a real bug: previously Windows returned `0` for every file's inode, so the watcher's own-write set matched `0 == 0` for all events — external edits (e.g. Obsidian) wouldn't reindex reliably. Real NTFS file indices fix that.

I added a small follow-up commit (b6c9f1d) with a rationale comment on the Windows branch (why the file index = inode analog, the open-handle requirement, FAT-returns-0 behavior) to match the repo's 'document the why' convention. Nothing functional changed. Thanks for the clean, well-scoped fix, @mrpaiva! 🙏

PR #39, opened 2026-05-27
feat(llm,docker): add claude-shim for Anthropic via claude -p OAuth

LLawli:
## Summary

Adds an optional Anthropic `/v1/messages` compatible HTTP shim that
translates requests into `claude -p` (Claude Code headless mode)
subprocess invocations, letting `ai-memory` route Anthropic LLM calls
through a Claude Max/Pro subscription instead of an API key. Opt-in
via a new docker compose profile; the default path keeps hitting
`api.anthropic.com`.

## Why

Consolidation, lint, and bootstrap burn API tokens per call. Users
who already pay for a Claude Max/Pro subscription have no built-in
way to substitute that. This adds one without introducing a new
`LlmProvider` or letting subprocess abstractions leak into the
ai-memory core.

## How it works

- `ai-memory`'s existing `AnthropicProvider` is redirected via
  `base_url` to a sibling container.
- The shim translates `/v1/messages` requests into
  `claude -p ... --output-format json` subprocess calls.
- For structured outputs (the `tools[0]` + `tool_choice` strategy
  used by consolidation), the shim injects the JSON schema into
  claude's system prompt, validates the result with the \`jsonschema\`
  crate (one retry on failure), and synthesises a `tool_use` content
  block — preserving the exact shape `complete_structured_raw`
  expects.

## What changed

**ai-memory core**
- `crates/ai-memory-llm/src/factory.rs` — Anthropic arm now applies
  `config.base_url` via `with_base_url` after resolving auth (was
  silently ignored).
- `crates/ai-memory-llm/tests/anthropic_base_url.rs` — wiremock
  integration test proving the redirect end-to-end.

**New shim (`docker/claude-shim/`)**
- Standalone Cargo project (not a workspace member): axum HTTP
  server, single-turn text + structured (`tool_use`) modes.
- Multi-stage Dockerfile bundles Node 20 +
  `@anthropic-ai/claude-code`.
- Refuses startup without `CLAUDE_CODE_OAUTH_TOKEN` or a resolvable
  `claude` binary.
- 16 tests (12 unit + 4 integration with a fake `claude` shell
  script in a tempdir).

**Compose / docs**
- `docker/docker-compose.yml` — opt-in `claude-shim` profile,
  internal `ai-memory-net` bridge, `AI_MEMORY_ALLOWED_HOSTS`
  widened so sibling client containers can reach the server by
  container name.
- `docker/.env.example` — env-var template documenting
  `CLAUDE_CODE_OAUTH_TOKEN` + the three `AI_MEMORY_LLM_*` activators.
- `README.md` — new section between **LLM Providers** and
  **Architecture** with setup walkthrough and caveats; existing
  content untouched.

## Test plan

- [x] \`cargo fmt --check\`, \`cargo clippy --workspace --all-targets -- -D warnings\`, \`cargo-deny check\` — local + CI green.
- [x] \`cargo test -p ai-memory-llm\` (incl. new wiremock integration test) — green.
- [x] Shim project standalone: \`cargo fmt/clippy/test\` — 16 tests passing.
- [x] End-to-end on a real repo: \`ai-memory bootstrap\` against an external project with the shim active produced 17 wiki pages (8 concepts + 4 ADRs + 4 gotchas + bootstrap manifest), all grounded in real commits/README/CLAUDE.md — zero Anthropic API spend.

## Caveats / non-goals (documented in README)

- Single-turn only; no streaming or \`cache_control\` breakpoints.
- Embeddings still need a real provider — \`claude -p\` doesn't
  generate them. Ollama with \`nomic-embed-text\` is the
  no-API-cost option.
- Per-request \`claude -p\` cold-start (~1-2s); a long-lived
  \`--input-format stream-json\` session is documented as future
  work.
- OpenAI/Gemini equivalent is out of scope (Anthropic-only by
  design — only the Anthropic arm gets the \`base_url\` patch).


akitaonrails:
Thank you for this, @LLawli — it's genuinely well-engineered work. The subprocess handling is safe (no shell, stdin-delivered prompts, timeouts), the JSON-Schema-validate-then-retry loop is a nice touch, the security posture (opt-in profile, internal-only network, non-root) is sound, and it actually builds with 16 passing tests. I read all of it carefully before deciding.

I'm going to **close this without merging**, and I want to be transparent about why — it's about *shape*, not quality.

### Why not merge as-is
ai-memory already has an established pattern for "use a subscription instead of an API key": **in-Rust OAuth providers** (`openai_oauth.rs`, `copilot.rs`) that read a credential and call the provider's HTTP API directly, wired through a small `auth.rs` abstraction + `factory.rs`. They live in the workspace, run under `clippy`/`cargo-deny`/`cargo test` in CI, and add no runtime dependencies.

The claude-shim is a different shape: a separate ~3.4k-line crate + Docker image that shells out to the `claude` CLI. That puts it outside the workspace/CI, adds a Node + `claude-code` install to maintain, and leans on subprocess-wrapping Claude Code as an unattended backend — which is the part I'm most wary of (see ToS note below).

### What I did instead
The key realization: Anthropic **OAuth and API-key hit the same `/v1/messages` with the same body — only the auth headers differ**. So instead of an external translator, I added an OAuth *auth mode* to our existing `AnthropicProvider`, which **reuses all of its tested tool-use / structured-output logic** (the translation your shim had to rebuild from scratch). OAuth mode sends `Authorization: Bearer` + `anthropic-beta: oauth-2025-04-20,claude-code-20250219` and omits `x-api-key`.

Result — new provider in ~360 lines incl. tests, in-workspace, in-CI:
**Commit: f9c432e** (https://github.com/akitaonrails/ai-memory/commit/f9c432e)

```bash
claude setup-token
export AI_MEMORY_LLM_PROVIDER=anthropic-oauth
export ANTHROPIC_OAUTH_TOKEN=<token>   # CLAUDE_CODE_OAUTH_TOKEN also accepted
```

### A caveat we now document loudly
Your PR pushed me to make this explicit: this OAuth path is **unofficial and against Anthropic's usage policies** — there's no sanctioned subscription API, and using it as an automated backend may get an account rate-limited or banned. We ship it as opt-in with a prominent "use at your own risk" warning (README, `docs/install.md`, compose). I couldn't smoke-test the live API in CI, so the header recipe (cross-checked against oh-my-pi's defaults) is validated by unit tests only.

### Could you test it?
You clearly have a working Claude OAuth setup — would you give `f9c432e` a spin against a real `claude setup-token` token and let me know if `anthropic-oauth` covers your use case? If the live API needs anything extra (a different beta value, a user-agent, etc.), that's a quick follow-up and your feedback would be the fastest way to nail it. Thanks again — this feature exists because you raised it. 🙏

LLawli:
Thanks for the careful read. I ran the test on a real use case (the one that prompted the shim in the first place), and here are the results.

**E2E:** Ran the bootstrap e2e against the Rust branch of XClaudeUsage ([`SrDarf/XClaudeUsage @ HighPerformanceXClaudeUsage`](https://github.com/SrDarf/XClaudeUsage/tree/HighPerformanceXClaudeUsage)) with `AI_MEMORY_LLM_PROVIDER=anthropic-oauth`. 14 pages generated (Haiku, 7638 input tokens), **zero auth errors**. Header recipe works against the live API as-is.

**A/B with the shim path** as a sanity check: at first glance the shim looked marginally more precise on citation (`file:line` refs, `rationale` populated). That impression didn't survive a controlled test.

**Hypothesis test** — reran all three paths with clean state, same model (Haiku 4.5), same 7638-token source bundle. Full diff + analysis in this gist: https://gist.github.com/LLawli/517dcc37da770bc7b8a02eb59d0fc790

| Marker | Shim (unpatched) | OAuth (unpatched) | OAuth (patched) |
|---|---|---|---|
| `file.ext:line` citations | 2 | 1 | 3 |
| Unique commit hashes cited | 6 | 8 | 7 |
| `§Section` references | 0 | 0 | 21 |
| Bare-file refs | 11 | 20 | 12 |
| Pages | 13 | 13 | 14 |
| Total lines | 527 | 443 | 571 |

Verdict (revising my earlier read): the shim's citation edge was **LLM variance, not a structural property**. `claude -p` is non-interactive and does NOT inherit the Claude Code harness system prompt, both paths land at `/v1/messages` with essentially the same prompt; default Haiku already cites reasonably well on both. But the system prompt **is** shapeable: a markdown-only patch to `bootstrap_system.md` (new `## Citations — be precise` section) takes `§Section` refs from 0 → 21 and triples `file:line` citations on the OAuth path. **Patched OAuth meets or exceeds shim on every citation dimension.** All cited commit hashes validated via `git cat-file` and asserted zero hallucinations across all three runs.

**TL;DR:** the shim doesn't cover a gap that needs covering.

Happy to send a proper PR with the prompt patch if you want it merged — or pick it up directly from the gist.

Thanks again — and the loud ToS disclaimer landing in README/docs/compose was the right call.

PR #44, opened 2026-05-27
fix(hooks): align antigravity pre-tool script contract

ArtroxGabriel:
Match pre-tool-use script behavior to the documented input/output contract. Raise explicit errors for other scripts and remove error suppression.

## What changed

- Aligned pre-tool-use script handling with the documented input/output [contract](https://antigravity.google/docs/hooks#inputoutput-contract) .
- Removed error suppression for non-matching scripts.
- Added explicit errors when a script does not match the expected pre-tool-use behavior.

## Why

- Fixes a contract mismatch between implementation and docs.
- Makes failures visible instead of silently suppressing them.
- Improves correctness and debuggability for hook execution.

## Test plan

- [ ] Manual test: not run

## Notes for reviewers

- This change is intentionally narrow: it updates script handling to match the documented contract and removes suppression paths for non-pre-tool scripts.

### DIFF
--- a/hooks/antigravity-cli/pre-tool-use.ps1
+++ b/hooks/antigravity-cli/pre-tool-use.ps1
@@ -1,4 +1,4 @@
 . "$PSScriptRoot\..\lib\ai-memory-hook.ps1"
 Invoke-AiMemoryHook -Event "pre-tool-use" -Agent "antigravity-cli"
-[Console]::Out.WriteLine("{}")
+[Console]::Out.WriteLine('{ "decision": "allow" }')
 exit 0
--- a/hooks/antigravity-cli/pre-tool-use.sh
+++ b/hooks/antigravity-cli/pre-tool-use.sh
@@ -10,7 +10,7 @@ PAYLOAD=$(cat)
 CWD=$(ai_memory_extract_cwd "$PAYLOAD")
 QS=$(ai_memory_marker_qs "$CWD")
 
-printf '%s' "$PAYLOAD" \
-    | ai_memory_post_hook "$SERVER/hook?event=pre-tool-use&agent=antigravity-cli${QS}" >/dev/null 2>&1 || true
-printf '{}\n'
+printf '%s' "$PAYLOAD" |
+    ai_memory_post_hook "$SERVER/hook?event=pre-tool-use&agent=antigravity-cli${QS}" >/dev/null 2>&1 || true
+printf '{"decision": "allow"}\n'
 exit 0


akitaonrails:
Thanks for the PR. I do not want to merge this as-is because it changes the hook failure semantics in a way that can break Antigravity.

Main blocker: the scripts no longer suppress `ai_memory_post_hook` stdout. `POST /hook` returns a plain `queued` body on success, so a successful `pre-tool-use.sh` can emit `queued{"decision":"allow"}` instead of valid JSON. That violates the stdout contract this PR is trying to fix.

Other concerns:
- Hook ingestion is intentionally fire-and-forget and non-disruptive. Server-down/network errors should not surface during routine telemetry hooks.
- The PowerShell scripts are left inconsistent with the `.sh` scripts.
- The PR has no CI result and the test checklist is still unchecked.

If you want this merged, please narrow the change to the actual contract fix:
- Keep POST stdout/stderr suppression for all telemetry calls: `>/dev/null 2>&1 || true`.
- Keep server/network failures non-blocking and quiet for normal capture hooks.
- Change only the Antigravity `PreToolUse` stdout response from `{}` to `{"decision":"allow"}` where required.
- Apply the same behavior consistently to `.sh` and `.ps1`.
- Add or update a small regression test if possible, and confirm fmt/clippy/tests pass.

With that shape, this should be a small safe PR rather than a semantic change to hook reliability.

ArtroxGabriel:
Thanks for the detailed feedback @akitaonrails .

I have updated the PR to address these points:

* **Restored stdout/stderr suppression**
* **Scoped the changes:** The logic is now narrowed down to only return `{"decision":"allow"}` for the Antigravity `PreToolUse` hook, while keeping other hooks silent and non-blocking.
* **Consistency:** Applied identical logic to both `.sh` and `.ps1` scripts.

I have also completed the test plan:
* [x] `cargo fmt --all -- --check` passes
* [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
* [x] `cargo test --workspace` passes

Please let me know if there is anything else that needs adjustment.

akitaonrails:
Verified against the original concerns and the diff matches every point you addressed:

- `>/dev/null 2>&1 || true` is preserved on the post-hook call — the `queued` body can't bleed into stdout (the original blocker).
- Only `pre-tool-use.{sh,ps1}` change; other hooks remain silent, so no contract violation elsewhere.
- Both `.sh` and `.ps1` get the identical `{"decision":"allow"}` output (the documented Antigravity PreToolUse contract).
- Hook stays fire-and-forget; server errors never surface to the agent.

Merging. Thanks @ArtroxGabriel for the careful revision. 🙏

PR #51, opened 2026-05-29
fix: skip -u flag on macOS to prevent UID mismatch crash

abnersajr:
## Summary

On macOS, the wrapper script passes `-u $(id -u):$(id -g)` to one-shot containers. This causes a **UID mismatch**: the data volume is owned by uid 1000 (the container's internal `ai-memory` user), but the host UID on macOS is typically 501/502.

The one-shot container then crashes with:

```
thread 'main' panicked at .../rolling.rs:156:14:
initializing rolling file appender failed: InitError {
  context: "failed to create log file",
  source: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
}
```

This affects all thin-client commands: `status`, `install-mcp`, `install-hooks`, `search`, `write-page`, etc.

## Fix

- Only pass `-u` on non-macOS platforms. On macOS, omitting `-u` lets the container run as its default uid 1000 which matches the volume owner. Docker Desktop's file sharing layer handles permissions correctly without explicit UID mapping.
- Initialize `USER_ARGS=()` before the platform case to satisfy `set -u`.
- Fix empty array expansions for `TTY_ARGS`, `NETWORK_ARGS`, and `ENV_ARGS` under `set -u` (same class of bug as #50).

## Test Plan

- [x] `bash -n bin/ai-memory` — syntax check passes
- [x] Verified on macOS: `ai-memory status`, `ai-memory install-mcp --client claude-code` run without permission errors
- [ ] Verified on Linux: `-u $(id -u):$(id -g)` still passed to containers

Supersedes #50 (includes the empty-array fix from that PR).

### DIFF
--- a/bin/ai-memory
+++ b/bin/ai-memory
@@ -248,6 +248,7 @@ done
 # 127.0.0.1 would mean "this helper container", so thin-client commands like
 # `status` and `bootstrap` could not reach the default server.
 NETWORK_ARGS=()
+USER_ARGS=()
 case "$(uname -s 2>/dev/null || true)" in
   Linux)
     if [ -z "${AI_MEMORY_SERVER_URL:-}" ]; then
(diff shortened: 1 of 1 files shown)


akitaonrails:
Thanks @abnersajr, long time no see🙏 — solid diagnosis. The macOS UID-mismatch crash is a real one and the Darwin / Linux split is the right shape for the fix; the set-u empty-array hardening on `TTY_ARGS` / `NETWORK_ARGS` / `ENV_ARGS` (rolling in #50's intent) is exactly the canonical guard.

Going to merge now, with a small follow-up commit on top to address two things the diff misses:

1. **Linux silently loses `-u`.** The `case` has `Linux)`, `Darwin)`, `*)` arms — but the `Linux)` arm only assigns `NETWORK_ARGS`, never `USER_ARGS`, so it falls through with the pre-case `USER_ARGS=()` default. Net effect: Linux users now also get no `-u`, regressing the prior behaviour (most visible to anyone with a non-1000 host UID or a bind-mounted `AI_MEMORY_DATA_DIR`). Fix: set `USER_ARGS=(-u "$(id -u):$(id -g)")` inside the `Linux)` arm so only `Darwin)` actually skips `-u`.
2. **`"${USER_ARGS[@]}"` needs the same `set -u` guard you applied to the other three arrays.** macOS's default `/bin/bash` is 3.2.57, which fails on empty-array expansion under `set -u` (the exact bug fix #2 in this PR is meant to address). It works in your testing because Homebrew's bash 5.x is almost certainly first on your `PATH`, but a default-bash macOS user would re-hit the same crash on the new line. Fix: `${USER_ARGS[@]+"${USER_ARGS[@]}"}`, matching the surrounding style.

Squash-merging now and committing the follow-up on top — full credit to you in both commits.

PR #55, opened 2026-05-29
feat(wiki): admission webhook chain (engine extension point)

djalmajr:
## What
An **admission webhook chain**: on a page write / consolidate, the engine POSTs the page + context to operator-configured external webhooks (`[[admission_webhooks]]` in config, or `AI_MEMORY_ADMISSION_WEBHOOKS__N__*`). Actor identity propagates into the admission context, and a skip-list prevents a webhook's own writes from looping back through the chain.

## Why
The write path has no extension seam today — extending it (e.g. mirroring every page to a private git remote for durable backup, or notifying an external system) means forking the engine. This adds the seam without touching the storage invariants.

## Design
- `failure_policy=ignore` by default: a flaky or absent webhook never blocks the engine's write path.
- An extension *seam*, not a runtime plugin system — external processors use the existing HTTP/MCP contract; they cannot bypass the sanitizer, hook backpressure, or the single-writer SQLite actor.
- Actor identity + skip-list flow through `AdmissionContext`, so loop-prevention is explicit and testable.
- Docs: `docs/admission-webhooks.md`.

## Verification
- `cargo test --workspace` — green, except the 3 `commands::reset::tests::*` that abort when other `ai-memory` processes are alive on the host (ambient, not a regression — same on upstream/main)

Rebased on v0.7.0. Purely additive — no contract change for existing agents.


akitaonrails:
Hey Djalma — we can see you've been actively rebasing on top of the v0.8 multi-user work and adding more capabilities (`memory_delete_page`, project-purge admission, actor identity propagation). Thanks for staying on top of it.

Quick context on the pause from our side: when you opened this we were still pre-multi-user, and several of the concerns I raised in my earlier audit (the actor model, the `mcp-auth` / Keycloak JWT scaffolding, the `sub`/`client` fields) were genuinely hard to resolve cleanly without a proper Phase 1 multi-user story in core first. So rather than ask you to design around scaffolding that didn't exist yet, we deliberately took the time to land the groundwork:

- **Phase 1 multi-user** (P1.1 → P1.8 + follow-ups) is all in `main` now — `ActorContext` + `AuthLevel` in `ai_memory_core`, the `users` table, `ai-memory user add/list/expire/revive/rotate-token` CLI, four-rung auth middleware injecting `Extension<ai_memory_core::ActorContext>` on every request, `pages.author_id` + `audit_log.author_id`, `/api/v1` author field, web UI "Last edited by" chip, `install-hooks --as-user`, full `docs/users.md`.
- **HTTPS direction settled** — we're not doing in-binary TLS; `docs/https-via-proxy.md` + Caddy/Cloudflare Tunnel docker compose templates landed instead, so the multi-user-beyond-loopback deployment story has a real answer.

This means your PR now sits on much firmer ground: the `Extension<ai_memory_core::ActorContext>` carrier already exists in the request layer (populated by the four-rung middleware from P1.3), the `X-Memory-Actor-*` header bridge isn't load-bearing anymore, and `mcp-auth` becomes one optional way to populate the actor rather than the assumed integration point.

We're happy to do a proper review pass when you're at a natural stopping point — drop a comment here (or convert the PR from draft / flag it as ready) and we'll work through the merge together.

Appreciate the persistence on this.

djalmajr:
Rebased onto v0.8.0 — ready for review, CI green, no conflicts.

**Rebase reconciliation:**
- `write_page` now sequences attribution + admission cleanly: stamp `last_modified_by` → run the admission chain (still free to mutate frontmatter/body) → derive title/links from the mutated markdown → emit/upsert.
- Converged the actor model onto `ai_memory_core::ActorContext` — deleted the duplicate `ai_memory_wiki::ActorContext`. `AdmissionContext.actor` is now the core type, `write_page` fills the webhook `ctx.actor` from `req.actor` (single identity source — the same value that stamps `last_modified_by`), and `actor_from_headers` returns the canonical type.

**Design call I'd like your input on — actor resolution as a resolver chain.**
The `users` table is a great fit for a self-contained single-user / homelab install. But identity often comes from *outside* ai-memory — e.g. an org fronting the MCP with Keycloak/OIDC (oauth2-proxy) so only authorized people reach it, and wanting those identities to flow through for attribution. There are already several sources in play: root token (rung 1), the `users` table (rung 2), and the external-proxy `X-Memory-Actor-*` bridge (the `sub`/`client` rung you scaffolded) — but today the MCP handler reads the header bridge directly to fill `req.actor`.

The cleaner shape — same primitive as the admission chain itself — is to make identity a **chain of resolvers**: the handler reads only `Extension<ActorContext>`, and the middleware composes resolvers (`bearer→users`, `root→template`, `X-Memory-Actor-*→external-proxy`), so a new source (OIDC introspection, mTLS, …) is a new resolver rather than a change to the handlers. The header bridge then becomes the external-proxy resolver — which is what your reserved `sub`/`client` fields are for. Native-token installs and Keycloak/IdP-fronted installs are both first-class, and the choice becomes a deployment decision instead of a code fork.

If you're up for that direction, I'll wire the handlers to read `Extension<ActorContext>` and move the header mapping into the middleware as the external-proxy resolver.

**Second point — `author_id` resolution.** The column + store plumbing are fully wired, but no write path resolves `ActorContext.user → UserId`, so `pages.author_id` / `audit_log.author_id` land NULL — attribution currently lives only in the frontmatter `last_modified_by`. Cleanest is the rung-2 resolver injecting the `UserId` alongside the actor; left it out here since it's your multi-user domain.

djalmajr:
On webhook delivery — added `blocking: bool` to `WebhookConfig` (default `true`, so existing configs are unchanged):
- `blocking = true` — runs synchronously in `run`, may mutate, a `Reject` aborts the write.
- `blocking = false` — dispatched fire-and-forget *after* the page lands on disk; the engine doesn't await it and ignores its response, so it can't mutate/reject. For pure mirrors/backups (e.g. `git-mirror`), a slow/down sink no longer adds latency to writes. Same `events` subscription + skip-list still apply.

Worst-case write latency is `Σ timeout_ms` over the **blocking** webhooks only. `write_page` / `delete_page` / `purge_project` all fire the non-blocking set after the blocking chain + the on-disk write. Test covers run-skips-it / async-dispatches-it; docs updated.

akitaonrails:
Merged manually on main at 556ce48 after rebasing/integrating with current main.

What I changed during integration:
- Preserved #61 project/workspace read-write resolution, #62 FTS tokenizer/trigger behavior, and #63 read-page DB fallback semantics.
- Hardened admission dispatch: bounded streaming response reads, second sanitizer pass after webhook mutations, non-blocking hooks only after durable operations, and batch writes through admission.
- Hardened destructive paths: delete runs blocking admission before file/index removal and restores the quarantined file on DB failure; purge admission now runs before DB/file deletion and non-blocking purge hooks observe after completion.
- Switched MCP/admin write/delete actor propagation to authenticated request extensions instead of trusting raw actor headers; regular DB users cannot use the skip-list header to bypass admission.
- Added regression coverage for admission mutations, batch admission, delete-page, purge rejection-before-delete, actor spoofing, DB-user skip-list bypass, and docs/tool surface updates.

Validation:
- cargo fmt --check
- git diff --check

Thanks for the feature work; I took it over to land the destructive-operation and post-#61/#62/#63 integration details.

PR #56, opened 2026-05-29
feat(mcp): cross-project recall — global search + broaden-on-miss guidance

djalmajr:
## What
Two related changes to how an agent recalls across projects:

1. **`global=true` on `memory_query`** — search every project in one call, with each hit annotated by its workspace + project. Reuses `ReaderPool::search_pages_with_meta` (already powering `/api/v1` global search); validates that `global` is not combined with `project`/`workspace`/`scopes`.
2. **Routing-snippet guidance** (`SNIPPET_BODY` + `MEMORY_INSTRUCTIONS`) teaching the agent to broaden when the current project misses: use `scopes` for named siblings, or `global=true` to search everything, and that `memory_query` returns snippets — read the full page with `memory_read_page`.

## Why
`memory_query` defaults to the current project. When an agent doesn't know which project holds a fact (cross-cutting infra/ops notes, a sibling app), it searches one project, gets nothing, and concludes "never recorded" — when the knowledge exists one project over. The capability (global search) existed at the reader layer but wasn't reachable over MCP; the guidance closes the behavioral gap.

## Verification
- `cargo test --workspace` — green except the ambient `commands::reset::tests::*` (alive-process abort; same on upstream/main)
- New: `memory_query_global_searches_all_projects`, `memory_query_global_rejects_explicit_scope`, `prompts_teach_cross_project_search_strategy`

Additive: `global` defaults to false, so existing single-project queries are unchanged.

### DIFF
--- a/crates/ai-memory-core/src/routing_snippet.rs
+++ b/crates/ai-memory-core/src/routing_snippet.rs
@@ -60,7 +60,7 @@ match the intent to the tool. They do not need to name the tool.
 
 | User says / situation | Tool |
 |---|---|
-| "have we discussed X?" / "search memory for Y" / before proposing architecture | `memory_query` |
+| "have we discussed X?" / "search memory for Y" / before proposing architecture | `memory_query` (current project; `scopes` for named siblings; `global=true` to search every project) |
 | "what's been going on" / "show recent activity" (light) | `memory_recent` |
 | "is ai-memory healthy?" / "how big is the wiki?" | `memory_status` |
 | "give me the stats" / structured snapshot for the agent to consume | `memory_briefing` |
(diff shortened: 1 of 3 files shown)


akitaonrails:
Audit pass — `global=true` mutex correctly rejects combination with `scopes/workspace/project`, hits across 2 workspaces / 3 projects surface in one call carrying their workspace+project annotation, and the validation properly treats whitespace-only strings as "not set".

Spotted one small inconsistency I'll patch as a follow-up commit instead of asking you to re-push: the new SNIPPET_BODY + MEMORY_INSTRUCTIONS "broaden the search" paragraphs both contain the sentence "**there is no global 'search everything' mode**" right after the *same* documents' table-row / bullet that advertises `global=true`. The standalone prose contradicts the bullet. Will rewrite the paragraph to describe both broadening modes (`scopes` for named siblings, `global=true` for unknown locations) and extend the `prompts_teach_cross_project_search_strategy` test to assert no contradictory phrasing.

Merging now.

PR #57, opened 2026-05-29
feat(links): cross-project wiki links + dependency graph

djalmajr:
## What
Wiki links can name another project, turning the per-project wikis into one dependency graph:

- `[[project:path.md]]` — a sibling project in the same workspace
- `[[workspace/project:path.md]]` — a project in another workspace

Bare links are unchanged (resolve within the source's own project). On top of the link primitive:
- **lint** reports dangling cross-project refs (typo'd project vs missing/renamed target page)
- **briefing** exposes `cross_project_dependents` / `cross_project_dependencies` per project
- **`GET /api/v1/graph`** returns the resolved cross-project edges for a graph view

## Why
Projects depend on each other (an app on shared infra/ops notes), but there was no way to make that explicit. Once it's an edge, the dependency is visible (graph, briefing) and breakage is catchable (lint).

## Design
- `links.to_page_id` is already a global id and `page_links` joins by id without a project filter — so a resolved cross-project link surfaces as a backlink for free. This wires authoring + resolution + display: `LinkTarget { workspace, project, path }`, a scope-aware resolver, deferred re-resolution across projects when a target lands later, and `RelatedPage` carrying the source's workspace/project.
- Migration rebuilds `links` with `to_workspace`/`to_project` in the PK. The migration version is provisional (V12 in our tree) — renumber to the next free upstream version on merge.

## Verification
- `cargo fmt` / `clippy -D warnings` — clean
- `cargo test --workspace` — green except the ambient `commands::reset::tests::*` (alive-process abort; same on upstream/main)
- New tests cover parsing (bare / cross-project / cross-workspace / URL), cross-project resolution, and graph+briefing+lint off one fixture

### DIFF
--- a/crates/ai-memory-consolidate/src/lint.rs
+++ b/crates/ai-memory-consolidate/src/lint.rs
@@ -107,6 +107,38 @@ pub async fn run_lint(
     let candidates = reader.decay_candidates(workspace_id, project_id).await?;
     let mut findings = rule_based_findings(&candidates);
 
+    // Dangling cross-project links: a `[[project:path]]` dependency that does
+    // not resolve. A broken inter-project edge is high-signal — surface it
+    // even on the zero-LLM path.
+    for dangling in reader
+        .dangling_cross_project_links(workspace_id, project_id)
+        .await?
+    {
+        let target = match &dangling.workspace {
+            Some(ws) => format!("{ws}/{}:{}", dangling.project, dangling.path),
+            None => format!("{}:{}", dangling.project, dangling.path),
+        };
+        let message = if dangling.project_exists {
+            format!(
+                "Page {} links to {} but that page does not exist in project `{}` \
+                 (missing, renamed, or deleted) — a broken cross-project dependency",
+                dangling.from_path, target, dangling.project,
+            )
+        } else {
+            format!(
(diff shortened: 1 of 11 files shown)


akitaonrails:
Solid design: making cross-project links an explicit edge (scope in the PK, `to_page_id` still global, `page_links` already joins by id without project filter) is the minimal change that earns the dep graph + briefing degree + lint dangling-ref check.

Migration audit — V12 rebuild is safe: existing rows project cleanly into `(to_workspace=NULL, to_project=NULL)` which is the same "local" semantics as before. PK widening is the minimum needed for `[[project:path]]` collision-free storage. Parser correctly excludes URL schemes (`://`, `mailto:`, etc.) so they're not parsed as `https://example.com` → project "https". URL/wikilink tests cover the regression surface.

Renumbering V12 → V13 on the way in (since #58 took V12). Will let the gate catch any unforeseen interaction with #56's `global=true` or #59's `write_target_ids` since all three landed today.

Merging now.

akitaonrails:
Merged in 667e4aa (local merge so V12 → V13 renumber lands in same commit — #58 took V12). 626/626 tests pass (+4 from this PR's parser regression suite). Thanks!

PR #58, opened 2026-05-29
feat(search): relevance-ranked recall — OR-join queries + accent-insensitive FTS

djalmajr:
## The problem
`prepare_fts5_query` joins query tokens with a space, which FTS5 reads as **AND**: every word must co-occur in one page. For a recall system that's a sharp recall cliff, and it fails silently — the caller gets zero rows and concludes "never recorded."

Concrete failures with the current behavior:

- **Natural-language queries return nothing.** An agent asking *"have we discussed cross project search strategy"* runs `MATCH 'cross project search strategy'` → AND → **0 hits**, even when a page matches 3 of the 4 words. Single-keyword queries work, which masks the bug; multi-word queries (how agents actually phrase recall) silently miss.
- **Portuguese / accents.** `"descricao da sessao"` never matches stored `"descrição da sessão"` — the `unicode61` tokenizer keeps diacritics, so an accent-free query (common when typing fast or from logs) finds nothing.

Both were hit in real usage: multi-term recall queries that *should* have matched returned empty, and the failure was invisible.

## The change
1. **OR-join bare tokens** (with bm25 ranking the callers already apply via `ORDER BY rank`). A query for N words now matches pages containing *any* of them, ranked by relevance — pages matching more (and rarer) terms rank first, so the top results are still the "AND-ish" matches; OR just adds a relevant tail instead of returning nothing. Explicit FTS5 syntax (`OR`/`AND`/`NOT`/`NEAR`, quoted phrases, parens) is detected and preserved verbatim — so exact-match is still available as an escape hatch.
2. **Accent-insensitive FTS** — rebuild the FTS tables (+ sync triggers) with `unicode61 remove_diacritics 2`, so `"descricao"` matches `"descrição"`. Contentless FTS, so source rows are untouched. (Migration version is provisional — renumber to the next free upstream version on merge.)

## Why OR is the right default (not AND)
For retrieval, recall + ranking beats precision-by-exclusion: bm25 already orders by relevance, so OR doesn't flood — it ranks. AND-default optimizes for "all words present," which is the surprising choice for a search box; every mainstream engine defaults to OR + ranking. Callers who want strict AND keep it via explicit operators / quotes.

## Verification
- `cargo fmt` / `clippy -D warnings` — clean
- `cargo test --workspace` — green except the ambient `commands::reset::tests::*`
- New: `bare_multi_word_is_or_joined`, `portuguese_accented_terms_or_join_and_keep_accents`, `single_word_has_no_or`, `search_is_accent_insensitive`

### DIFF
--- a/crates/ai-memory-store/migrations/V13__fts_remove_diacritics.sql
+++ b/crates/ai-memory-store/migrations/V13__fts_remove_diacritics.sql
@@ -1,0 +1,70 @@
+-- Accent-insensitive full-text search (Portuguese-friendly).
+--
+-- The FTS5 tables were created with `unicode61 tokenchars '/_-'`, which keeps
+-- diacritics: a query for "descricao" never matches stored "descrição". The
+-- tokenizer is fixed at CREATE time, so the only way to change it is to drop
+-- and recreate the FTS table (+ its sync triggers) and rebuild from the
+-- content table. `remove_diacritics 2` folds diacritics across the full
+-- Unicode range, so "descricao"/"descrição", "sessao"/"sessão", etc. unify.
+--
+-- Both FTS tables are contentless (`content='pages'` / `content='observations'`),
+-- so the source rows are untouched; only the derived index is rebuilt.
+
+-- ── pages_fts ────────────────────────────────────────────────────────────
+DROP TRIGGER pages_fts_ai;
+DROP TRIGGER pages_fts_ad;
+DROP TRIGGER pages_fts_au;
+DROP TABLE pages_fts;
+
+CREATE VIRTUAL TABLE pages_fts USING fts5(
+    title, body,
+    content='pages',
+    content_rowid='rowid',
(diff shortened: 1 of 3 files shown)


akitaonrails:
Audit looks good — the explicit-syntax detection preserves `OR`/`AND`/`NOT`/`NEAR` operators, parens, and quoted phrases verbatim, so the escape hatch for exact-match callers is intact. V13's FTS5 rebuild is content-preserving (both FTS tables are `content='pages'`/`content='observations'`, source rows are untouched). One mechanical adjustment on merge: since #59 didn't take a migration slot, the next free version is V12 — renumbering V13 → V12 on the way in. Will also add a couple of unit tests covering explicit-syntax preservation since that's the load-bearing escape hatch.

Merging now.

akitaonrails:
Merged in c1cc610 (local merge so V13 → V12 renumber lands in the same commit — #59 didn't take a migration slot, V12 is the next free version). Tests green (617/617, +4 from this PR). Thanks!

PR #59, opened 2026-05-30
fix(mcp): write_page / lint / forget_sweep honor the session's project (and explicit scope)

djalmajr:
## The problem
Three MCP handlers resolve `(workspace, project)` in ways that ignore what the agent asked for or where the session is actually working:

- **`memory_write_page { project: "X" }` silently writes to the *current* project when X doesn't exist yet.** The target is resolved with `find_project` (find-only); a name not in the DB falls through to the session's active project. So an agent told "save this in the `infra` project" — where `infra` hasn't been created yet — lands the page in whatever project the hook last activated, with no error. The page just ends up in the wrong place.
- **`memory_lint` and `memory_forget_sweep` always target the static default project.** Both pass the server's *baked* `workspace_id`/`project_id` (fixed at construction from `--project`) to `run_lint`/`run_sweep`, and neither accepts a `project` argument — they ignore the hook-published active project entirely. In a namespaced multi-project server, lint and the retention sweep run against one hardcoded project regardless of the session, so e.g. a cross-project dangling-link audit can never reach the project that holds the broken link.

Both were hit in live use: a write meant for a fresh project polluted the current one, and lint kept returning nothing because it was auditing the wrong project.

## The change
- **New `write_target_ids`**, used only by `memory_write_page`: an explicit `project` (and optional `workspace`) resolves with **get-or-create**, so a write to a named project always lands there — creating it if needed — never silently in the current project. With no explicit project, the existing active-project-wins behavior is preserved.
- **`memory_lint` / `memory_forget_sweep`** now resolve through the same find-only path the read tools use (`effective_ids_for_read_args`) and accept optional `project` / `workspace` args. No explicit project → the session's active project (hook-published), then the baked default.
- Reads stay find-only (resolving a missing project must not create it); only the write path get-or-creates.

## Why split read vs write resolution
Read/lint/sweep and write have opposite requirements: a read of a missing project must *not* create it (find-only); a write to a named project must land there (get-or-create). Sharing one resolver is exactly what produced the silent misroute — they're now distinct.

## Verification
- `cargo fmt` / `clippy --workspace --all-targets -D warnings` — clean
- `cargo test --workspace` — green except the ambient `commands::reset::tests::*`
- New tests: `memory_write_page_creates_explicit_project` (write to a non-existent project creates it and does not leak into the current one); `memory_forget_sweep_targets_the_explicit_project` (an episodic page in a named project is a sweep candidate only when the sweep points there)

### DIFF
--- a/crates/ai-memory-mcp/src/server.rs
+++ b/crates/ai-memory-mcp/src/server.rs
@@ -354,9 +354,16 @@ struct WritePageArgs {
     #[serde(default)]
     pinned: bool,
     /// Project to write into. Omit to target the project you're currently
-    /// working in (resolved from recent hook activity). **Omit unless the user explicitly names a *different* project.**
+    /// working in (resolved from recent hook activity). When set to a name
+    /// that doesn't exist yet, the project is **created** — so writes always
+    /// land where you asked, never silently in the current project. **Omit
+    /// unless the user explicitly names a *different* project.**
     #[serde(default)]
     project: Option<String>,
+    /// Workspace to write into. Only honoured together with an explicit
+    /// `project`; created if it doesn't exist. Omit for the current workspace.
+    #[serde(default)]
+    workspace: Option<String>,
 }
 
 #[tool_router]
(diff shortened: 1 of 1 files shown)


akitaonrails:
Solid catch on all three bugs. The split between `write_target_ids` (get-or-create, write semantics) and `effective_ids` / `effective_ids_for_read_args` (find-only, read semantics) is exactly the right shape — a write to a named project must land there, never silently rewrite the current one. Reusing `effective_ids_for_read_args` for `lint` + `forget_sweep` keeps them consistent with every other read-shaped tool.

Verifying gates locally before merge.

PR #60, opened 2026-06-01
feat(admin): move-project — move a project into another workspace

djalmajr:
`rename-project` changes a project's name within a workspace, but there's no way to move a project **across** workspaces. This adds `move-project` (CLI + `POST /admin/move-project`).

## Use cases
- Consolidating workspaces: you can't rename workspace `A → B` when `B` already exists, so you move A's projects into B one at a time.
- Reorganizing: splitting deploy- or client-specific projects into their own workspace while leaving shared ones in the common one.

## Behaviour — two strategies, picked by the destination

The destination decides which path runs; the response reports it as `moved_via`.

### `"true-move"` — fresh destination (lossless, the common case)
When the destination workspace has **no** same-named project, the move is a low-level re-stamp:
- Re-stamp `workspace_id` across every domain table for the project in **one writer transaction**, keeping the same `project_id` (`projects`, `pages`, `sessions`, `observations`, `handoffs`, `audit_log`). `page_embeddings` and `links` are keyed by `page_id`, so they follow with no re-stamp.
- `fs::rename` the project dir `<wiki>/<from_ws>/<proj>` → `<wiki>/<to_ws>/<proj>` (atomic within one wiki root).
- SQL commits **before** the rename, so when the watcher sees the new-path events it re-derives the same `(ws, proj)` the DB already holds and the reindex is a sha256 no-op.

This is O(1) (one transaction + one rename), re-embeds nothing, and **preserves everything** — sessions, observations, handoffs and the full supersession history all travel with the project.

### `"copy-purge"` — destination already has a same-named project (merge)
Two distinct `project_id`s can't be re-stamped into one (it would collide on `UNIQUE (workspace_id, name)`), so the source's latest pages are copied into the existing destination project through the normal write path (sanitization, link re-resolution, FTS index, and — on deploy — the admission/git-mirror webhooks all fire). Each page's existing embedding is **carried over verbatim** instead of recomputed; the source vector is reused only when it matches the configured embedder's `{provider, model, dim}`, so the one-model-per-index property holds. **Copy-before-purge**: only after every page copies successfully is the source purged; any copy failure aborts before the purge, and an unreadable source page is skipped and also blocks the purge, so a fixed re-run is safe.

In the merge path only durable pages move — sessions/observations/handoffs and the SQLite supersession history are dropped with the source purge (page history lives in the wiki's git history). The true-move path has no such loss.

Common to both: a same-workspace move is rejected (points at `rename-project`); `--confirm` is required.

## Design question — answered
The earlier revision asked whether sessions/observations should be carried over or whether "durable pages only" was the right scope. The answer here is the dual-path above: the **fresh-destination move is now lossless** (sessions/observations/history preserved via re-stamp), and "durable pages only" remains the behaviour **only** for the merge case, where re-stamp is structurally impossible. This also makes the common move O(1) instead of O(pages) — no per-page re-embed or webhook.

## Operational caveat
Moving the project the **current session** is writing to will have the next lifecycle hook re-create the source under the old workspace. Point the repo's `.ai-memory.toml` at the destination workspace **before** moving a live project. Documented in `docs/lifecycle-ops.md`.

## Tests
Integration (`admin_move.rs`): fresh-dest true-move (same `project_id`, nothing purged, dir renamed), true-move preserves a session + observation, merge-into-existing still copy+purge, embedding carry-over, confirm gate, 404 on unknown source, same-workspace rejection. Unit (`ops.rs`): re-stamp moves all domain rows; name-collision rolls the transaction back leaving the source intact.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ priors are at the [bottom](#influences-and-prior-art).
   Server runs local (loopback) OR on a homelab box (LAN/VPN/cloud)
   with bearer-token auth.
 - **Thin-client CLI.** `ai-memory status`, `bootstrap`, `purge-project`,
-  `rename-project`, `lint`, `embed`, `forget-sweep`, `backup` are
+  `rename-project`, `move-project`, `lint`, `embed`, `forget-sweep`, `backup` are
   all HTTP clients of the running server - never touch SQLite or
   wiki files directly. `status` also reports passive LLM/embedding
   provider health from the last real provider call. Server is the
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -96,11 +96,12 @@ pub enum Command {
     /// Useful after renaming the project's directory on disk so the hook
     /// router keeps writing into the same logical project.
     RenameProject(RenameProjectArgs),
-    /// Move a project into another workspace. Copies every latest page into
-    /// the destination (a NEW project, or MERGES into an existing same-named
-    /// one) through the normal write path, then PURGES the source. The
-    /// source-side purge is irreversible — requires `--confirm`. Note:
-    /// sessions/observations/handoffs do NOT migrate (only durable pages).
+    /// Move a project into another workspace. A fresh destination is a
+    /// lossless TRUE MOVE (re-stamp workspace_id, keep project_id, rename the
+    /// dir) — sessions/observations/handoffs and history all survive. A
+    /// destination that already holds a same-named project MERGES via
+    /// copy+purge (only durable pages migrate, source purged). Either way the
+    /// operation is irreversible — requires `--confirm`.
     MoveProject(MoveProjectArgs),
     /// Remove ai-memory's wiring (hooks, MCP, instructions) from all
     /// detected agents. Dry-run unless `--apply`.
@@ -360,8 +361,8 @@ pub struct MoveProjectArgs {
     /// Destination workspace. Auto-created if it doesn't exist.
     #[arg(long)]
     pub to_workspace: String,
-    /// REQUIRED — the move PURGES the source project after copying.
-    /// Without this flag the CLI errors out.
+    /// REQUIRED — the move re-stamps (true-move) or copies+purges (merge) the
+    /// source, both irreversible. Without this flag the CLI errors out.
     #[arg(long)]
     pub confirm: bool,
 }
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -96,6 +96,12 @@ pub enum Command {
     /// Useful after renaming the project's directory on disk so the hook
     /// router keeps writing into the same logical project.
     RenameProject(RenameProjectArgs),
+    /// Move a project into another workspace. Copies every latest page into
+    /// the destination (a NEW project, or MERGES into an existing same-named
+    /// one) through the normal write path, then PURGES the source. The
+    /// source-side purge is irreversible — requires `--confirm`. Note:
+    /// sessions/observations/handoffs do NOT migrate (only durable pages).
+    MoveProject(MoveProjectArgs),
     /// Remove ai-memory's wiring (hooks, MCP, instructions) from all
     /// detected agents. Dry-run unless `--apply`.
     Uninstall(UninstallArgs),
@@ -339,6 +345,25 @@ pub struct RenameProjectArgs {
     /// New project name. Must be non-empty and contain no slashes.
     #[arg(long)]
     pub to: String,
+}
+
+/// Arguments for `move-project`.
+#[derive(Debug, Args)]
+pub struct MoveProjectArgs {
+    /// Source workspace. Defaults to `default`.
+    #[arg(long, default_value_t = crate::config::DEFAULT_WORKSPACE.to_string())]
+    pub from_workspace: String,
+    /// Project name to move. When omitted, auto-derived from the basename
+    /// of the current git repo root (or CWD if no git repo).
+    #[arg(long)]
+    pub project: Option<String>,
+    /// Destination workspace. Auto-created if it doesn't exist.
+    #[arg(long)]
+    pub to_workspace: String,
+    /// REQUIRED — the move PURGES the source project after copying.
+    /// Without this flag the CLI errors out.
+    #[arg(long)]
+    pub confirm: bool,
 }
 
 /// Arguments for `install-instructions`.
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -371,6 +371,13 @@ pub struct MoveProjectArgs {
     /// and the schema rejects any stale write).
     #[arg(long)]
     pub force: bool,
+    /// Merge conflict policy (copy-purge path only): what to do when a source
+    /// page's path already exists in the destination with different content.
+    /// `block` (default) aborts and lists the conflicts; `overwrite` lets the
+    /// source supersede the destination page; `duplicate` keeps both (source
+    /// lands under a de-duplicated path).
+    #[arg(long, value_parser = ["block", "overwrite", "duplicate"], default_value = "block")]
+    pub on_conflict: String,
 }
 
 /// Arguments for `install-instructions`.
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -365,6 +365,12 @@ pub struct MoveProjectArgs {
     /// source, both irreversible. Without this flag the CLI errors out.
     #[arg(long)]
     pub confirm: bool,
+    /// Override the live-session guard. By default the server refuses (409) to
+    /// move the project a hook session is actively writing to; `--force`
+    /// proceeds anyway (still safe — the move keeps the active pointer correct
+    /// and the schema rejects any stale write).
+    #[arg(long)]
+    pub force: bool,
 }
 
 /// Arguments for `install-instructions`.
--- a/crates/ai-memory-cli/src/commands/mod.rs
+++ b/crates/ai-memory-cli/src/commands/mod.rs
@@ -19,6 +19,7 @@ pub mod install_instructions;
 pub mod install_mcp;
 pub mod lint;
 pub mod llm_test;
+pub mod move_project;
 pub mod openclaw_plugin;
 pub mod purge_project;
 pub mod read_page;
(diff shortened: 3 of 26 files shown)


akitaonrails:
Thanks @djalmajr. This is a valuable feature and the overall direction is worth keeping. Cross-workspace moves are a real gap: they help consolidate accidental workspaces, split client/deploy-specific projects, and clean up bad routing. The fresh-destination true-move path is especially useful because it can preserve `project_id`, sessions, observations, handoffs, page history, and embeddings without an O(pages) copy/re-embed pass.

Please rework this before we can merge it. The main issues we found are:

- The branch is stale. It is based before the fixes from #61, #62, and #63, so merging it as-is conflicts with or regresses current `main`, including the read-page workspace/fallback fixes and the FTS path-search migration/tokenizer fixes.
- The true-move path commits SQLite before renaming the wiki directory. If `fs::rename` fails, the DB says the project moved but files remain under the source workspace. Returning a manual-repair message is not enough for a destructive admin operation; we need a safer failure model or a compensating/reconcile path.
- Live hook routing/cache invariants are unsafe. The hook router caches `(workspace_id, project_id)` per cwd and republishes that active project on cache hits. After a true move, the same `project_id` belongs to a new workspace, but cached hooks can keep writing observations/logs with the old workspace id. Since the schema does not enforce `(project_id belongs to workspace_id)`, this can silently create split-brain rows instead of failing cleanly.
- Moving the project that the current session is writing to needs more than documentation. The PR documents the caveat, but the server still allows a live move that the next hook can partially undo or corrupt via stale cache/active-project state.
- The copy-purge merge path is not atomic. It copies pages one at a time, then purges. If a later page is unreadable, the source is left intact, but the destination already contains partial copied/superseded pages. That may be acceptable only if it is explicitly reported as partial, idempotent, and tested as such.
- Destination path conflicts are under-specified. If source and destination both have the same page path, the merge path supersedes the destination page. Please define whether source wins, destination wins, or conflict blocks the move, then report and test it.
- The embedding carry-over test does not exercise the copy-purge embedding code. The test moves into a fresh destination, so it takes the true-move path where embeddings follow automatically by `page_id`. Please add a destination-existing merge test that proves `without_embedder()` plus `store_embedding()` carries the vector onto the new destination page id.
- The docs and CLI text are stale/misleading in places. For example, the CLI warning says move-project copies pages then purges, but true-move does neither. The test module header also still says sessions/observations/handoffs are not migrated, which is false for true-move.
- The merge/rebase needs to preserve the current `CHANGELOG.md`, `MEMORY_INSTRUCTIONS`/routing snippet changes, `docs/ARCHITECTURE.md`, read-page admin tests, and V17 FTS tokenizer/trigger behavior from current `main`.

What I would like to see in the next revision:

- Rebase onto current `main`.
- Make true-move safe with respect to hook cache/active-project state and live sessions.
- Avoid or repair DB/filesystem split-brain when the directory rename fails.
- Add real copy-purge tests for embedding carry-over, same-path conflicts, and partial-copy reporting.
- Tighten the CLI/docs/test comments so they accurately describe true-move versus copy-purge.
- Run the full local gate after the rework: `cargo fmt --check`, `git diff --check`, `TAILWIND_SKIP=1 cargo test --workspace`, and `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings`.

Thanks again for pushing this. The feature is useful, but it touches destructive data movement and live hook routing, so we need these invariants nailed down before merge.

djalmajr:
Thanks for the thorough review — reworked and pushed. How each point was addressed:

**Rebased onto current `main`** — the branch now sits on top of #61/#62/#63. Only `docs/ARCHITECTURE.md` is touched (additive: adds `move-project` to the command list); CHANGELOG, the routing snippet, the read-page admin tests, and the V17 FTS tokenizer/trigger are untouched.

**True-move failure model** — the move is now all-or-nothing. The destination dir is pre-checked absent; if `fs::rename` fails after the SQL re-stamp committed, the re-stamp is rolled back (the op is symmetric — `move_project_workspace(proj, dst_ws, src_ws)`) and the call returns `move aborted (nothing changed)`. A double fault (rollback also fails) surfaces a precise manual-repair message. No DB-ahead-of-disk state on the normal failure path.

**Hook cache / `(workspace_id, project_id)` invariant** — added a migration with `BEFORE INSERT` triggers on pages/sessions/observations/handoffs that ABORT when `workspace_id` disagrees with the project's workspace. A stale cached write now fails cleanly instead of silently creating a split-brain row, and the router's existing evict-and-re-resolve path recovers. INSERT-only, so the move's own re-stamp (UPDATE) is unaffected.

**Live-session move** — the handler now refuses (409) to move the project the hook router has published as the active project; `--force` / `force: true` overrides. On any successful move the active pointer is republished to the destination (true-move) or cleared (copy-purge), so the next hook resolves cleanly.

**copy-purge partial / idempotent** — partial copies are explicitly reported (`pages_skipped` + `source_purged: false`), and a fixed re-run is idempotent (identical content is a no-op supersession). Tested.

**Same-path conflicts** — defined as "duplicate, keep both": when a source page's path already exists at the destination with different content, the source page lands under a de-duplicated path (`<stem>-from-<src_workspace>.md`, then `-2`, …) and the remap is reported in a `conflicts` array. This is a destructive data-movement operation, so the default avoids silently overwriting either side — neither the destination's page nor the source's is lost. Identical content stays a no-op supersession at the same path; the lossless `true-move` path remains the way to preserve paths and links.

**Tests** — added copy-purge embedding carry-over (destination-existing project, proves `without_embedder()` + `store_embedding()` carries the vector onto the new page id), same-path conflict duplication, partial-copy reporting, idempotent re-run, plus rollback-on-rename-failure and the active-project guard/force.

**Docs / CLI / comments** — the CLI pre-confirm warning and the `admin_move.rs` module header now describe true-move vs copy-purge correctly; `docs/lifecycle-ops.md` documents the failure model, the live-session guard, and conflict duplication.

**Gate** — `cargo fmt --check`, `git diff --check`, `TAILWIND_SKIP=1 cargo test --workspace`, and `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings` all clean.

djalmajr:
**Follow-up after further hardening (pushed).** A few things changed since the summary above:

- **Same-path conflicts are now an explicit `on_conflict` policy with `block` as the default** (the earlier revision defaulted to "duplicate"). This addresses the source-wins / dest-wins / block point directly: `block` aborts the whole merge with 409, lists the conflicting paths, and leaves the source intact — the safe default for a destructive op. `overwrite` (source supersedes the destination page) and `duplicate` (keep both, under a de-duplicated path) are opt-in via `--on-conflict`, each reported in the `conflicts` array.

- **True-move is now rename-FIRST, SQL-commit-LAST**, so the DB is never ahead of disk: a rename failure touches nothing; a SQL failure renames the dir back; a crash between the two steps leaves at most an orphan dir at the destination with the DB still wholly at the source (recoverable), never a row pointing at a missing file. Any watcher reindex during that window trips the pairing trigger and is rejected cleanly.

- **The hook-cache recovery is now covered by an integration test, not just reasoning.** A cached project moved to another workspace makes the next hook's stale write trip the `(workspace_id, project_id)` trigger; the router evicts + re-resolves into a consistent pair instead of writing a split-brain row (the test only passes with the trigger present). The move also proactively evicts the router's per-cwd cache for the moved project.

- Validated end-to-end against a real server in a container (true-move, block/overwrite/duplicate, partial-skip, and the live-session guard exercised through an actual `/hook` event), plus a `duplicate`-mode edge case where a source page's natural path collided with another page's de-duplicated slot.

The other points from the previous comment (rebase onto `main`, the live-session guard, copy-purge embedding carry-over, and the full gate) are unchanged.

akitaonrails:
Re-reviewed latest head `806113f` after #55 landed. The focused happy-path tests pass:
- `TAILWIND_SKIP=1 cargo test -p ai-memory-mcp --test admin_move`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-store move_project`

I still do not think this is safe to merge as-is.

Findings:

1. High: true-move has a race that can orphan concurrent writes. `true_move_project` renames the directory before the DB restamp (`crates/ai-memory-mcp/src/admin.rs:1754`, `:1768`). `Wiki::write_page` writes the file before the DB upsert (`crates/ai-memory-wiki/src/wiki.rs:628`, `:637-650`). A concurrent explicit MCP/admin/CLI write using the old `(src_ws, project_id)` during that window can recreate the source dir and leave a file there while the DB row is moved to the destination, or fail the V18 trigger after the file is already written. The active-project guard only covers the published hook active project, not every explicit writer. This needs a move/quiesce lock or a different protocol, plus a controlled concurrent-write regression.

2. High: fresh-destination true-move bypasses the #55 admission/git-mirror contract. The true-move path calls `Wiki::rename_project_dir` directly (`crates/ai-memory-mcp/src/admin.rs:1754`; `crates/ai-memory-wiki/src/wiki.rs:189-205`) and the comment explicitly says no admission webhook (`crates/ai-memory-mcp/src/admin.rs:1960-1964`). Copy-purge notifies via write/purge, but true-move silently moves the wiki tree from an external mirror perspective. A configured admission/git-mirror webhook will not learn to remove the old path or materialize the new one. Add a move admission op or another explicit notification strategy, with tests.

3. Medium: `on_conflict=block` only compares body text, but copy then writes source metadata/title/tier/pinned over the destination when bodies are equal (`crates/ai-memory-mcp/src/admin.rs:1999-2005`, `:2147-2157`). `upsert_page` can supersede on metadata/frontmatter changes, so the safe default can still mutate destination page metadata and then purge the source. Either compare full page identity for conflict detection or document/test this behavior explicitly.

4. Medium: V18 only prevents future bad inserts. It adds `BEFORE INSERT` triggers (`crates/ai-memory-store/migrations/V18__enforce_project_workspace_pairing.sql:19-49`) but does not detect existing split-brain rows before installing the invariant. For a migration meant to protect `(workspace_id, project_id)` pairing, either fail with a diagnostic query, repair, or document that it is future-write-only and add coverage.

5. Low/docs: missing `CHANGELOG.md` for a new destructive CLI/admin operation. `docs/lifecycle-ops.md:18` still says “All five commands” after adding `move-project`, and the true-move sequence is internally inconsistent (`docs/lifecycle-ops.md:142-150` says restamp then rename while also saying rename-first; `:221-222` describes a restamp-committed/rename-failed mode that the current code is trying to avoid).

The rework is much closer than the previous revision, but the true-move concurrency and admission semantics are still load-bearing blockers.

akitaonrails:
Merged into main with follow-up hardening in `f7374bf` and merge commit `0772a38`. Hardened true-move with a wiki mutation gate + pre-disk pair validation, added `move_project` admission notifications, made copy-purge conflict detection compare full page identity, and added V18 split-brain preflight. Validation passed: `cargo fmt --check`, `git diff --check`, `TAILWIND_SKIP=1 cargo test --workspace`, and `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings`.

PR #61, opened 2026-06-01
fix(mcp): write_target_ids defaults workspace to the active project's

djalmajr:
`memory_write_page` with an explicit `project` but no `workspace` resolved the workspace to the server's baked `--workspace` (e.g. `default`), ignoring the hook-published `ActiveProject`.

So from a cwd routed (via `.ai-memory.toml`) to a non-default workspace, a write like `{ project: "foo" }` would silently land in — and create — `default/foo` instead of `<cwd-workspace>/foo`. Reads (`effective_ids`) already consult `ActiveProject`; writes didn't, for the workspace dimension.

This defaults the workspace to the `ActiveProject`'s workspace when none is given, falling back to the baked default only when no `ActiveProject` has been published yet (early startup / no hooks). Passing `workspace` explicitly still targets the baked/shared workspace on purpose.

Adds `write_target_ids_defaults_workspace_to_active_project`.

### DIFF
--- a/crates/ai-memory-mcp/src/server.rs
+++ b/crates/ai-memory-mcp/src/server.rs
@@ -534,6 +534,15 @@ impl AiMemoryServer {
     /// (find-only, for reads): a write to a named project must land there, not
     /// silently fall back to the current project. With no explicit `project`,
     /// the active-project-wins behaviour is preserved (issue #2).
+    ///
+    /// When an explicit `project` is given but **no** explicit `workspace`, the
+    /// workspace defaults to the hook-published [`ActiveProject`]'s workspace
+    /// (the cwd the agent is working in) — NOT the server's baked `--workspace`.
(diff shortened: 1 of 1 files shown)


akitaonrails:
Thanks @djalmajr. I brought this into main because the direction was right: project-only writes should use the hook-published active workspace.

What was missed: the read side still needed to follow the same active-workspace resolution, otherwise a project-only write could land in the active workspace and a later project-only read could miss it. I added that follow-up and a regression test.

Landed in main:
- merge: https://github.com/akitaonrails/ai-memory/commit/0aa0f880e1dbcdaf9b1c0b64b7fe47fbd8a9bb1c
- follow-up fix: https://github.com/akitaonrails/ai-memory/commit/524da829262f20d58fd2dc284435156c0f4a9800

PR #62, opened 2026-06-01
feat(store): index page path/slug in FTS so filename searches hit

djalmajr:
## Problem
Search (`memory_query`, `/admin/search`, and `read_page` query-mode) FTS5-matches only a page's **title + body**. A distinctive slug/filename like `followup-bulk-rename-runbook-titles` is missed unless those words also appear in the prose — so you can't reliably find a page by its path.

## Fix
Migration **V17** adds a `pages.path_search` column holding the path normalised for search, and includes it in the `pages_fts` index. The FTS stays **content-backed** (external content; no body duplication) — `path_search` is a real column the writer keeps in sync.

The path is indexed in **both forms** so either query style hits:
- **segments** — `/` and `.` → space, *keeping* `-`/`_` (FTS token chars), so the whole hyphenated slug stays one token (`"foo-bar"` matches);
- **words** — also split `-`/`_`, so a single word (`bar`) matches.

`notes/foo-bar.md` → `notes foo-bar md notes foo bar md`.

`ops::path_search_text` (live writes) is kept **byte-identical** to the V17 backfill SQL, so live-indexed and `rebuild`-indexed text agree (same bm25 term frequencies, not just the same match set).

No reader SQL change: FTS `MATCH` already searches every column, and `snippet()` still targets body (column 1).

## Tests
- `path_search_text_indexes_slug_and_words` — the normaliser output.
- `fts_matches_page_by_path_slug_not_in_body` — a page whose slug is in neither title nor body is found by the quoted slug, by a single path word, and the body stays searchable (regression).
- Full `cargo test -p ai-memory-store` green (migration applies + rebuilds cleanly); mcp integration tests green.

### DIFF
--- a/crates/ai-memory-store/migrations/V17__pages_fts_index_path.sql
+++ b/crates/ai-memory-store/migrations/V17__pages_fts_index_path.sql
@@ -1,0 +1,57 @@
+-- Index the page PATH (slug/filename) in the FTS table so a search for a
+-- distinctive slug like `followup-bulk-rename-runbook-titles` finds the page
+-- even when the slug never appears in the prose body or title.
+--
+-- The FTS tokenizer keeps `/`, `_`, `-` as token characters (so identifiers
+-- stay whole), which means a raw path `notes/foo-bar.md` tokenizes as the
+-- single glued token `notes/foo-bar` plus `md`. We index a NORMALISED copy of
+-- the path in BOTH forms so either query style hits:
+--   * segments — `/` and `.` → space, KEEPING `-`/`_` (so the whole hyphenated
+--     slug stays one token: `"foo-bar"` matches);
+--   * words — also split `-`/`_` (so `bar` alone matches).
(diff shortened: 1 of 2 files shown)


akitaonrails:
Thanks @djalmajr. I brought this into main because indexing normalized page paths in FTS is the right fix for filename and slug searches.

What was missed: the migration recreated the FTS tokenizer without `remove_diacritics 2`, which would regress accent-insensitive search, and the update trigger would reindex FTS on access-counter-only updates. I added the follow-up to preserve tokenizer behavior, narrow the trigger, and cover both with tests.

Landed in main:
- merge: https://github.com/akitaonrails/ai-memory/commit/835914969ad88b5e496836f205bd5eec9de79b23
- follow-up fix: https://github.com/akitaonrails/ai-memory/commit/dd2b8d0839d7b69203639de825637cc5f08d5589

PR #63, opened 2026-06-01
fix(mcp): make read_page reliable — cross-workspace scope + DB fallback

djalmajr:
Fixes the two symptoms in the long-standing `read-page-by-query-misses` report: `memory_read_page` (and `/admin/read-page`) could fail to return a page that search finds fine.

## 1. Cross-workspace scope
`memory_read_page` resolved via `effective_ids(project)` only — it had **no `workspace` arg**, so a page in a *different* workspace (a sibling project on a shared server) was unreachable, and an explicit `project` from another workspace silently fell back to the active project (→ reading the wrong on-disk path → `os error 2`).

It now takes an optional `workspace` and resolves through `effective_ids_for_read_args` — the **same chain `memory_query` uses** — so explicit `workspace` + `project` reaches the right page.

## 2. `os error 2` on a page that exists
The on-disk markdown read could fail when the index is momentarily ahead of disk (recently-written page; watcher/disk skew), returning *"No such file or directory"* for a page that's clearly in search results.

Both handlers now **fall back to the store's faithful copy** of the body (new `ReaderPool::page_body_by_ids`) before erroring, and only 404 when there's genuinely no `is_latest` row. The MCP response tags a fallback read with `"served_from":"db-fallback"`.

Markdown stays the source of truth; the DB copy is written in the **same transaction** as the page, so serving it is safe.

## Tests
`admin_read_page.rs`:
- DB fallback serves a page that's in the store but not on disk;
- on-disk happy path still reads from disk;
- a genuinely-absent page still 404s (the fallback doesn't mask real misses).

Full `cargo test -p ai-memory-mcp -p ai-memory-store` green.

### DIFF
--- a/crates/ai-memory-mcp/src/admin.rs
+++ b/crates/ai-memory-mcp/src/admin.rs
@@ -461,10 +461,40 @@ async fn handle_read_page(
                 Json(serde_json::to_value(&resp).unwrap_or_else(|_| serde_json::json!({}))),
             )
         }
-        Err(e) => (
-            StatusCode::NOT_FOUND,
-            Json(serde_json::json!({ "error": e.to_string() })),
-        ),
+        // On-disk read failed — fall back to the DB's faithful copy before
+        // 404ing (index/disk skew; see gotchas/read-page-by-query-misses).
+        Err(disk_err) => match state
+            .reader
+            .page_body_by_ids(ws, proj, page_path.as_str())
+            .await
+        {
+            Ok(Some(stored)) => {
+                let frontmatter: serde_json::Value = serde_json::from_str(&stored.frontmatter_json)
+                    .unwrap_or(serde_json::Value::Null);
+                let title = frontmatter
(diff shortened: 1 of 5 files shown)


akitaonrails:
Thanks @djalmajr. I brought this into main because explicit `workspace` + `project` reads and a missing-file DB fallback make `memory_read_page` much more reliable.

What was missed: the fallback caught every wiki read error, which would hide malformed frontmatter, permission problems, or corrupted markdown even though disk is the source of truth. The prompt/docs and tests also needed to describe the cross-workspace contract and `served_from` fallback diagnostic.

Landed in main:
- merge: https://github.com/akitaonrails/ai-memory/commit/6621459e66697229226b7a2d3c4c3957ac74f7fe
- follow-up fix: https://github.com/akitaonrails/ai-memory/commit/d3538c5826388faa0c7ffe54b17de49a22388181

PR #64, opened 2026-06-01
fix(llm): support streamed Codex responses for openai-oauth

rikelmyso7:
## What changed

Fixed the `openai-oauth` provider so ChatGPT/Codex-backed requests use the current responses streaming contract. The provider now requests `text/event-stream` when `stream=true`, reconstructs the final response from SSE events such as `response.output_text.delta` and `response.completed`, and stops sending the unsupported `max_output_tokens` field on this endpoint. Added focused regression tests for the request shape and SSE parsing, and documented the fix in `CHANGELOG.md`.

## Why

`ai-memory bootstrap` was failing against `openai-oauth` even after successful ChatGPT/Codex login. The first provider error was:

- `Stream must be set to true`

After switching to streamed responses, the next provider error was:

- `Unsupported parameter: max_output_tokens`

This left a supported provider path broken for real bootstrap/consolidation flows.

## Test plan

- [x] `cargo test --workspace --all-targets` passes
- [x] Manual test:
  - Ran `ai-memory auth login openai-oauth`
  - Started server with `AI_MEMORY_LLM_PROVIDER=openai-oauth`
  - Reproduced bootstrap failure with the upstream implementation
  - Verified the patched build completes `ai-memory bootstrap` successfully against a real ChatGPT Pro/Plus OAuth session

## Notes for reviewers

This is intentionally scoped to `openai_oauth.rs` plus a changelog entry. No provider factory, CLI, or config behavior changed. The SSE parser only consumes the events needed to reconstruct the completed response and surfaces provider-side failure events as `LlmError::Provider`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `memory_read_page` now accepts explicit `workspace` + `project` for sibling
   projects and falls back to the stored DB body only when the markdown file is
   missing, not when the disk source of truth is corrupt or unreadable ([#63]).
+- `openai-oauth` now speaks the current ChatGPT/Codex responses stream format
+  for bootstrap/consolidation requests and avoids sending the unsupported
+  `max_output_tokens` field on that endpoint.
 
 ## [0.8.1] - 2026-05-30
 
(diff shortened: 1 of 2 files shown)


akitaonrails:
Merged on main at a557416.

Validation after merge plus the #66 follow-up fix:
- cargo fmt --check
- git diff --check

Thanks for the focused provider fix.

akitaonrails:
Follow-up hardening landed on main at 78c121b after the merge: the openai-oauth SSE parser now treats `response.incomplete` and `response.cancelled` as provider failures, with regression tests for terminal failure events and streams that end without `response.completed`.

Validation after the follow-up:
- cargo fmt --check
- git diff --check

PR #65, opened 2026-06-02
feat(serve): host the HTTP surface under a configurable base path

djalmajr:
## Why

ai-memory currently mounts every route at the host root (`/mcp`, `/api/v1`, `/hook`, `/web`). That's perfect when it owns a hostname, but homelab/reverse-proxy setups often want to host it under a **URL subpath** of an existing host — e.g. `https://admin.home/wiki` alongside other services on `admin.home`. Without prefix awareness the SPA's absolute `/api/v1` calls and `/web` asset URLs collide with whatever else lives at that host's root.

## What

Adds an **opt-in** `--base-path` / `AI_MEMORY_BASE_PATH` to `serve`:

- **Empty (default) → byte-identical to today.** Every route stays at root; no behavior change for existing deploys.
- Set e.g. `/wiki` → the whole HTTP surface nests under it: `/wiki/mcp`, `/wiki/api/v1`, `/wiki/hook`, and the web UI.
- The value is normalized to `/<core>` (leading slash, no trailing, collapsed slashes, safe charset); `/` and `` both mean root.
- The bundled web UI is made subpath-aware by injecting `<base href>` + a `<meta name="ai-memory-base-path">` into the served `index.html`, so the SPA builds its asset and `/api/v1` URLs under the prefix. (The companion SPA change reads that meta; the server-rendered fallback links are noted as a follow-up.)
- Also exposes `--web-slug` / `AI_MEMORY_WEB_SLUG` (default `/web`) so the UI mount point under the base is configurable.

Behind a reverse proxy you point the prefix at the engine and **preserve** it (no strip) — the engine now owns the prefix end-to-end.

## Compatibility

The feature is entirely opt-in and the default path is unchanged — verified that an unset `AI_MEMORY_BASE_PATH` renders the identical route tree as before. Diff is confined to `serve` (`cli.rs` + `commands/serve.rs`); no store/migration/MCP-protocol changes.

## Tests

- New `web_base_tests` module covering prefix normalization (root fallback, trailing/duplicate slash collapse, unsafe-char rejection), `<base href>` derivation, and head injection.
- Existing serve tests updated for the new mount signature; full suite green.

## Notes / follow-ups

- Both frontends are now subpath-aware: the custom SPA (injected `<base href>` + meta) and the built-in server-rendered browser (relative URLs + injected `<base href>` on every HTML response).
- Validated live in a homelab behind nginx-ingress at `https://<host>/wiki` (MCP + browser UI both reachable under the prefix).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- `ai-memory serve --transport http` can host the entire HTTP surface under a
+  configurable subpath with `--base-path` / `AI_MEMORY_BASE_PATH`; `/mcp`,
+  `/hook`, `/admin/*`, `/api/v1`, and the web UI all move under that prefix.
+  The web UI mount can also be changed with `--web-slug`, and custom
+  `--web-ui-dir` SPAs receive injected `<base href>` plus
+  `ai-memory-base-path` metadata for same-origin API calls behind reverse
+  proxies ([#65]).
 - `ai-memory move-project` can move projects across workspaces via the admin
   API. Fresh destinations use a lossless true move that keeps the same
   `project_id`, sessions, observations, handoffs, embeddings, and page history;
@@ -24,6 +31,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   before removal ([#55]).
 
 ### Fixed
+- Custom `--web-ui-dir` frontends no longer serve raw `/index.html` without
+  base-path injection; direct index requests and SPA fallback routes now return
+  the injected shell, while static assets remain untouched ([#65]).
 - `move-project` true moves now run through a wiki mutation gate: normal
   page writes/reindexes validate the `(workspace_id, project_id)` pair before
   touching disk, while true moves hold the exclusive side across the directory
--- a/README.md
+++ b/README.md
@@ -369,6 +369,13 @@ Useful entry points:
   screenshots and e2e tests — lives at
   [djalmajr/ai-memory-ui](https://github.com/djalmajr/ai-memory-ui).
 
+  When a reverse proxy hosts ai-memory under a URL subpath, set
+  `--base-path` (or `AI_MEMORY_BASE_PATH`) so every HTTP surface moves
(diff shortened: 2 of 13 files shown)


akitaonrails:
Draft review notes. I would not mark this ready yet.

Findings:

1. High: custom SPA `index.html` can bypass the base/meta injection. The code reads and injects one copy for the fallback (`crates/ai-memory-cli/src/commands/serve.rs:917-929`), but `ServeDir::new(dir)` still serves explicit files directly. A request to `{base}{web_slug}/index.html` will return the raw on-disk `index.html`, without `<base href>` or `ai-memory-base-path`, so the SPA can still call root `/api/v1` or load root assets under subpath deployments. Add a test for `/wiki/web/index.html` and route it through the same injected body.

2. Medium: missing `index.html` becomes a blank 200. `std::fs::read_to_string(dir.join("index.html")).unwrap_or_default()` (`serve.rs:917`) silently converts a bad `--web-ui-dir` into an empty HTML fallback. That should fail startup or return a clear 500/404; otherwise operators get a successful but blank UI.

3. Medium/docs: this is a user-visible deployment feature (`--base-path`, `AI_MEMORY_BASE_PATH`, `--web-slug`, `AI_MEMORY_WEB_SLUG`) but `CHANGELOG.md` has no entry and the reverse-proxy docs are not updated. Per project rules, installation/deployment/support-reference changes need changelog and README/docs coverage in the same commit.

4. Low: rustfmt is red. I reproduced `cargo fmt --check` in a temp worktree; it reports formatting diffs in `crates/ai-memory-cli/src/commands/serve.rs` and `crates/ai-memory-web/src/templates.rs`.

The route-nesting direction looks useful, and the added router tests cover the main prefix behavior. The custom-SPA explicit-index path and docs/changelog are the main things I would fix before review-ready.

akitaonrails:
Merged after maintainer hardening. Added startup error handling for custom index reads, explicit injected routes for custom SPA index/fallback handling, regression coverage for subpath direct index/client routes and `--web-slug /` API shadowing, plus README/changelog/proxy/frontend docs. Local gate passed: `cargo fmt --check`, `git diff --check`, `TAILWIND_SKIP=1 cargo test --workspace`, and `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings`.

PR #66, opened 2026-06-02
fix(cli): write-page and read-page resolve project identically (ghost write)

djalmajr:
## Problem

`write-page` and `read-page` resolved the target `(workspace, project)` through **different** code paths, so a write followed by a read-back could silently target two different projects — the classic "ghost write": the write returns a `page_id`, but the read-back reports the page does not exist.

- `read-page` / `search` use `resolve_project_name` — an explicit `--project` wins, otherwise the project is auto-detected from the current directory.
- `write-page` instead hard-defaulted `--project` to `"scratch"` (a clap default that masked "no value given") and **bypassed** the cwd auto-detect.

So from any directory not literally named `scratch`, with no explicit `--project`:
- the write landed in `default/scratch`,
- the read-back looked in `default/<cwd>` → `404 No such file`.

## Reproduction

In a local container (`serve` + CLI against it):

- **Before:** 60/60 ghosts. The write reported `under default/scratch`; the read 404'd against `default/<cwd>`; the file existed on disk under `scratch`.
- **After:** 0/30 ghosts; both write and read-back resolve to `default/<cwd>`.

## Fix

Make `WritePageArgs.project` an `Option<String>` (no default) and route it through the same `resolve_project_name` resolver the read path uses. Explicit `--project` still wins; with none, the write and the read-back that follows it now always resolve to the same project.

## Tests

All `ai-memory-cli` tests pass (178). No store/migration/protocol changes — the fix is confined to CLI project resolution.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -262,7 +262,7 @@ match the intent to the tool. They do not need to name the tool.
 
 | User says / situation | Tool |
 |---|---|
-| "have we discussed X?" / "search memory for Y" / before proposing architecture | `memory_query` |
+| "have we discussed X?" / "search memory for Y" / before proposing architecture | `memory_query` (current project; `scopes` for named siblings; `global=true` to search every project) |
 | "what's been going on" / "show recent activity" (light) | `memory_recent` |
 | "is ai-memory healthy?" / "how big is the wiki?" | `memory_status` |
 | "give me the stats" / structured snapshot for the agent to consume | `memory_briefing` |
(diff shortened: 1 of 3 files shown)


akitaonrails:
Landed the bug fix on main as f37f210 (`fix(cli): align write-page project resolution`) instead of merging this branch directly.

Reason: the PR branch is stale relative to current main and includes an unrelated `CLAUDE.md` routing refresh. I kept the functional change only, added `CHANGELOG.md`, and added parser coverage proving `write-page` no longer hard-defaults `--project` to `scratch`.

Validation:
- TAILWIND_SKIP=1 cargo test -p ai-memory-cli write_page_project_is_optional_for_shared_resolution
- TAILWIND_SKIP=1 cargo test -p ai-memory-cli resolve_project_name
- Full gate also passed after landing: cargo fmt --check, git diff --check, TAILWIND_SKIP=1 cargo test --workspace, TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings

PR #68, opened 2026-06-02
feat(web): render [[wikilinks]] as clickable internal links

djalmajr:
## What

The built-in server-rendered wiki browser left `[[wiki links]]` as literal text — only standard `[label](url)` markdown became clickable. This renders the wikilink forms the engine's link extractor (`ai_memory_wiki::extract_links`) already understands as clickable internal links:

- `[[notes/foo]]` → current workspace/project
- `[[notes/foo|label]]` → explicit display label
- `[[project:path]]` → sibling project, current workspace
- `[[workspace/project:path]]` → fully qualified

Targets resolve to the page route via `page_href`, so they inherit whatever prefix the mount uses (host root or a reverse-proxy subpath). Bare targets get a `.md` suffix; anchors/queries are stripped. External schemes, path traversal, and empty targets are left as literal `[[…]]`. Wikilinks inside fenced code blocks and inline code are not rewritten.

## How

`[[…]]` is rewritten into ordinary markdown links *before* parsing — pulldown-cmark consumes `[...]` as reference-link syntax, so source-level preprocessing is the robust hook. The pass skips fenced + inline code, mirroring the link extractor's fence handling.

## Tests

5 new unit tests (resolution against current project, label + `.md` suffix, cross-project/cross-workspace scope, code-block/inline-code skip, malformed/external/traversal kept literal). Full `ai-memory-web` suite + clippy green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- The read-only web browser now renders `[[wiki links]]` as clickable internal
+  links to the target page. Supports `[[path]]`, `[[path|label]]`,
+  `[[project:path]]`, and `[[workspace/project:path]]`, resolved against the
+  current page's project unless the target carries its own scope; bare targets
+  get a `.md` suffix. External schemes, path traversal, and links inside fenced
+  or inline code are left as literal text ([#68]).
 - `ai-memory move-project` can move projects across workspaces via the admin
   API. Fresh destinations use a lossless true move that keeps the same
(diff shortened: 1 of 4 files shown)


akitaonrails:
Merged after maintainer hardening. Resolved the post-#65 conflict, kept web links relative to the injected base href, and routed wikilink targets through the canonical PagePath validator so invalid/traversal/non-page targets remain literal. Local gate passed: `cargo fmt --check`, `git diff --check`, `TAILWIND_SKIP=1 cargo test -p ai-memory-web`, `TAILWIND_SKIP=1 cargo clippy -p ai-memory-web --all-targets -- -D warnings`, `TAILWIND_SKIP=1 cargo test --workspace`, and `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings`.

PR #86, opened 2026-06-08
feat: reindex-from-wiki — make 'DB is rebuildable from files' operable (_meta.md + reindex)

djalmajr:
## Make "the DB is rebuildable from files" operable

`design-decisions.md §3` chose markdown-in-git as the source of truth with SQLite
as a **derived index**, and states *"DB is rebuildable from files — corruption is
recoverable"*. Today nothing actually rebuilds it: zeroing `db/` and restarting does
**not** reconstruct pages, because the watcher's reconcile only *validates* scopes
(`ensure_project_workspace`) — it never creates the `(workspace, project)` rows, so
every reindex fails `NotFound`. And the human-readable workspace/project **names**
live only in the DB, not in the wiki tree.

This PR closes that gap with three cohesive pieces:

### 1. Self-describing wiki (`_meta.md`)
The engine writes a tiny per-scope manifest — `wiki/<ws-uuid>/_meta.md` (workspace
name) and `wiki/<ws-uuid>/<proj-uuid>/_meta.md` (project name + `repo_path`) — so the
markdown tree alone carries everything needed to rebuild. `Wiki::backfill_scope_manifests`
runs on `serve` startup, is **idempotent** (unchanged content is never rewritten, so it
doesn't churn the wiki git), and `_meta.md` is excluded from page indexing.

### 2. `ai-memory reindex`
A lifecycle command (server stopped, like `reset`/`restore`) that walks the wiki tree,
recreates workspaces/projects from each scope's `_meta.md` via the new
`ensure_{workspace,project}_with_id` store ops (preserving the ids the tree is keyed
by), and reindexes every page. Point it at a freshly-migrated `db/` to move a data dir
onto a clean migration lineage **without** carrying a divergent `refinery_schema_history`.

### 3. Page detection by content, not filename
The reserved-name skip (`log.md` / `log-YYYY-MM.md` / `bootstrap.md`) used a filename
match, which **silently dropped** a real page that happened to live at `log.md`. Now a
reserved-named file is skipped only when it lacks YAML frontmatter (i.e. it really is the
raw `## [..]` event ledger); a frontmatter file is a page and is indexed. `_meta.md` is
always skipped.

## Scope / non-goals
Only the markdown-derived state is rebuilt: pages, links, FTS (and embeddings can be
recomputed with `embed`). Episodic DB-only state — sessions, observations, handoffs,
decay counters — is **not** reconstructed; that is the "compile-don't-hoard" churn the
design already sheds.

## Tests
- `reindex_page_by_content_not_filename` — a frontmatter page named `log.md` is indexed;
  `_meta.md` and the raw ledger are not.
- `backfill_then_reindex_rebuilds_from_wiki_alone` — write pages (incl. one at `log.md`),
  backfill manifests, then `reindex_all` on a **fresh store** rebuilds the named scopes +
  all pages from the wiki tree alone.

Validated against a real 659-page instance: a backup's wiki rebuilds 659/659 pages
(including the `log.md`-named page), recovers workspace/project names + `repo_path`, and
produces a clean migration lineage with no operator-specific carry-over.


akitaonrails:
Thanks @djalmajr — this turned the rebuild-from-wiki design contract into an operator-safe lifecycle path. Merged as 4b8f4bfd4f3c3bea510fda0b9d55c7491252a6c9 after the audit hardening follow-up and green CI.

PR #98, opened 2026-06-14
fix(admin): propagate authenticated actor into purge/move admission

djalmajr:
## Problem

`/admin/purge-project` and `/admin/move-project` build their `AdmissionContext` with `..Default::default()`, so the `actor` field is always empty — even when the auth middleware resolved one. An admission webhook that authorizes by user (e.g. a per-user `scope-guard` ACL) therefore sees an empty user and rejects every purge/move:

```
admission webhook scope-guard 403 Forbidden: user '' not allowed to purge_project in '<ws>'/'<proj>'
```

With such a webhook enabled, the destructive admin ops are effectively unusable. `handle_write_page` already threads the actor (via `WritePageRequest::actor`); the destructive handlers were the gap.

## Fix

Extract the auth-middleware `Extension<ActorContext>` (same pattern as `handle_write_page`) and set `actor` on the admission contexts:

- `handle_purge_project` → `purge_ctx`
- `handle_move_project` → threaded into `true_move_project` (`move_ctx`) and `copy_purge_merge` (the source `purge_ctx`)
- `rename-project` is unaffected (it runs no admission chain)

`copy_purge_merge` crosses the 7-arg clippy threshold, so it gets `#[allow(clippy::too_many_arguments)]` (matching the existing store-layer helpers).

## Tests

- `purge_project_admission_carries_the_actor` — the admission webhook receives `ctx.actor.user`.
- `scope_guard_blocks_purge_without_actor_allows_with_actor` — a `Reject`-policy webhook authorizing by `ctx.actor.user`: empty actor → 500, `actor=alice` → 200 (the 200, not 404, also proves the denied purge left the project intact).

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and the `ai-memory-mcp` test suite are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Admin destructive ops (`/admin/purge-project`, `/admin/move-project`) now
+  propagate the authenticated actor (from the auth middleware's
+  `Extension<ActorContext>`) into the admission context, so a `scope-guard`
+  admission webhook can authorize them by user. They previously built the
+  context with an empty actor, which a per-user scope-guard ACL rejected with
+  `403 user '' not allowed to purge_project`, making purge/move unusable on any
+  instance running scope-guard. `rename-project` is unaffected (it runs no
+  admission chain).
+
 ## [1.0.6] - 2026-06-14
 
 ### Changed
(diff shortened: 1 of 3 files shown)


akitaonrails:
Thanks @djalmajr — the admission actor propagation fix was well scoped, and the regression tests made the remaining merge-copy gap straightforward to close. Merged as aa1488c7aacdcbbd6067ccf6f412a16582e016f9.

PR #99, opened 2026-06-15
fix(hook): give 429 a free retry + add mid-session spool drain

djalmajr:
Two hook-spool delivery-reliability fixes (one PR — both touch the spool drain path).

## 1. A 429 no longer burns a spooled event's retry budget

`post_hook` returned a bare `bool`, so the drain treated a **429** (`hook queue full` — the router's saturation signal via `try_acquire_owned()`; the event was **never processed**) identically to a genuine 401/5xx/transport failure: it bumped `attempts` and dropped the event after `MAX_ATTEMPTS = 8`. A saturation burst therefore silently discarded real observations.

`post_hook` now returns `PostOutcome { Delivered, Saturated, Failed }`. The drain leaves a `Saturated` (429) entry queued **without** bumping attempts (`MAX_AGE_MS` still bounds it against a permanently-saturated server); only `Failed` (other non-2xx / transport error) counts toward `MAX_ATTEMPTS`.

## 2. Mid-session catch-up drain

Per-event hooks only enqueue, so the spool was drained solely at session boundaries (`session-start` cleanup, `session-end` flush). Under a heavy session, enqueue outpaced the boundary drain and the backlog grew until the next boundary.

`post-tool-use` now runs a **size-triggered, tightly time-boxed (~250 ms)** drain once the spool crosses a threshold (default `32`, override `AI_MEMORY_HOOK_INCREMENTAL_THRESHOLD`), so a hot session keeps the backlog flat without ever stalling a tool call. A light session pays only a `read_dir` (the new `spool_len`). The 250 ms value is both the total budget **and** the per-event timeout, so a single slow POST stays sub-second. A detached background drain was rejected: the hook process exits immediately, tearing down the runtime before a spawned task could finish.

`rename`/other events are untouched; boundaries remain the main flush.

## Automated tests

- `post_hook` outcome mapping: 429 → `Saturated`, 2xx → `Delivered`, unreachable → `Failed`.
- `drain` leaves a 429 entry queued at `attempts == 0` across `> MAX_ATTEMPTS` passes (never dropped).
- `spool_len` counts entries / 0 for a missing dir.
- `should_incremental_drain` predicate (only `post-tool-use`, only at/above threshold).
- `incremental_drain_threshold` env parse + fallback.

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p ai-memory-cli` are green.

## Runtime evidence (real binary + local server)

**#1 (429):** spooled events drained against a 429 server stay queued at `attempts=[0..0]` across repeated passes (nothing dropped); flipping the server to 202 delivers all of them.
```
server=429: spool grows 1→4, attempts stay [0..0] across 3 drains  (Saturated: no bump, no drop)
flip to 202: one drain → spool 0, 5 POSTs delivered
```

**#2 (incremental drain):**
```
Case A: 31 backlog + 1 post-tool-use (→32 ≥ threshold) → spool 0, server received 32 POSTs
Case B: 5  backlog + 1 post-tool-use (→6  < threshold) → spool stays 6, server delta 0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   healthy DB). Optional `?workspace=&project=` scope; reports only, never
   mutates, so it is safe to run on any cadence. Purely semantic mislandings
   (no cwd/session anomaly) are out of scope by design.
+- Mid-session hook-spool drain: on `post-tool-use`, once the local spool backlog
+  crosses a threshold, the hook runs a tightly time-boxed (~250 ms) catch-up
+  drain so a heavy session keeps the backlog flat instead of waiting for the next
+  session boundary. Tunable via `AI_MEMORY_HOOK_INCREMENTAL_THRESHOLD`
+  (default 32 events).
+
+### Fixed
+- Hook spool no longer counts a server `429` (saturation / `hook queue full`)
+  against a spooled event's `MAX_ATTEMPTS` retry budget: transient backpressure
+  keeps the event queued without burning an attempt (`MAX_AGE_MS` still bounds
+  it), so a saturation burst no longer silently discards real observations.
 
 ## [1.1.0] - 2026-06-16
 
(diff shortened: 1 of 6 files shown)


akitaonrails:
Heads up: we changed a lot of code on main around auto-improvement scheduling, pending proposal flows, migrations, and upgrade safety. Please rebase this PR on the latest main once those changes land, then re-run the relevant tests locally and update the PR with the results before we review/merge.

djalmajr:
Rebased onto `main` (v1.0.11) — applied cleanly, no conflicts.

Local results on the rebased branch:
- `cargo fmt --all --check`: clean
- `cargo clippy -p ai-memory-cli --all-targets -- -D warnings`: clean
- `cargo build --workspace`: ok
- `cargo test -p ai-memory-cli`: 288 passed / 0 failed, including the cases this PR adds — `post_hook_saturated_on_429`, `drain_429_keeps_event_queued_without_bumping_attempts`, `spool_len_counts_entries`, and `should_incremental_drain_only_post_tool_use_over_threshold`.

djalmajr:
**Real-world repro of the failure mode the mid-session drain addresses.**

In a long-running agent session, per-event hooks enqueue to the spool but the only drains fire at `session-start` / `session-end`. A session that produces more than `MAX_SPOOL_FILES` (10000) captures before it ends pins the spool at the cap — and since enqueue evicts oldest-first once full, the earliest captures are dropped before they ever reach the `session-end` drain.

Observed on a multi-hour session: spool pinned at 10000, every remaining entry `attempts: 0` / `last_error: null` under valid auth (not a delivery failure — they were simply never drained), while a manual `session-start` drain only clears the small prior-session remainder per invocation. So the loss is purely a function of **drain cadence**, not auth or connectivity.

The incremental mid-session drain keeps the spool below the cap during the session, so nothing is evicted before delivery.

Issue #97, opened 2026-06-14
Cross-project scope bleed: MCP tools without explicit scope can resolve to the wrong project under per-actor auto-scoping


djalmajr:
## Summary

When a single operator runs multiple concurrent agent sessions across different projects against one shared (HTTP-remote) ai-memory server, MCP tool calls that omit explicit `workspace`/`project` can resolve to the wrong project — the last one that generated hook activity for that operator — instead of the calling session's project. Reads return another project's data; writes can land in the wrong bucket.

## Root cause — identity mismatch between two channels

1. The hook ingress publishes the active project keyed by `ActorKey{user, session_id}`, where `session_id` is the agent's hook session id (`router.rs`, via `ActiveProject::set_for`). `set_for` also writes a global single slot and a user-only slot (`active_project.rs`).
2. The MCP request path derives its actor from the `mcp-session-id` / headers (`server.rs`, `actor_key_from_parts`) — a different identifier than the hook's session id.
3. On lookup (`get_for`, `active_project.rs`) the scoped key misses, so resolution falls back to the user-only / global slot. With a single operator that slot is shared across every repo, so "last hook activity wins."

## Why the usual fixes don't apply to an HTTP-remote deployment

- MCP `roots` (the cwd-sharing channel) is delivered only to local stdio servers, not HTTP servers, and is being deprecated. An HTTP-remote server never receives the client cwd on a tool call — not via roots, headers, env, or `clientInfo`.
- Consolidation is not a source of the bleed: it resolves the project from `session_project_ids(session_id)` — the project of the session's own observations — which is correct (`consolidator.rs`).

## Current workaround

Pass `workspace` + `project` together on tool calls where the project matters; that routes through `lookup_ids` and bypasses the contaminable fallback slot. (One facet — `memory_handoff_begin`/`memory_handoff_accept` not accepting `workspace` at all — is fixed in #96.)

## Proposed durable fix (design input wanted)

A local stdio-proxy mode: ship the `ai-memory` binary as a thin local MCP stdio server that reads the client cwd (`roots` / `CLAUDE_PROJECT_DIR`), stamps the resolved project, and forwards to the remote HTTP backend. This restores per-tool-call cwd availability and fixes auto-scoping universally, without depending on the agent passing explicit scope.

Alternatives: per-repo identity tokens (heavier ops); or formally documenting explicit `workspace` + `project` as the supported pattern for multi-project operators.

## Code references

[...]

PR #100, opened 2026-06-15
feat(admin): read-only cross-project contamination audit

djalmajr:
A cheap, recurring, **read-only** way to detect likely cross-project contamination — the kind of mislanding the auto-scope-bleed bug (#97) produced — without an LLM or a multi-agent sweep.

## What

`GET /admin/audit-contamination` (+ `ai-memory audit-contamination`) runs two HIGH-precision, SQL-only heuristics in the read pool:

- **`session_wrong_bucket`** — a session whose `cwd` longest-prefix-resolves to a *different* project than the one it landed in (the auto-scope-bleed signature). Resolved with the **same** `find_project_by_cwd_prefix` the runtime uses, so the audit never flags a session a live resolve would not (exact prefix + cwd-safety parity — no duplicated SQL to drift).
- **`observation_session_drift`** — an observation whose `project_id` disagrees with its owning session's. On a healthy DB this is always empty, so a non-empty result is a regression **tripwire** (e.g. a CI/cron probe alerting on non-zero counts).

The endpoint **only reports** (never mutates), accepts an optional `?workspace=&project=` scope (restricts to one landed bucket), and is safe to run on any cadence.

**Honest scope:** this detects only contamination that leaves a *structural* trace (a cwd/session anomaly). Purely **semantic** mislandings — a `concepts/`/`decisions/` page whose topic belongs elsewhere with no cwd/session signal — are not SQL-detectable and are out of scope by design.

## Tests

- Store integration (`crates/ai-memory-store/tests/audit_contamination.rs`): a seeded fixture with both contamination signatures + clean controls asserts each check flags exactly the contaminated rows; a clean-DB case returns nothing; a scoped run restricts to the landed bucket.
- The admin route-enumeration/authz test covers the new `GET /admin/audit-contamination` route.

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are green.

## Runtime evidence (real server + CLI, end-to-end)

Seeded a DB (2 sessions, 2 observations, 3 projects) with one wrong-bucket session and one drifted observation, served over HTTP, and ran the CLI thin-client:

```
$ ai-memory audit-contamination
Contamination audit: 1 session(s) mis-bucketed, 1 observation(s) drifted.
  [high] observation … in auditws/proj-b but its session (…) is in project proj-a
  [high] session … landed in auditws/proj-a but its cwd (/w/b/sub) resolves to project proj-b
$ ai-memory audit-contamination --json    # summary: {"sessions_misbucketed":1,"observations_drifted":1}
```
The clean session/observation are correctly not flagged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   instance running scope-guard. `rename-project` is unaffected (it runs no
   admission chain).
 
+### Added
+- `GET /admin/audit-contamination` (and `ai-memory audit-contamination`) — a
+  read-only, SQL-only structural cross-project contamination audit. Flags
+  sessions whose `cwd` longest-prefix-resolves to a different project than the
+  one they landed in (the auto-scope-bleed signature, resolved with the same
+  prefix logic the runtime uses) and observations whose project disagrees with
+  their owning session (a regression tripwire that should stay empty on a
+  healthy DB). Optional `?workspace=&project=` scope; reports only, never
+  mutates, so it is safe to run on any cadence. Purely semantic mislandings
+  (no cwd/session anomaly) are out of scope by design.
+
 ## [1.0.6] - 2026-06-14
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `GET /admin/audit-contamination` (and `ai-memory audit-contamination`) — a
+  read-only, SQL-only structural cross-project contamination audit. Flags
+  sessions whose `cwd` longest-prefix-resolves to a different project than the
+  one they landed in (the auto-scope-bleed signature, resolved with the same
+  prefix logic the runtime uses) and observations whose project disagrees with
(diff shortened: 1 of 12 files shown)


akitaonrails:
Heads up: we changed a lot of code on main around auto-improvement scheduling, pending proposal flows, migrations, and upgrade safety. Please rebase this PR on the latest main once those changes land, then re-run the relevant tests locally and update the PR with the results before we review/merge.

djalmajr:
Rebased onto `main` (v1.0.11). One conflict, in `crates/ai-memory-store/src/lib.rs` — the `pub use reader::{…}` re-export list, where main added `AutoImproveCandidateSession` and this branch adds the `Contamination*` types; resolved by unioning the list.

Local results on the rebased branch:
- `cargo fmt --all --check`: clean
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo build --workspace`: ok
- `cargo test -p ai-memory-store --test audit_contamination`: 3/3 pass
- `cargo test --workspace`: green except `ai-memory-mcp`'s `autoscope_stress` (`per_session_isolates_concurrent_*`, `ttl_eviction_under_concurrent_insertion`), which fail non-deterministically. They reproduce on unmodified `main` (258452d) as well — one clean run, one failing run — so it is a pre-existing concurrency/timing flake, not introduced here (this change only adds a read-only contamination-audit reader query + admin route + CLI).

PR #122, opened 2026-06-20
feat(hook): batched spool delivery via POST /hook/batch to amortize per-event RTT

djalmajr:
## Summary

Under heavy **parallel** load (many agent sessions sharing one `data_dir`/hook-spool) draining to a **remote, gated** server, the sequential per-event drain falls behind: each POST pays TLS + network RTT + the edge auth hop, and the 250 ms mid-session drain clears only a handful per pass. The spool grows to `MAX_SPOOL_FILES` and `prune_spool_file_count` evicts the **oldest entries at enqueue time without delivering them** — silent capture loss (the `attempts:0 / last_error:null` symptom).

This PR adds a **batched delivery path** that collapses N round-trips into 1, plus a few supporting fixes. It’s built on top of `v1.1.3`.

> Heads-up: this also turns the **two currently-red CI checks on `v1.1.3` green** — see “CI” below.

## The fix (batch delivery)

- **Server — `POST /hook/batch`** (`ai-memory-hooks`): accepts a JSON array of `{url, body}`, reuses `HookEnvelope::from_query_and_body` + `process()` per event, processed **inline and fail-fast** (stops on the first error, returns `{accepted: K}`). One ingest permit + one auth pass per batch. The per-event query is parsed from each item’s `url` via `serde_urlencoded` (the same crate axum’s `Query` extractor uses).
- **Client — batched drain** (`ai-memory-cli`): the drain groups consecutive entries sharing the same endpoint + bearer, bounded by count (256) and bytes (8 MiB), and deletes only the acked prefix. **Automatic fall-back to per-event `POST /hook` on 404/405**, so a new client stays compatible with an un-upgraded server during rollout. `/hook` is unchanged.
- **Observability**: `prune_spool_file_count` now warns on stderr when it evicts undelivered events at the cap, so sustained loss is visible instead of silent.

`drain()`’s signature is unchanged, so the `session-start`/`session-end`/incremental call sites are untouched.

### Why batch (not concurrency)
The dominant cost is **network round-trip per request**, not the auth itself (a JWKS-caching forwardAuth validates locally in ~ms). Batching amortizes RTT directly (1 request instead of N) and preserves attribution (one forwardAuth pass / one actor per drain). Idempotency (a `UNIQUE client_key` + `ON CONFLICT`) is a sensible follow-up that would close the duplicate-on-retry window a batch widens.

### Production validation
Deployed against a Keycloak-gated remote: a backlog of **2132 events drained in ~15 s (~140/s)**, spool 2574 → 442, **zero eviction**, zero error. The same backlog per-event would take ~8 min at the sequential ceiling (~4/s, TTFB ~228 ms) and typically hit the cap → loss. **≈35× throughput.**

## Supporting fixes (independent commits, easy to split if you prefer)

- **`fix(store)`: prefix-match `repo_path`/`cwd` with `_`/`%` literally.** `find_project_by_cwd_prefix` silently dropped any cwd/`repo_path` containing a LIKE wildcard (`is_safe_cwd_for_prefix_match` rejected the cwd; the SQL rejected the `repo_path`), so an agent in e.g. `/home/u/my_project/src` never matched its parent project. Now the stored `repo_path` is **escaped** in the LIKE pattern (literal match, never a wildcard) and the cwd `_`/`%` rejection is dropped. Adds a regression test.
- **`test`: robustness.** (1) `process_guard::sibling_processes` made the destructive-command tests flake (a dev box always has another `ai-memory` alive) — opt out under `cfg!(test)` / `AI_MEMORY_TEST_NO_PROCESS_GUARD`; the dedicated `#[ignore]`d guard test still covers it. (2) `autoscope_stress` replaced a fixed 15 ms settle with a poll. (3) the repo-root prefix-match test no longer fails on macOS (`/var`→`/private/var` + `_` in the temp path). `cargo test --workspace` now passes repeatedly.
- **`fix(deps)`: bump `git2` 0.20 → 0.21** to clear `RUSTSEC-2026-0184` (unsound `Blame::blame_buffer` advisory). Adjusts the three `Commit::summary()`/`body()` call sites for the new `Result<Option<&str>, _>` return.

## CI
The two failing checks on `v1.1.3` are fixed here:
- **`test (macos-latest)`** — failed on the repo-root prefix-match test (deterministic on macOS due to the `/var` symlink + the `_` in the temp dir hitting the prefix-match `_` rejection). Fixed by the prefix-match + test-robustness commits.
- **`cargo-deny`** — failed on `RUSTSEC-2026-0184` (`git2 0.20.4`). Fixed by the `git2` bump. `cargo deny check` is now fully green.

## Test plan
- New: `/hook/batch` round-trip, per-event fallback on 404/405, `handle_hook_batch` ack + 429 + query parse, literal-`_` prefix-match.
- `cargo test --workspace` green (incl. macOS); `cargo deny check` green; existing spool tests unchanged.
- End-to-end validated in production (numbers above).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   default, receives proposal JSON on stdin, fails closed on command/timeout/JSON
   errors or insufficient score delta, and records eval failures as rejected
   candidates without running from hook paths.
+- Hook spool drains now use `POST /hook/batch` when the server supports it,
+  grouping compatible queued lifecycle events into bounded batches to amortize
+  remote request latency. Older servers fall back to the existing per-event
+  `POST /hook` path.
 
 ### Fixed
 - Auto-improvement eval gates now apply timeouts to the full child interaction,
@@ -35,7 +39,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   aliases still prefix-match later events into the same project.
 - Updated `git2` to the patched 0.21 line to clear new RustSec unsoundness
   advisories in libgit2 bindings.
-<<<<<<< HEAD
 - Native Claude Code hooks now capture array-shaped `tool_response` content and
   recognize the native `user-prompt-submit` event token, restoring prompt text
   and tool output bodies for native installs.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,6 +140,7 @@ dependencies = [
  "regex",
  "serde",
  "serde_json",
+ "serde_urlencoded",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
--- a/crates/ai-memory-cli/src/commands/hook_capture.rs
+++ b/crates/ai-memory-cli/src/commands/hook_capture.rs
@@ -177,6 +177,75 @@ pub async fn post_hook(
     }
 }
 
+/// Outcome of one `POST /hook/batch` request — many spooled events delivered in
+/// a single round-trip, so a draining client amortizes TLS + network RTT + the
+/// edge auth hop over the whole batch instead of paying it per event.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BatchOutcome {
+    /// Server committed the leading `usize` items (contiguous prefix, oldest
+    /// first). Equals the request length on full success; a smaller value means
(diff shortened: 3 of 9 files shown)


akitaonrails:
Merged after CI passed. I also added follow-up hardening before merge: bounded batch drain behavior, budget-before-POST handling, and regression coverage for zero-budget drain, partial ack, endpoint split, and bearer split cases.

PR #123, opened 2026-06-21
feat(web): folder-scoped + all-projects chat with create/edit/delete

cristianodewes:
## What changed

Adds an LLM chat to the built-in `/web` browser, grounded in the wiki.

- **Folder / project chat** (`POST /w/{ws}/{proj}/chat`): open a chat on any folder (and its sub-folders) or a whole project — reachable from the project list, the project tree (per folder + a project-wide button), and any page view. Context is hybrid: an outline of every in-scope page plus the bodies of the pages most relevant to the question (embedding cosine over stored vectors, BM25 fallback).
- **All-projects chat** (`POST /chat`): a read-only chat from the home page grounded in every project, with cross-project citations.
- **Create / edit / delete**: in a project chat the model can modify memory via structured actions. `create`/`edit` apply immediately through `Wiki::write_page` (sanitization, attribution, git history, and the SQLite index stay consistent); `delete` is deferred and applied only after an explicit confirmation click (`POST /w/{ws}/{proj}/chat/delete`). The all-projects chat is read-only.
- **Rendering**: assistant replies render as Markdown (fenced code blocks, lists, inline code, emphasis) via a small dependency-free, HTML-escaping client.

Chat is available only when an LLM provider is configured; otherwise every UI affordance is hidden and the endpoints return `503`. The server's existing `Arc<dyn LlmProvider>` + embedder are threaded into `WebState` via a new `router()` signature; the read-only JSON `/api/v1` surface is unchanged and does not expose chat.

Also fixes an **`openai-oauth` (Responses API) multi-turn bug**: assistant turns were serialized as `input_text`, which the API rejects with `400 invalid_value` on `input[n].content[0]`, so every *second* chat turn failed. They are now `output_text`. Single-shot callers (consolidation, auto-improve, PreCompact) never send an assistant turn, so the bug only surfaced with multi-turn chat.

## Why

The `/web` browser was read-only and required `docker exec`/MCP to reason over or curate memory. This lets you discuss a folder's memories — or the whole wiki — and curate them (create/edit/delete) straight from the browser, with destructive deletes gated behind explicit confirmation. The `openai-oauth` fix is a prerequisite for any multi-turn chat on that provider.

## Test plan

- [x] `cargo test --workspace` passes (one pre-existing **Windows-only** failure unrelated to this PR — see Notes)
- [x] Manual: ran the server in Docker against a real wiki with `openai-oauth` / `gpt-5.5`; verified folder chat, all-projects chat (cross-project citations), multi-turn, Markdown rendering, autonomous create, and confirm-then-delete end to end.

New regression tests: in `crates/ai-memory-web/tests/routes.rs` — folder scoping, `503`/`404` gating, create-applies, delete-deferred-then-confirmed, and global chat; and `codex_request_types_assistant_turns_as_output_text` in `ai-memory-llm`.

## Notes for reviewers

- **Write capability & safety**: mutations go only through `Wiki::write_page` / `Wiki::delete_page` (no raw file writes); scope is resolved with `lookup_existing_scope` (no-create, fail-closed); writes are confined to the chat's project; the all-projects chat is read-only; and `delete` always requires a human confirmation click. On a deployed server `/web` sits behind the bearer token, so only token holders can write.
- **First write/POST surface on `/web`**: the crate was deliberately read-only; the chat is the one interactive surface and never mutates outside the wiki write path.
- **Provider mechanism**: tool use is done via structured output (`complete_structured`) because the `LlmProvider` trait has no native function-calling; the schema is a `{ reply, actions[] }` turn. Replies are non-streaming (the trait has no streaming API).
- **Default-on**: chat (incl. write) is enabled whenever an LLM provider is configured. Happy to gate it behind a dedicated flag (e.g. `--enable-web-chat`) if you'd prefer an explicit opt-in for the write surface.
- **Pre-existing CI reds (not introduced here)**: `cargo-deny` fails on the `git2 0.20.4` RustSec advisories (RUSTSEC-2026-0183 / -0184) and `test (macos-latest)` fails on `router::tests::host_resolved_repo_root_override_records_repo_path_when_visible` — both are already red on `main` and both are fixed by #122. The single local `cargo test` failure is the Windows-only `render_shared::tests::grok_script_payload_uses_grok_bundle` (it hard-codes POSIX `/` separators; passes on the Linux/macOS CI).


akitaonrails:
Holding this one for now. After rebase the local full gate passes, but the branch is not mergeable yet because it crosses from read-only web browsing into browser/LLM-mediated writes without the required guardrails:\n\n- browser chat POSTs do not work with the current auth model (Basic/cookie are GET-only; non-GET requires Bearer), and accepting cookies for POST would need CSRF protection;\n- write/delete handlers do not extract/authorize  or propagate /author/admission context, so multi-user attribution and admission audit are lost;\n- create/edit writes are applied immediately from LLM output without confirmation/review, which is prompt-injection/data-integrity risk;\n- folder scope limits context only, not mutation targets;\n- edits overwrite metadata (, , , frontmatter) and create/edit semantics are not checked;\n- UI/docs still describe  as read-only in several places.\n\nI do not want to patch this into a rushed security model. Recommended path: split this into a read-only chat PR first, or add a proper pending-confirmation mutation flow with CSRF, actor attribution, scope enforcement, existence checks, and metadata preservation before merge.

akitaonrails:
Correction to my previous comment (shell stripped several code-formatted names): holding this one for now. After rebase the local full gate passes, but the branch is not mergeable yet because it crosses from read-only web browsing into browser/LLM-mediated writes without the required guardrails:

- browser chat POSTs do not work with the current auth model: Basic/cookie are GET-only; non-GET requires Bearer. Accepting cookies for POST would need CSRF protection.
- write/delete handlers do not extract/authorize AuthLevel or propagate ActorContext/author/admission context, so multi-user attribution and admission audit are lost.
- create/edit writes are applied immediately from LLM output without confirmation/review, which is prompt-injection/data-integrity risk.
- folder scope limits context only, not mutation targets.
- edits overwrite metadata such as kind, tier, pinned, and frontmatter; create/edit semantics are not checked.
- UI/docs still describe the web browser as read-only in several places.

Recommended path: split this into a read-only chat PR first, or add a proper pending-confirmation mutation flow with CSRF, actor attribution, scope enforcement, existence checks, and metadata preservation before merge.

akitaonrails:
Thanks for the work here. I am going to close this because the proposed browser chat mutation surface is too far out of scope for ai-memory as maintained here: it turns the web UI into an interactive create/edit/delete assistant and brings a larger product surface, safety model, auth/CSRF design, confirmation flow, and mutation workflow that should not be folded into core ai-memory in this shape. This would be better served as a separate companion project or experimental web editor that can evolve independently against ai-memory, rather than becoming part of the core project.

akitaonrails:
Closing as out of scope for the core project. A separate companion/web-editor project would be a better home for this direction.

akitaonrails:
Revising the direction here: the browser chat/editor idea is useful, but it should not patch write-capable chat routes and LLM-mediated create/edit/delete behavior into core ai-memory. The better path is an optional companion web-editor crate or separate project.

Planned companion shape:

- standalone `ai-memory-web-editor`, not a default core workspace member;
- use `/api/v1` for read-only workspace/project/page/search/graph/overview context;
- run folder/project/all-project chat in the companion;
- keep all-project chat read-only by default;
- model-proposed writes become pending actions with explicit diff/confirmation;
- the companion backend, not the browser, calls existing ai-memory write/delete surfaces with a server-side token;
- CSRF/session policy, rate limits, confirmation state, mutation scoping, and prompt-injection defenses live in the companion;
- no direct SQLite access, no direct wiki-file writes, and no bypass of auth, admission webhooks, actor attribution, scope resolution, or indexing.

So the change of mind is: this is not something to merge into core, but it is a reasonable companion web-editor direction. I documented the boundary and implementation plan in `docs/companion-crates.md` and linked it from the README.

Issue #116, opened 2026-06-19
Native Windows Claude Code hooks silently drop all capture: install-hooks emits a \\?\ verbatim --data-dir that breaks the event spool


cristianodewes:
**Version**

```
ai-memory 1.1.0  (code path still present on main @ v1.1.2)
```

**Summary**

On native Windows, every capture hook (`UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `PreCompact`, `Stop`, `SessionEnd`) silently records nothing. The server stays frozen at its initial counts even across many active sessions, while each hook exits `0` and prints `{}` as if it succeeded. Only `SessionStart` works (it injects a handoff).

Root cause is an inconsistency between two halves of the **same binary**:

1. `install-hooks` bakes a **verbatim / extended-length** `--data-dir "\\?\C:\..."` into the generated `settings.json`.
2. The `hook` subcommand's event **spool write fails** when `--data-dir` carries the `\\?\` prefix — and the failure is swallowed (`let _ = hook_spool::enqueue(...)`), so nothing is ever spooled and therefore nothing is ever drained.

This is not a user misconfiguration: it reproduces straight from the documented `install-hooks --apply` flow.

**What happened**

- `memory_status` / `memory_briefing` stayed at the same `sessions` / `observations` counts for days across multiple real sessions.
- `hook-spool/` was always empty.
- Each native hook returned `{}` / exit 0 (fail-open), so nothing surfaced the loss.

**Root cause detail**

`ai-memory install-hooks --agent claude-code` (print mode, default data-dir) renders, for all 7 events:

```
"...\ai-memory.exe" --data-dir "\\?\C:\Users\me\AppData\Local\ai-memory" hook --event <event> ...
```

The `\\?\` prefix appears even when a clean `--data-dir C:\Users\me\AppData\Local\ai-memory` is passed explicitly — the binary canonicalizes the data dir to the Windows verbatim form (also visible in the startup log: `data_dir=\\?\C:\Users\me\AppData\Local\ai-memory`). `render_shared.rs::native_data_dir_arg` then renders it verbatim via `to_string_lossy()`.

At hook time, `commands/hook.rs::run()` does:

```rust
let dd = resolve_data_dir(data_dir.as_deref());   // = \\?\C:\...
let spool = hook_spool::spool_dir(&dd);
...
let _ = hook_spool::enqueue(&spool, &entry);       // error discarded
```

[...]

PR #124, opened 2026-06-21
fix(hooks): warn instead of silently swallowing spool retry-count persist failures

matheus-rodrigues00:
## What

The headless hook spool's drain loop bumps an event's retry-attempt counter on a
failed delivery and persists it back to disk via `rewrite_entry` (an atomic
temp-write + rename). That persist was swallowed:

```rust
// crates/ai-memory-cli/src/commands/hook_spool.rs — PostOutcome::Failed
let _ = rewrite_entry(&path, &entry);   // fallible write, error discarded
result.remaining += 1;
```

If the rewrite fails (disk full, a read-only dir, a permission change, or a
Windows `\\?\` verbatim-path write that never lands), the bumped `attempts` lives
only in memory and is lost. The on-disk entry keeps its old count, so it never
reaches `MAX_ATTEMPTS` (8) and retries on **every** drain boundary — with no
operator signal, and with `DrainResult` still reporting `remaining` as if nothing
were wrong — until the 7-day `MAX_AGE_MS` prune eventually drops it. The failure
is silent twice over.

## Why this fits the project's direction

This is the same bug class already fixed in **v1.1.3 (issue #116)**: the sibling
`let _ = enqueue(...)` swallow was changed to emit a sanitized stderr warning
(see that release's `### Fixed` block). This applies the identical remedy to the
one remaining `rewrite_entry` swallow. There is no separate tracking issue for
this one — the v1.1.3 precedent is the rationale.

## The change

- Route the persist through a small `note_retry_persist` seam that emits a
  sanitized stderr warning instead of swallowing the error, and reports whether
  the count was persisted.
- The warning carries **no path** — a raw spool path can be a Windows verbatim
  `\\?\…` path, exactly the leak class #116 was about. It mirrors the wording and
  style of the existing enqueue-failure warning.
- Strictly **fire-and-forget**: it only warns — never panics, returns, or blocks
  — so the hook binary still exits 0 and the drain loop keeps draining.
- CHANGELOG `### Fixed` entry under `[Unreleased]`.

## Tests

- `note_retry_persist_reports_failure` / `note_retry_persist_reports_success` —
  unit-test the warn-vs-not decision by feeding a synthetic `io::Result`. This is
  **root-proof**: CI runs as root, which ignores `chmod`, so a read-only-dir test
  would never actually fault.
- `drain_stays_robust_when_retry_count_cannot_persist` — end-to-end: forces a
  uid-independent rewrite failure (occupying the entry's `<name>.json.tmp` path
  with a directory so the atomic temp-write cannot be created) and asserts the
  drain stays robust — no panic, the event is still counted as `remaining`,
  nothing is dropped, and the entry survives for the next boundary.

---

Pushed as a few small, logically separate commits (refactor → fix + CHANGELOG →
tests); happy to squash on merge. The four CI gates apply — `cargo fmt --check`,
`cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`,
`cargo deny check` — and were run locally in a `rust:1.95` container.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,9 +35,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   aliases still prefix-match later events into the same project.
 - Updated `git2` to the patched 0.21 line to clear new RustSec unsoundness
   advisories in libgit2 bindings.
+<<<<<<< HEAD
 - Native Claude Code hooks now capture array-shaped `tool_response` content and
   recognize the native `user-prompt-submit` event token, restoring prompt text
   and tool output bodies for native installs.
+- The headless hook spool now warns (instead of silently swallowing) when it
+  cannot persist a bumped retry-attempt count back to disk, matching the v1.1.3
+  enqueue-failure fix. A failed atomic rewrite previously lost the in-memory
+  attempt bump, so a poison spool entry never reached `MAX_ATTEMPTS` and kept
+  retrying on every drain boundary with no operator signal until the 7-day
+  age-out pruned it. The warning is sanitized and carries no path (a raw spool
+  path can be a Windows verbatim `\\?\…` path).
 
 ## [1.1.3] - 2026-06-20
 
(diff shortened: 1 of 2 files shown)


akitaonrails:
Thanks @matheus-rodrigues00 — I rebased this on top of current main, kept the changelog entry alongside the native hook capture note, reran the full local gate, and CI is green. Merged.

PR #127, opened 2026-06-23
Fix consolidation recency for later corrections

### LINKED ISSUE #126: Long-session consolidation keeps a fact's stale early version and drops its later in-session correction
**Version**
Output of `ai-memory --version`: ai-memory 1.1.3
(Docker; local image = v1.1.3 + PR #121 hook-capture fix. The consolidation code
referenced below is unmodified upstream v1.1.3.)

**What happened**
In a long session (~2h, 349 observations, well over the projection cap), a forced
`memory_consolidate` produced a session page asserting two facts in their early,
later-corrected form. Each was stated early and then explicitly corrected later in
the SAME session, but the page kept the stale early version:

1. Deploy: page said "manual deploy" and suggested "consider Kamal". The session
   had later established Coolify + Traefik (Docker).
2. Payment: page said "likely Stripe (not confirmed)". The session had later
   confirmed Stripe.

This is not random hallucination. There is no recency handling anywhere in the
consolidation path, so a fact's first version tends to win over its correction.

**What I expected**
The consolidated page should reflect the session's final state. When a fact is
corrected later in the same session, the corrected value should win
(last-write-wins), and the correction should not be dropped during observation
selection on long sessions.

**Steps to reproduce**
1. Run a session long enough to exceed MAX_PROJECTED_OBSERVATIONS = 256, in which a
   fact is stated early and corrected later (e.g. "deploy is manual" early,
   "deploy is Coolify+Traefik" later).
2. Run `memory_consolidate` (dry_run=false) for that session.
3. Read sessions/<id>.md: it reflects the early, stale version of the fact.

Contrast: a shorter session (under the 256 cap, nothing dropped) consolidates the
same kind of correction correctly, so the failure tracks exceeding the cap.

**Environment**
- OS: macOS (Apple Silicon); server runs on a Synology NAS (x86_64)
- Docker? yes; local image built from v1.1.3 + PR #121
- Agent: Claude Code
- Transport: http

**Relevant logs**
None. This is a silent correctness issue, not a crash. Code analysis below.

---

**Root cause (code, v1.1.3)**
No notion of recency exists in either selection or synthesis:

1. Synthesis. `single_consolidate_system.md` enforces faithfulness but never says
   to prefer the most recent statement when observations disagree. Observations
   are fed chronological ascending (`ai-memory-store` reader,
   `ORDER BY created_at ASC`), so the stale version is read first and wins by
   primacy.
2. Selection. When a session exceeds MAX_PROJECTED_OBSERVATIONS = 256,
   `project_observations` drops observations by lowest `observation_score`
   (`projection.rs`). That score has no recency term (`created_at` is rendered but
   never scored). It is driven by a hardcoded kind weight (UserPrompt +100 ...
   PostToolUse +30 ... PreToolUse +5), a hardcoded keyword list (+45 for "fix",
   "error", "bug", "decision"...), and first/last anchors. `importance` enters
   only as a 1 to 10 base, assigned purely by event kind (`importance_for` in
   `router.rs` is a const fn of the event, so every PostToolUse is 5). A
   plainly-worded correction with no signal keyword scores like routine output and
   can be pruned before the model sees it. In the failing 349-observation session,
   ~93 observations were dropped in selection.

Note: the lint pass only compares across pages, and the auto-improve reviewer is
off by default, so neither reconciles an intra-session correction inside a single
freshly-written session page.

**Suggested fixes (small, complementary, separable)**
1. Synthesis: one line in `single_consolidate_system.md` (and the batch prompt):
   when later observations contradict earlier ones about the same fact, treat the
   most recent as authoritative; describe the final state, not superseded drafts.
2. Selection: add a recency term to `observation_score` so later observations are
   not at a disadvantage when the cap forces drops.
3. Optional: let `importance` reflect content rather than event kind alone (or
   widen the keyword set), so a correction is not dropped just for lacking a
   keyword.

Happy to test a prompt patch.

**Related**
#69 and #102 address large-session consolidation from the request-size /
schema-adherence angle (windowing observations to fit the budget). This is the
correctness complement: given the observations fit, which ones survive selection
and which version of a changed fact wins. Neither touches recency or
last-write-wins.

snvtac:
## What changed

- Bias single-page and batch consolidation projection toward recent observations so late corrections survive long-session count caps and character-budget pruning.
- Teach both consolidation prompts to treat later contradictory observations as the authoritative final state.
- Add regression coverage for long sessions, tight budgets, penultimate corrections before `SessionEnd`, and both single/batch request wiring.
- Stabilize the auto-improve eval-gate test so non-timeout error cases are not misclassified as timeouts under concurrent test load.

## Why

Fixes #126. Long sessions could keep an early stale version of a fact while dropping the later correction during observation projection, causing consolidation to write outdated state.

## Test plan

- [x] `cargo test --workspace --all-targets` passes
- [x] Manual test: focused regression tests for recent/correction projection and single/batch request wiring pass

## Notes for reviewers

Recent-biased projection is intentionally crate-internal and only used by consolidation. The public `ObservationProjectionConfig` shape is unchanged, and auto-improve continues to use the default projection path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Long-session consolidation now reserves and protects recent observations
+  under count and character pressure, gives later observations a recency score
+  boost, and instructs single-page and batch prompts to treat later corrections
+  as the authoritative final state. This helps keep stale early facts from
+  winning after observation projection drops lower-signal entries. (issue #126)
+
 ## [1.2.0] - 2026-06-23
 
 ### Added
--- a/crates/ai-memory-consolidate/prompts/batch_consolidate_system.md
+++ b/crates/ai-memory-consolidate/prompts/batch_consolidate_system.md
@@ -36,6 +36,9 @@ Do:
   pages with the right `kind` classification.
 - Preserve the user's actual phrasing for decisions and rules —
   these are load-bearing.
+- When later observations contradict earlier observations about the
+  same fact, treat the later observation as authoritative. Record the
+  session's final state, not superseded drafts or stale early guesses.
 - Write the page at whatever length the observations *actually*
   warrant. Don't pad with generic tutorial filler, but don't
   truncate substance either. Dense fact beats artificial
--- a/crates/ai-memory-consolidate/prompts/single_consolidate_system.md
+++ b/crates/ai-memory-consolidate/prompts/single_consolidate_system.md
@@ -22,6 +22,10 @@ the current page body provided. Do NOT:
 If the session yields nothing durable beyond the narrative,
 return a short session log with no fabricated structure.
 
+When later observations contradict earlier observations about the
+same fact, treat the later observation as authoritative. Record the
+session's final state, not superseded drafts or stale early guesses.
+
 ## Style rules
 
 1. Title: short, descriptive (≤ 80 chars). No filler.
--- a/crates/ai-memory-consolidate/src/auto_improve.rs
+++ b/crates/ai-memory-consolidate/src/auto_improve.rs
@@ -1993,19 +1993,22 @@ mod tests {
             (
                 write_eval_script("#!/bin/sh\ncat >/dev/null\nexit 7\n"),
                 "eval_gate_error",
+                5,
             ),
             (
                 write_eval_script("#!/bin/sh\ncat >/dev/null\nprintf 'not-json'\n"),
                 "eval_gate_error",
+                5,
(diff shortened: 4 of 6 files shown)


akitaonrails:
This is now superseded by the fix that landed directly on `main` in b1af7d2 and shipped in `v1.2.2` via 0243727.

The shipped fix covers the same issue:
- late-session corrections get a bounded recency boost during projection;
- single-page and batch consolidation prompts now treat later same-session contradictions as authoritative;
- regression tests cover late correction preservation and prompt guidance.

Release: https://github.com/akitaonrails/ai-memory/releases/tag/v1.2.2

Closing this PR to avoid merging a now-conflicting duplicate branch.

PR #143, opened 2026-07-01
fix(opencode): add session.deleted handler and dispose hook to close sessions

LuizFernando991:
## What changed

Added two missing session lifecycle handlers to the generated OpenCode plugin template (`build_opencode_plugin`):

1. **`session.deleted` event handler** — maps OpenCode's `session.deleted` bus event to `postHook("session-end", ...)`, closing the session in ai-memory when it ends.

2. **`dispose` hook** — sends `session-end` for all tracked sessions when the plugin is unloaded, ensuring sessions are properly closed even when openCode shuts down without firing `session.deleted`.

## Why

OpenCode sessions were being started in ai-memory but never finished. The plugin listened for `session.created`, `session.idle`, and `session.compacted`, but not `session.deleted` — which is the event OpenCode's SDK fires when a session is deleted. Additionally, no cleanup happened when the plugin was disposed (openCode process exiting).

Every other agent already closes sessions properly:
- **Claude Code** has explicit `SessionEnd` + `Stop` hooks in `settings.json`
- **OMP** maps `session_shutdown` to `postHook("session-end")` (line 1981)

## Test plan

- [ ] `cargo test --workspace` passes (new assertions in `opencode_plugin_uses_real_plugin_hooks`)
- [ ] Manual test: generated plugin is valid TypeScript, exported hooks match `@opencode-ai/plugin` 1.17.8 interface

## Notes for reviewers

The `HOOK_IMMEDIATE_EVENTS` set already included `"session-end"`, so the queued flush infrastructure was ready to handle this event — only the triggers were missing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- OpenCode generated plugins now close sessions from the official
+  `session.deleted` event and a deduped best-effort `dispose` fallback, so
+  OpenCode sessions can produce automatic session summaries and handoffs without
+  duplicate `session-end` emissions.
+
 ## [1.6.0] - 2026-07-01
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


akitaonrails:
Merged as b937b781a6fc9bd2aeb97a39a66dc9870f780759. I added the deduped OpenCode close helper, docs/changelog updates, and focused regression coverage before merge; local full gate passed.

PR #146, opened 2026-07-02
Windows follow-up: harden Git Bash paths and fix flaky MCP test

lhzapata:
Follow-up to d4675ee. Two regressions surfaced after that PR landed when validating on a real Windows 11 + Git Bash machine.

## Problems

**`fix(windows): harden git bash native paths` (82b1b84)**

Two issues missed by the initial pass:

- The drain-lock guard in `hook_spool` only checked `WouldBlock` to detect a busy lock. On Windows, `fs2` byte-range locks can also surface as native `ERROR_LOCK_VIOLATION` (code 33), so the single-flight guard was raising an error instead of skipping cleanly when a background drain was already running.
- `libgit2` can fail to reopen a freshly initialised wiki repo on native Windows when the path contains a dot-prefixed segment (common under `tempfile`). The `should_try_commit_cli_fallback` check only matched `ErrorCode::NotFound`; `ErrorClass::Os` path-resolution failures were left unhandled, breaking wiki checkpoints during delete and purge operations under those paths.

A test assertion in `ai-memory-store` also needed to account for path normalisation already applied when reading back a stored repo path key.

**`test(mcp): wait for raw hook routing to settle` (d1ad7cd)**

`fire_raw_hook_and_settle` in the autoscope stress test used a fixed 15 ms sleep after 20 yield points. On slower or loaded machines this was not enough for the hook to be reflected in the router state, causing intermittent failures. The fixed sleep was replaced with an active polling loop (up to 200 × 5 ms) that probes the router and returns as soon as the expected content is visible.

## Changes

- Added `is_drain_lock_busy_error()` in `hook_spool` to also match `ERROR_LOCK_VIOLATION` on Windows alongside `WouldBlock`.
- Extended `should_try_commit_cli_fallback` in `ai-memory-wiki/git.rs` to trigger on `ErrorClass::Os` / "failed to resolve path" errors on Windows.
- Corrected test assertions in `ai-memory-store/ops.rs` to normalise the expected path key before comparison.
- Simplified irrefutable `let` destructures in `hook_drain_process` (clippy suggestion from the Windows build).
- Replaced the fixed sleep in `autoscope_stress::fire_raw_hook_and_settle` with a condition-based poll loop.

No new dependencies. Unix paths unchanged.

## Tests

Validated on Windows 11 + Git Bash and an Arch Linux container:

- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] `cargo deny check`
- [x] `TAILWIND_SKIP=1 cargo test --workspace --locked --all-targets`
- [x] `TAILWIND_SKIP=1 cargo clippy --workspace --locked --all-targets -- -D warnings`
- [x] `TAILWIND_SKIP=1 cargo build --locked --release -p ai-memory-cli`
- [x] `./target/release/ai-memory[.exe] --version` → `ai-memory 1.7.1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   1 MiB (previous contents move to `hook-drain.log.old`), so an agent
   pointed at a chronically unreachable server can no longer grow the log
   without bound.
+- Windows hook-spool drain locks now treat native lock-violation responses as
+  expected lock contention, so overlapping background drains skip cleanly
+  instead of failing the single-flight guard.
+- Windows wiki checkpoints now fall back to the Git CLI for native libgit2
+  path-resolution failures when reopening freshly initialised repos, keeping
+  delete and purge operations usable under dot-prefixed temp or wrapper paths.
 
 ### Changed
 - Post-audit cleanup, no behavior change: the `AI_MEMORY_HOOK_PLATFORM`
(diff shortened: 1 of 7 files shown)


akitaonrails:
Thanks @lhzapata — the tight gating on the libgit2 fallback (ErrorClass::Os + message match, so unknown errors still propagate) and the real-machine validation write-up made this an easy audit. Merged as 05d047e.

PR #149, opened 2026-07-04
fix(llm): normalize Gemini schema `type` arrays to `nullable`

### LINKED ISSUE #148: Gemini structured output 400: `Option<T>` `type:["…","null"]` not normalized to `nullable`
## Summary

Every Gemini structured-output call (consolidation, `auto-improve`) fails with `400 INVALID_ARGUMENT` as soon as the response schema contains an optional field. `ai-memory status` reports:

```
llm: gemini/gemini-2.5-pro error (status 400; {
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"type\" at 'generation_config.response_schema.properties[...]': Proto field is not repeating, cannot start list. ..."
  }
})
```

Consolidation is effectively dead: raw observations and embeddings still work, but no wiki pages regenerate.

## Root cause

`schemars` encodes `Option<T>` as a Draft-2020-12 `type` **array** — `{"type": ["string", "null"]}`. Gemini's `responseSchema` is an OpenAPI-3 subset that only accepts a single `type` string; a nullable field must be expressed as `{"type": "string", "nullable": true}`.

`prepare_schema_for_gemini` (`crates/ai-memory-llm/src/gemini.rs`) already normalizes `$ref` / `$defs` / `oneOf` / `additionalProperties` / … but it never rewrites `type` **arrays**, so the array reaches the API verbatim and Gemini rejects it (`"type" … Proto field is not repeating, cannot start list`).

## Minimal repro (raw API — no ai-memory needed)

```bash
KEY=...   # any valid Gemini key

# 200 OK — the nullable form Gemini expects:
curl -s -w '\n%{http_code}\n' \
  "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-pro:generateContent?key=$KEY" \
  -H 'content-type: application/json' \
  -d '{"contents":[{"parts":[{"text":"x"}]}],"generationConfig":{"response_mime_type":"application/json","response_schema":{"type":"object","properties":{"title":{"type":"string","nullable":true}}}}}'

# 400 INVALID_ARGUMENT — the type-array form schemars emits for Option<T>:
curl -s -w '\n%{http_code}\n' \
  "https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-pro:generateContent?key=$KEY" \
  -H 'content-type: application/json' \
  -d '{"contents":[{"parts":[{"text":"x"}]}],"generationConfig":{"response_mime_type":"application/json","response_schema":{"type":"object","properties":{"title":{"type":["string","null"]}}}}}'
```

## Impact

Any consolidation schema carrying an `Option<_>` field 400s on **every** Gemini model — the endpoint validates the schema at the request layer, so switching models (flash / 2.0) does not help. It is neither an auth nor a quota problem (a plain `generateContent` with the same key returns 200).

## Fix

Extend `prepare_schema_for_gemini` to collapse `type` arrays: `["<t>", "null"]` → `type: "<t>"` + `nullable: true` (and an `anyOf` fallback for genuine multi-type unions). PR incoming.

## Environment

- ai-memory `1.7.1` (Docker, `linux/arm64`)
- `AI_MEMORY_LLM_PROVIDER=gemini`, `AI_MEMORY_LLM_MODEL=gemini-2.5-pro`
- Embeddings (`gemini-embedding-001`) unaffected.

taiar:
## What

`schemars` encodes `Option<T>` as a Draft-2020-12 `type` **array** (`["string", "null"]`). Gemini's `responseSchema` only accepts a single `type` string, so any consolidation schema with an optional field fails with:

> `400 INVALID_ARGUMENT … Unknown name "type" … Proto field is not repeating, cannot start list`

`prepare_schema_for_gemini` already inlines `$ref`s and strips `oneOf` / `additionalProperties` / … but left `type` arrays untouched. This adds a `normalize_nullable_types` pass that collapses them into Gemini's OpenAPI-3 shape:

- `["<t>", "null"]` → `type: "<t>"`, `nullable: true`
- `["null"]` (or empty) → drop `type`, keep `nullable: true`
- genuine multi-type union → `anyOf` of single-`type` schemas (+ `nullable` when `"null"` was present)

## Why

Consolidation / `auto-improve` are dead on Gemini as soon as a schema carries an `Option<_>` field — and it's request-layer validation, so switching models (flash / 2.0) doesn't help. It's neither auth nor quota: a plain `generateContent` with the same key returns 200.

## Testing

- `cargo test -p ai-memory-llm gemini` → 9 passed, incl. 3 new tests (nullable collapse, nested-in-array-items, multi-type union → `anyOf`).
- `cargo fmt -- --check` and `cargo clippy -p ai-memory-llm` clean.
- Verified against the live API: `{"type":"string","nullable":true}` → **200**; `{"type":["string","null"]}` → **400** (repro in #148).

Fixes #148

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   provider blocks ([#147]).
 
 ### Fixed
+- Gemini structured output no longer fails with `400 INVALID_ARGUMENT …
+  "type" … Proto field is not repeating, cannot start list` when a schema
+  contains an optional field. `prepare_schema_for_gemini` now collapses
+  schemars' Draft-2020-12 `type` arrays (e.g. `["string", "null"]` for
+  `Option<T>`) into Gemini's single `type` + `nullable: true` form, so
+  consolidation / auto-improve work again with `gemini-2.5-pro` and other
+  Gemini models.
 - The detached drainer's `logs/hook-drain.log` now rotates once it exceeds
   1 MiB (previous contents move to `hook-drain.log.old`), so an agent
   pointed at a chronically unreachable server can no longer grow the log
--- a/crates/ai-memory-llm/src/gemini.rs
+++ b/crates/ai-memory-llm/src/gemini.rs
@@ -577,6 +577,15 @@ mod tests {
     }
 
     #[test]
+    fn prepare_schema_null_only_type_array_drops_type_keeps_nullable() {
+        let schema = json!({ "type": ["null"] });
+        let prepared = prepare_schema_for_gemini(schema).unwrap();
+        assert!(prepared.get("type").is_none());
+        assert!(prepared.get("anyOf").is_none());
+        assert_eq!(prepared.get("nullable").unwrap(), &json!(true));
+    }
+
+    #[test]
     fn build_request_disables_default_thinking_for_25_flash() {
         let provider =
             GeminiProvider::new(SecretString::from("test-key"), "gemini-2.5-flash").unwrap();
--- a/crates/ai-memory-llm/src/gemini.rs
+++ b/crates/ai-memory-llm/src/gemini.rs
@@ -278,6 +278,11 @@ fn first_text(response: &GeminiResponse) -> Option<String> {
 /// Gemini won't accept. `anyOf` is preserved (the only composition
 /// keyword Gemini accepts).
 ///
+/// Finally we normalise `type` *arrays* (schemars emits `["string",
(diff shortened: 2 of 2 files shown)


akitaonrails:
Thanks @taiar — the raw-API curl repro in #148 (200 vs 400 on the exact payload shapes) made this trivially verifiable, and putting the pass inside prepare_schema_for_gemini means no call path can miss it. Merged as ca43823.

Pushed one small test onto your branch pre-merge (c63bfd8) pinning the ["null"]-only collapse arm.

PR #150, opened 2026-07-04
fix: rootless Docker UID remap breaks install-mcp/install-hooks/install-instructions/install-skills/setup-agent

### LINKED ISSUE #151: install-mcp/install-hooks/install-instructions/install-skills fail under rootless Docker (UID remap breaks host bind-mount writes)
**Version**
`ai-memory 1.8.0`

**What happened**
Every `install-*` subcommand that writes to a host bind-mounted path fails under rootless Docker, with either `Permission denied (os error 13)` or a misleading `directory ... does not exist` error.

**What I expected**
These commands write to `~/.claude/settings.json`, `~/.claude.json`, the hook-staging dir, and `$PWD/CLAUDE.md`/skill dirs, same as they do under rootful Docker.

**Steps to reproduce**
1. Install Docker in rootless mode (`dockerd-rootless-setuptool.sh install`).
2. `ai-memory install-hooks --agent claude-code --apply` → `Error: creating staging dir ~/.local/share/ai-memory/hooks/claude-code` / `Permission denied (os error 13)`.

**Environment**
- OS: Linux Mint 22.3 (Ubuntu 24.04 base)
- Docker? yes, rootless — `docker info --format '{{.SecurityOptions}}'` reports `name=rootless`; Docker 29.1.3, image `akitaonrails/ai-memory:latest`
- Agent: Claude Code
- Transport: http

**Relevant logs**
```
Error: creating staging dir /home/holocaster/.local/share/ai-memory/hooks/claude-code

Caused by:
    Permission denied (os error 13)
```

**Root cause**
`bin/ai-memory` always runs `docker run -u "$(id -u):$(id -g)" ...`. Under rootless Docker (rootlesskit), container UID **0** maps back to the real host UID via the primary user-namespace mapping, but any *non-zero* UID passed via `-u` — including the host UID passed explicitly here — is instead routed through the subordinate-UID range (`/etc/subuid`, typically `100000+`), which owns nothing on the host filesystem. So every bind-mounted write from a helper container invoked with the host UID lands as an unrelated, unmapped UID and fails.

Verified directly:
```
$ docker run --rm --entrypoint sh -u 0:0 -v "$PWD:/t" akitaonrails/ai-memory:latest -c "touch /t/probe; id"
uid=0(root) gid=0(root) groups=0(root)
$ stat probe
Uid: ( 1000/holocaster)   Gid: ( 1000/holocaster)   # correct — root maps back to host user

$ docker run --rm --entrypoint sh -u 1000:1000 -v "$PWD:/t" akitaonrails/ai-memory:latest -c "touch /t/probe2; id"
uid=1000(ai-memory) gid=1000(1000) groups=1000(1000)
$ stat probe2
Uid: (100999/ UNKNOWN)   Gid: (100998/ UNKNOWN)  # wrong — unrelated subordinate-range UID
```

**Fix**
Opened as #150: detect rootless Docker via `docker info --format '{{.SecurityOptions}}' | grep -q rootless` and use `-u 0:0` instead of the host UID for just the subcommands that write host-side files (`install-mcp`, `install-hooks`, `setup-agent`, `install-instructions`, `install-skills`). Thin-client commands (`status`, `bootstrap`, `search`, …) are unaffected since they only touch the `/data` named volume, not a host bind mount.

holocaster:
Fixes #151

## Summary

Rootless Docker maps container UID **0** back to the real host UID via rootlesskit's primary user-namespace mapping, but any *non-zero* UID passed via `-u` — including the host UID the wrapper always passes — is instead routed through a separate subordinate-UID range (`/etc/subuid`, typically `100000+`) that owns nothing on the host filesystem. So every `bin/ai-memory` subcommand that writes to a host bind-mounted path (`~/.claude/settings.json`, `~/.claude.json`, the hook-staging dir, `$PWD/CLAUDE.md`, skill directories) failed under rootless Docker with `Permission denied` or a misleading "does not exist" error.

This fixes it: detect rootless Docker via `docker info --format '{{.SecurityOptions}}'` and run as `-u 0:0` only for the subcommands that write host-side files (`install-mcp`, `install-hooks`, `setup-agent`, `install-instructions`, `install-skills`). Thin-client commands (`status`, `bootstrap`, `search`, …) are untouched — they only touch the `/data` named volume, which isn't host-visible and has no such issue.

Verified manually against the published image:
```
$ docker run --rm --entrypoint sh -u 0:0 -v "$PWD:/t" akitaonrails/ai-memory:latest -c "touch /t/probe; id"
uid=0(root) gid=0(root) groups=0(root)
$ stat probe   # Uid: (1000/holocaster) — correct, root maps back to host user

$ docker run --rm --entrypoint sh -u 1000:1000 -v "$PWD:/t" akitaonrails/ai-memory:latest -c "touch /t/probe2; id"
uid=1000(ai-memory) gid=1000(1000) groups=1000(1000)
$ stat probe2  # Uid: (100999/UNKNOWN) — wrong, unrelated subordinate-range UID
```

## Versioning

Per `CONTRIBUTING.md`'s semver policy: this is a **patch**-level bug fix. No public API, CLI surface, or on-disk format changes — `bin/ai-memory`'s behavior is identical for rootful Docker (the common case) and only changes the UID for host-config-writing commands specifically under rootless Docker, where those commands were already broken.

## Test plan

- [x] `cargo test --workspace` (includes 2 new tests: `rootless_docker_uses_root_uid_only_for_host_config_commands`, `rootful_docker_keeps_host_uid_for_host_config_commands`, both using a fake `docker` binary that dispatches on `info` vs `run` to simulate the rootless/rootful distinction)
- [x] `cargo deny check`
- [x] Reproduced and fixed the failure against the real published image on rootless Docker (Linux Mint 22.3, Docker 29.1.3)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- `bin/ai-memory` no longer fails `install-mcp`, `install-hooks`,
+  `setup-agent`, `install-instructions`, and `install-skills` under
+  rootless Docker. Rootless Docker maps container UID 0 back to the real
+  host user but routes any other UID (including the host UID the wrapper
+  always passed via `-u`) through an unrelated subordinate-UID range, so
+  every write to a bind-mounted host path (`~/.claude/settings.json`,
+  the hook staging dir, `$PWD/CLAUDE.md`, skill directories) failed with
+  `Permission denied` or a misleading "does not exist" error. The wrapper
+  now detects rootless Docker via `docker info --format
+  '{{.SecurityOptions}}'` and runs as `-u 0:0` for just the commands that
+  write host-side files; thin-client commands (`status`, `bootstrap`, …)
+  are unaffected since they only touch the `/data` named volume.
+
 ## [1.8.0] - 2026-07-04
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `bin/ai-memory` no longer fails `install-mcp`, `install-hooks`,
-  `setup-agent`, `install-instructions`, and `install-skills` under
-  rootless Docker. Rootless Docker maps container UID 0 back to the real
+  `setup-agent`, `install-instructions`, `install-skills`, `uninstall`,
+  and `backup` under rootless Docker. Rootless Docker maps container UID
+  0 back to the real
   host user but routes any other UID (including the host UID the wrapper
   always passed via `-u`) through an unrelated subordinate-UID range, so
   every write to a bind-mounted host path (`~/.claude/settings.json`,
--- a/bin/ai-memory
+++ b/bin/ai-memory
@@ -257,7 +257,8 @@ done
 NETWORK_ARGS=()
 # Default: map the container process to the host user so files written
 # through bind mounts (~/.claude/settings.json, $PWD/, …) stay editable
-# by the invoking user. macOS is the one exception (see Darwin arm).
+# by the invoking user. macOS is one exception (see Darwin arm); rootless
+# Docker is the other (see the WRITES_HOST_FILES check below).
 USER_ARGS=(-u "$(id -u):$(id -g)")
 
 # Does this invocation RENDER host-side agent config? install-mcp,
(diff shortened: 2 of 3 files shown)


akitaonrails:
Thanks @holocaster — the touch/stat probes against the published image made the subordinate-UID diagnosis indisputable, and putting the rule behind the WRITES_HOST_FILES marker gave the fix a clean extension point.

Used exactly that extension point pre-merge (0533d31): `uninstall` edits the same host agent-config files and `backup` writes its tarball to a host path, so both now get the same rootless `-u 0:0` treatment, with matching rows in the regression test. Merged as 30c1e45; #151 auto-closed.

PR #165, opened 2026-07-10
chore(ci): unbreak main — clear RUSTSEC-2026-0204 and fix the macOS rootless test

djalmajr:
CI is currently red on `main` (and therefore on every open PR) on two independent, gating checks. This bundles both fixes so `main` goes green in one merge.

## 1. `cargo-audit` / `cargo-deny` — RUSTSEC-2026-0204
Invalid pointer dereference in the `fmt::Pointer` impl for crossbeam-epoch's `Atomic`/`Shared`. `crossbeam-epoch` is transitive (`sysinfo -> rayon -> rayon-core -> crossbeam-deque`). Bumping the lockfile entry `0.9.18 -> 0.9.20` (a compatible patched release) clears the advisory at the source — no `--ignore` entry needed.

Verified: `cargo audit` (with the existing CI ignore set) exits 0; `cargo deny check advisories` → `advisories ok`.

## 2. `test (macos-latest)` — rootless-Docker UID test
`rootless_docker_uses_root_uid_only_for_host_config_commands` asserts the Linux rootless subordinate-UID behavior (`-u 0:0` for host-config subcommands), but read the **real** host OS via `uname -s`. On a macOS runner the wrapper's `Darwin` arm deliberately clears `USER_ARGS` (Docker Desktop handles bind-mount perms), so the test hit the wrong branch and failed. Fixed by installing a fake `uname` reporting `Linux` on `PATH` inside the test helper — the same technique `run_wrapper_on_fake_macos` already uses — so the Linux path is exercised deterministically on any runner.

Verified on macOS (where it failed): `cargo test -p ai-memory-cli --test packaging` is fully green (9/9). Test-harness only; no wrapper/production code changed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -696,9 +696,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-epoch"
-version = "0.9.18"
+version = "0.9.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b82ac4a3c2ca9c3460964f020e1402edd5753411d7737aa39c3714ad1b5420e"
+checksum = "2d6914041f254d6e9176c01941b21115dcfb7089e55135a35411081bd106ef3f"
 dependencies = [
  "crossbeam-utils",
 ]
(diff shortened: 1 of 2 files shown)


djalmajr:
Superseded by #162 (already merged): it bumps crossbeam-epoch to 0.9.20 for RUSTSEC-2026-0204 and fixes the macOS rootless behavior in the wrapper itself (plus the Windows uninstall test). Closing as redundant.

PR #174, opened 2026-07-12
docs(instructions): resync AGENTS.md with the slim routing snippet + guard drift

djalmajr:
The committed root `AGENTS.md` still carried the pre-slim managed block (inline tool-routing table, "Use Retrieved Memory As Operating Guidance", "Learning Review") while `SNIPPET_BODY` had been slimmed to defer detail to the installed Agent Skills. Nothing forced the committed copy to track the generator, so agents and forks reading `AGENTS.md` directly inherited stale routing guidance.

## Changes
- **Regenerate `AGENTS.md`** from the current `SNIPPET_BODY` (`ai-memory install-instructions --target AGENTS.md --no-skills`), preserving all content outside the `ai-memory:start`/`ai-memory:end` markers (48 insertions / 117 deletions — the old long block collapses to the slim one).
- **Add a `routing_snippet` unit test** (`committed_agents_md_matches_snippet_body`) that extracts the managed block from the committed `AGENTS.md` (via the existing `find_marker_line` helper) and asserts it equals `full_block()`. It goes red whenever the two drift and names the regeneration command in the failure message. Only `AGENTS.md` is checked (root `CLAUDE.md` is a pointer; `README.md` is prose).

## Validation
- `cargo test -p ai-memory-core routing_snippet` → 5/5 green (incl. the new guard).
- `cargo fmt -p ai-memory-core -- --check` clean; `cargo clippy -p ai-memory-core --all-targets -- -D warnings` clean.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,135 +4,66 @@
 This project uses [ai-memory](https://github.com/akitaonrails/ai-memory)
 for cross-session continuity.
 
-**Default to the current project — always.** Every ai-memory tool
+**Default to the current project - always.** Every ai-memory tool
 auto-scopes to the project resolved from your session's working
-directory. **Do NOT pass `project`, `workspace`, or `cwd` arguments unless the user
-explicitly references a *different* project by name** (e.g. "what did we
-decide in the `other-app` project?"). Phrases like "this project",
-"here", "we", "our work", "where did we leave off" all mean the *current*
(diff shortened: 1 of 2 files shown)


akitaonrails:
Thanks @djalmajr — the drift-guard test that names the regeneration command in its failure message is exactly how a sync invariant should be enforced. Merged as-is.

PR #175, opened 2026-07-12
feat(store): audit-log attribution for purge-project and rename-project

djalmajr:
The append-only `audit_log` only recorded page upserts (create/supersede) and decay soft-deletes. The destructive/administrative **project** ops wrote nothing, so the exact question V16 was added to answer — *"which user wiped project X?"* — had no answer.

## Changes
- `ops::purge_project` and `ops::rename_project` take an `author_id` and write an attributed `audit_log` row **inside their transaction**. `rename_project` is now transactional, so the audit row commits atomically with the `UPDATE` — a name-collision rollback leaves no row.
- The authenticated operator (`Extension<UserId>`, `NULL` when single-user / unauthenticated) is threaded from the admin handlers (`/admin/purge-project`, `/admin/rename-project`) through the writer actor into the store.
- Internal sub-purges — move-project's copy-purge step and the hook router self-heal — pass `author_id = None`; they are distinct ops, not standalone operator purges.

## Tests
- `audit_log_records_purge_project_with_author` / `_rename_project_with_author`: exactly one attributed row per op.
- `audit_log_omits_rename_project_on_collision`: a rolled-back rename writes no row (proves tx atomicity).

## Scope
`delete_page` and handoff transitions stay unaudited for now (follow-up): `delete_page` needs a wider `&Connection`→`&mut Connection` signature change rippling to many call sites, and single-page deletes are recoverable via the git mirror — so they ship separately to keep this focused on the **irreversible** project-level ops.

## Validation
- `cargo test -p ai-memory-store` (157) / `-p ai-memory-mcp` (121) / `-p ai-memory-hooks` (110) all green.
- `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean across the three crates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `purge-project` and `rename-project` now write attributed rows to the
+  append-only `audit_log`, inside the same transaction as the operation
+  itself — so "which user wiped project X?" finally has an answer, and a
+  rolled-back rename (name collision) leaves no phantom trail. The
+  operator identity comes from the authenticated admin request and is
+  `NULL` for single-user/unauthenticated servers; internal sub-purges
+  (move-project's copy-purge step, the hook router's self-heal) are
+  deliberately not attributed as standalone purges ([#175]).
+
 ## [1.11.4] - 2026-07-11
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


akitaonrails:
Thanks @djalmajr — making the rename transactional specifically so a collision rollback can't strand a phantom audit row (and proving it with a test) is the detail that makes an audit trail trustworthy. Merged with a CHANGELOG fixup (54a066a) pushed pre-merge.

PR #177, opened 2026-07-12
feat(recall): opt-in `[recall] default_global` to broaden default-scoped queries

djalmajr:
A meta-repo (e.g. `ai-memory` itself, which must see `ai-memory-ops` / `infra` constantly) had to pass `global=true` by hand on every `memory_query`. This adds a per-repo opt-in: `[recall] default_global = true` in a project's `.ai-memory.toml` makes a **default-scoped** `memory_query` behave as `global=true`.

## End-to-end wiring (reuses the `drop_subagent` pipeline)
- **CLI marker parse** (`hook_capture`): parse `default_global` — quoted or bare, section-agnostic so `[recall] default_global = true` works — and forward it as a `default_global` query flag on the `/hook` POST.
- **Hook envelope** (`payload`): `HookQuery.default_global` → `HookEnvelope.recall_default_global_requested` (truthy interpretation, shared with `drop_subagent`).
- **`ActiveProject`** (core): the per-actor entry (now a named `Entry`) and the single fallback slot carry a `default_global` bool, published alongside the project pointer by the hook router. The `get_for` tuple-resolution path is **byte-for-byte unchanged**; a new `default_global_for(actor)` reads the flag with the *same* slot dispatch and shares TTL/cap eviction.
- **`memory_query`** (mcp): when the caller passes **no** explicit scoping (`global`/`scopes`/`workspace`/`project`) **and** the actor opted in, route to the existing global search. **Strict precedence:** an explicit arg always wins.

## Safety
Opt-in and default-`false` — a repo without the marker is byte-identical to today. `resolve_project_ids` keeps its signature via a `#[cfg(test)]` wrapper; production paths call `resolve_project_ids_inner(..., default_global)`, so the ~70 existing call sites are untouched.

`memory_recent` stays project-scoped for now (a global "recent" needs a new cross-project reader) — documented follow-up.

## Tests
Marker parse (append/omit), envelope mapping, `ActiveProject` round-trip + isolation + false-default, and an **end-to-end `memory_query` gate** (unscoped query broadens to global; explicit `workspace`+`project` still scopes to one project).

## Validation
`cargo test` on core/hooks/cli/mcp — all green; `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- New `.ai-memory.toml` marker option `[recall] default_global = "true"`
+  ([#177]): sessions in the marked tree make an *unscoped* `memory_query`
+  behave as `global=true`, so meta-repos that constantly need
+  sibling-project context stop passing `global=true` by hand. Strictly
+  opt-in and per-repo; explicit `workspace`/`project`/`scopes`/`global`
+  arguments always win. While active, unscoped queries return
+  cross-project `global_hits` (workspace+project annotated) instead of
+  project `hits` + `global_scope_hits`; `memory_recent` stays
+  project-scoped (documented follow-up).
 - `purge-project` and `rename-project` now write attributed rows to the
   append-only `audit_log`, inside the same transaction as the operation
   itself — so "which user wiped project X?" finally has an answer, and a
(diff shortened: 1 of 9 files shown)


akitaonrails:
Thanks @djalmajr — the strict-precedence gate and especially forwarding the flag through the session-sticky publish path (which landed only days ago) made this audit easy; the ActiveProject Entry refactor keeping get_for byte-identical was verified. Merged with a fixup (fmt + CHANGELOG + marker-file.md docs) pushed pre-merge.

PR #178, opened 2026-07-13
feat(devin): add first-class Devin CLI integration

lhzapata:
## Summary

This PR adds Devin as a first-class ai-memory agent instead of treating it as just another HTTP MCP client.

Devin already exposes useful integration points through MCP, hooks, skills, and its own project context model. It is built by Cognition and includes workflows such as DeepWiki, but the important part for ai-memory is simpler: Devin emits different hook payloads and benefits from explicit session, cwd, and post-compaction handling.

## What changed

- Added `devin` as a supported agent kind across parsing, storage, migrations, and tests.
- Added Devin hook support for `SessionStart`, `PostToolUse`, and `PostCompaction`.
- Added robust handling for real Devin payloads that may omit `session_id` and `cwd`.
- Added Devin MCP, hook, and skills installation paths.
- Added Devin uninstall and reinstall coverage, including flat `.devin/hooks.v1.json` behavior.
- Fixed a bug where an explicit `--server-url` matching the compiled default was silently overridden by `AI_MEMORY_SERVER_URL`, which could point config at the wrong running server instead of the one actually requested.
- Updated docs and changelog for Devin support.

## Why this is different from MCP-only

As-is, ai-memory could already integrate with Devin over HTTP, but that required manually pointing Devin at the right MCP address. Beyond that manual setup, the old path did not give ai-memory enough structure to reliably connect Devin events into a useful session story.

The first-class path adds that missing structure: real hook events are captured, events without `session_id` are still accepted, cwd is derived and documented, PostCompaction summaries become durable session checkpoints, and handoff context can be injected back into Devin through hook output.

## Validation

All gates below ran against this PR's full diff.

- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] `cargo deny check`
- [x] `TAILWIND_SKIP=1 cargo test --workspace --locked`
- [x] `TAILWIND_SKIP=1 cargo clippy --workspace --locked --all-targets -- -D warnings`
- [x] `cargo build --locked --release -p ai-memory-cli`
- [x] `ai-memory --version`
- [x] Same gate from a clean clone on Arch Linux (`cargo test` passes except two pre-existing tests that assume non-root execution and fail under any root-level test run; unrelated to Devin)
- [x] Real Devin acceptance runs: preflight and MCP visibility, MCP-only baseline, full Devin candidate integration, real `PostCompaction`, negative cases for unavailable server, invalid MCP URL, malformed payload, historical ledger memory, repeated reinstall, and uninstall preservation
- [x] Linux multi-agent battery ran Claude, Codex, and Devin in an Arch container.

## Known limitations

- Devin does not currently expose subagent hooks in the same shape as Claude, so the integration documents that boundary instead of pretending it exists.
- The global (user-level) Windows Devin skills install path is only covered by shape/unit tests that check the file gets written correctly; we did not run a live Devin session loading skills from a global-scope install on Windows, since every real-world acceptance run used a project-local install. Exercising the global path end-to-end is a follow-up.
- In the Linux multi-agent battery, Devin's baseline run without ai-memory installed still saw the MCP tools listed but could not call them, apparently blocked by `--permission-mode`. That makes the Linux "without ai-memory" comparison for Devin a lower bound rather than a clean baseline; revisiting this is a follow-up before drawing strong conclusions from it.
- `setup-agent` keeps its existing rendering path and should be audited separately if exact default `server_url` overriding behavior becomes part of its contract.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,3 @@ evals/runs/
 
 # ai-memory atomic-write backups (apply_atomic in apply_shared.rs)
 *.bak-*
-
-# SonarQube scanner working directory.
-.scannerwork/
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ evals/runs/
 
 # ai-memory atomic-write backups (apply_atomic in apply_shared.rs)
 *.bak-*
+
+# SonarQube scanner working directory.
+.scannerwork/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   payload `cwd` still wins, followed by `DEVIN_PROJECT_DIR`, then the hook
   process working directory. This keeps real Devin `SessionStart` /
   `PostToolUse` fixtures routable without inventing a payload field.
+  Payloads without a `session_id` are bridged the same way: a per-host id is
+  minted at `SessionStart`, reused for later events, and cleared at
+  `SessionEnd`; set `AI_MEMORY_SESSION_ID` in the hook environment to pin an
+  externally managed run id. A payload-supplied id always wins.
 
 ## [1.12.0] - 2026-07-12
 
@@ -282,6 +286,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   '{{.SecurityOptions}}'` and runs as `-u 0:0` for just the commands that
   write host-side files; thin-client commands (`status`, `bootstrap`, …)
   are unaffected since they only touch the `/data` named volume.
+
 ## [1.8.0] - 2026-07-04
 
 ### Added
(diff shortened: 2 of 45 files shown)


akitaonrails:
Thanks @lhzapata — this is one of the most complete agent integrations this repo has received. V28 following the V26 rebuild pattern exactly (triggers recreated, `zero` carried forward, and the upgrade-chain test running from a pre-V26 database) meant the migration needed zero changes, and the Devin-scoped session/cwd bridging with body-values-always-win precedence was easy to verify because every branch had a test. The `--server-url` Option fix was a genuinely good catch beyond the Devin scope.

Merged with a small maintainer fixup on top of your rebase: uninstall now also sweeps the Windows global `%APPDATA%\devin\skills` root, the `DEVIN_PROJECT_DIR` / `AI_MEMORY_SESSION_ID` fallbacks (and the single-slot concurrent-sessions caveat) are documented in docs/mcp-install.md + CHANGELOG, plus minor test/dedup nits.

PR #179, opened 2026-07-13
feat(store): audit-log attribution for delete-page + handoff lifecycle

djalmajr:
Completes the `audit_log` follow-up after #175 (which covered `purge_project` / `rename_project`). The remaining destructive / state-transition ops now write audit rows too, so the append-only trail finally answers *"who deleted the gotcha page about X?"* — V16's headline case.

## Changes
- **`delete_page`** takes `&mut Connection` + an `author_id`, captures the latest page id **before** the `DELETE`, and writes an attributed `delete_page` audit row — **only on a real deletion** (a no-op delete of a missing page writes nothing, so idempotent misses do not pollute the trail). The operator (`Extension<UserId>`, `NULL` when single-user / unauthenticated) is threaded from `memory_delete_page` and the admin `delete-page` handler through the writer actor.
- **`insert` / `accept` / `cancel_handoff`** wrap their write in a transaction and audit the lifecycle (`op` + `workspace`/`project`), scoped via a small `handoff_scope` lookup. The author is **NULL by design** — handoffs are agent/session-keyed, not owned by a DB user.

## Tests
- `audit_log_records_delete_page_with_author_and_page_id` — one attributed row pointing at the deleted page id.
- `audit_log_omits_delete_page_noop` — a no-op delete writes no row.
- `audit_log_records_handoff_lifecycle_with_null_author` — insert/accept/cancel each write a scoped row with a NULL author.

## Validation
`cargo test` on store (161) / wiki / mcp (122 + 83) — all green; `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `memory_delete_page` / admin `delete-page` now write one attributed
+  `audit_log` row pointing at the deleted page id, in the same transaction
+  as the delete — completing the "who deleted the gotcha page about X?"
+  trail that purge/rename attribution started. Idempotent no-op deletes
+  write nothing. The handoff lifecycle (insert / accept / cancel) is also
+  audited, scoped to the handoff's workspace/project with a NULL author by
+  design — handoffs are agent/session-keyed, not owned by a DB user
+  ([#179]).
+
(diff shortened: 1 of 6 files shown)


akitaonrails:
Thanks @djalmajr — the no-op discipline (idempotent delete misses and non-transitions audit nothing) and capturing the page id before the DELETE inside the same transaction made this land without a single change to the logic. Merged with a small fixup pushed pre-merge (CHANGELOG entry + regression assertions for the no-op transition guards).

PR #183, opened 2026-07-13
fix(consolidate): attribute memory_consolidate writes to the request actor

djalmajr:
`memory_consolidate` wrote its consolidated page as a **hard-coded anonymous** actor: the MCP handler never extracted the request identity, so the write carried no user. Two consequences on any deploy:
1. the on-disk page's `last_modified_by` was always empty (wrong attribution), and
2. an actor-gated admission webhook rejects the write (empty actor) even though `memory_write_page` on the same connection — which *does* carry the actor — is accepted.

## Change
- `memory_consolidate` now takes the request `Parts` and derives `actor` + `author_id` (the exact pattern `memory_write_page` uses).
- `consolidate_session` / `consolidate_session_multi` / `build_update` thread the `actor` + `author_id` into the `WritePageRequest` **and its admission context**, replacing the hard-coded `ActorContext::anonymous()`.
- The **automatic** session-end consolidation in the hook router is system-initiated (no request actor), so it stays anonymous — deliberately unchanged.

## Test
`build_update_stamps_request_actor_and_author`: the passed actor + author_id land on the request and its admission context (not anonymous).

## Scope
This is the actor-propagation half of the consolidate follow-up. Two related, independent improvements remain as follow-ups: resolving the consolidation target from where the session's observations actually landed (not a frozen session row), and running the admission preflight *before* the LLM call so a rejected scope fails fast without burning an LLM round-trip.

## Validation
`cargo test -p ai-memory-consolidate -p ai-memory-hooks -p ai-memory-mcp` green · `cargo fmt --check` · `cargo clippy --all-targets -- -D warnings` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- `memory_consolidate` now attributes its consolidated page to the request's
+  authenticated identity (same derivation as `memory_write_page`) instead of
+  a hard-coded anonymous actor — so `last_modified_by` is populated and an
+  actor-gated admission webhook authorizes the write by user rather than
+  rejecting the empty actor. The automatic session-end / compaction
+  consolidation in the hook router is system-initiated and deliberately
+  stays anonymous ([#183]).
+
 ### Changed
 - `audit-contamination` no longer flags an observation whose project differs
   from its session's home project (the old `observation_session_drift` CHECK
--- a/crates/ai-memory-consolidate/src/consolidator.rs
+++ b/crates/ai-memory-consolidate/src/consolidator.rs
@@ -103,6 +103,8 @@ impl Consolidator {
         &self,
         session_id: SessionId,
         dry_run: bool,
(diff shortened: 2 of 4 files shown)


akitaonrails:
Thanks @djalmajr — all four of today's PRs (#180, #181, #182, #183) are merged. Each was tight: the #177-identical strict precedence in #181, the correct per-event-cwd reasoning behind dropping CHECK B in #182, and reusing the exact `memory_write_page` identity derivation here all made the audits fast.

One ask for next time: each PR again landed without a `CHANGELOG.md` `[Unreleased]` entry, and I pushed a fixup to each branch pre-merge (plus, for this one, the main-merge resolving the overlap with #178's checkpoint-label refactor, and for #181 the marker-file.md doc line + a shared precedence helper). The PR template now has an explicit CHANGELOG checkbox and CONTRIBUTING.md documents it as a merge gate — ticking that box is the one thing between your PRs and a first-pass merge.

djalmajr:
Understood... thanks for the fixups on #180-183, and apologies for the repeated CHANGELOG omission. I've folded the [Unreleased] entry into my flow: #186 and #187 both shipped with their entries and the template checkbox ticked, so those shouldn't need a fixup pass. Appreciate the tight reviews.

PR #184, opened 2026-07-14
fix(store): clear error when data schema is newer than the binary

lhzapata:
## The problem

Opening a store whose schema is newer than the running binary currently fails with refinery's raw error text:

```
migration V28__sessions_devin_agent_kind is missing from the filesystem
```

That wording is misleading. Nothing is actually missing from disk — migrations are embedded into the binary at compile time via `refinery::embed_migrations!`. What really happened is the opposite of what the message implies: the *data* is ahead of the *binary*. Some other, newer ai-memory build already migrated this store to a schema version this binary has never heard of.

This is easy to hit in practice: point an older release binary at a data directory that a newer build (or a beta/pre-release build) already migrated forward, and you get a confusing error that reads like a missing file on disk rather than "run a newer build."

## The fix

The store layer now inspects refinery's error kind directly. When it's the `MissingVersion` case — an applied migration absent from this binary's compiled-in set — it's remapped to a new `StoreError::DataSchemaAhead` variant with an actionable message:

```
memory database schema is newer than this ai-memory build: the store has
migration V28 (sessions_devin_agent_kind) applied, but this build only ships
migrations through V26. Run an ai-memory release at least as new as the one
that wrote this data; a newer store cannot be opened by an older binary.
```

Every other migration failure is untouched and still surfaces as `StoreError::Migration`.

## Tests

- One test forges a future-version row directly in `refinery_schema_history` (the only way to reliably reproduce `MissingVersion` from real refinery behavior, since the error type can't be constructed externally) and asserts the real migration run maps it to `DataSchemaAhead`.
- A second test pins the rendered message text, so it can't silently regress back to refinery's raw wording.

No other behavior changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Opening a store whose schema is *newer* than the running binary now fails
+  with an actionable error instead of refinery's raw "migration V… is missing
+  from the filesystem" wording. When an applied migration is absent from the
+  binary's compiled-in set (the data was migrated by a newer ai-memory build),
+  the store layer now returns `DataSchemaAhead`, which names the offending
+  migration, reports the highest schema version this build ships, and tells the
+  operator to run a build at least as new as the one that wrote the data. Every
+  other migration failure is unchanged.
+
 ## [1.13.0] - 2026-07-14
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


akitaonrails:
Thanks @lhzapata — merged with zero fixups: the forged `refinery_schema_history` row was the right way to exercise the real `MissingVersion` path, the pinned-message test keeps the wording from regressing, and the CHANGELOG entry arrived complete on first submission (first PR through the new template checklist — appreciated).

PR #187, opened 2026-07-15
feat(admin): merge-workspace — fold one workspace's projects into another

djalmajr:
## What

Adds `POST /admin/merge-workspace {from, to, confirm}` — fold every project of one workspace into another, then delete the emptied source workspace. This completes the workspace CRUD surface next to the existing rename/delete-workspace endpoints.

## Why

`move-project` get-or-creates its destination workspace, so consolidating two workspaces (or cleaning up a typo'd/temporary one) meant moving projects **one at a time by hand** and then being left with an empty shell that nothing removes. `merge-workspace` does the whole fold in one call.

## How

The merge is sugar over move-project, not a new destructive path. Two existing handlers are split into a reusable **core** that returns its report/summary as data plus a thin HTTP wrapper:

- `move_project_core(state, req, actor) -> Result<MoveProjectReport, (StatusCode, Json)>` — all the move validation + true-move / copy-purge dispatch.
- `delete_workspace_core(state, workspace, force, actor) -> Result<DeleteWorkspaceResult, (StatusCode, Json)>`.

`handle_move_project` / `handle_delete_workspace` become one-line wrappers over their cores, so their existing HTTP behavior is unchanged (every prior move/delete test stays green).

`handle_merge_workspace` then:
1. Requires `confirm=true` (destructive); rejects `from == to`; 404s an unknown source (no phantom-workspace creation).
2. Lists the source's projects and runs `move_project_core` for each — a project with no destination twin is a lossless **true-move** (re-stamp `workspace_id`, keep `project_id`, rename the dir); a project that already exists in the destination is a **copy-purge** content merge under `on_conflict` (default `block`). `force` overrides the per-project live-session guard.
3. Deletes the now-empty source workspace via `delete_workspace_core(..., force=false)`.

**Failure semantics:** each per-project move is independently committed, so the merge stops at the first failing project and reports what already moved — the failing project and the source workspace are left intact for the operator to resolve and re-run (moves are idempotent). If every project moves but the final empty-shell delete races a concurrent write, the response is success-with-caveat (`source_workspace_deleted: false`) rather than a hard error, since no data was lost.

Admin-only, matching rename/delete-workspace (no CLI subcommand, consistent with those endpoints).

## Tests

- Folds multiple fresh projects into a new destination and deletes the empty source (all true-moves).
- A project that collides with the destination merges by copy-purge while a solo project true-moves — both pages survive under the merged project.
- Missing `confirm` is refused with nothing moved.
- `fmt` + `clippy -D warnings` clean; full `ai-memory-mcp` suite green (existing move/delete-workspace tests unchanged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- New admin endpoint `POST /admin/merge-workspace {from, to, confirm}`: fold
+  every project of one workspace into another, then delete the emptied source
+  workspace — completing the workspace CRUD surface alongside rename/delete.
+  It is sugar over move-project: each source project runs the same validated
+  path, so a project that already exists in the destination merges by content
+  (copy-purge under `on_conflict`) while a fresh one is a lossless re-stamp.
+  Destructive, so `confirm=true` is required; `force` overrides the
+  live-session guard per project. It stops at the first failing project — the
+  moves already committed stand, and the failing project plus the source
+  workspace are left intact for the operator to resolve and re-run (moves are
+  idempotent) ([#187]).
+
 ### Fixed
 - `install-mcp --server-url` now appends the `/mcp` path when given a base
   URL (the same value `install-hooks --server-url` takes), instead of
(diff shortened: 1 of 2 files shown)


akitaonrails:
Thanks @djalmajr — both PRs merged. #186's PanicLlm proof-tests (dry_run literally cannot reach the model) and #187's core/wrapper split keeping every existing move/delete test green untouched are exactly the shapes that make audits fast. Also: second batch in a row with complete, house-style CHANGELOG entries on arrival — the first-pass merges are the payoff.

#186 got one maintainer fixup pre-merge: docs/admission-webhooks.md now documents that consolidate ops reach blocking webhooks up to twice (empty-body preflight before the LLM, then the real write-time check — decisions, not delivery events), plus a small error-mapping cleanup in the new reader method. #187 merged as-is with just the main sync.

Issue #155, opened 2026-07-08
memory_read_page: path/query both optional in schema breaks OpenCode (path=null loop)


JonasGz:
## Summary

`memory_read_page` declares **both** `path` and `query` as optional (`type: ["string", "null"]`, `default: null`) while the tool actually requires **one of them**. MCP clients that omit/serialize null-default args (OpenCode is the one I hit) end up sending neither, so the server errors with *"neither argument supplied"*. The model doesn't recover and retries the identical call in a tight loop (60+ times observed).

Server + tool are fine — the exact same MCP server works when the argument is provided. The failure is purely the schema not encoding the "exactly one of path/query" constraint, which some clients handle worse than others.

## Environment

- ai-memory `1.8.0` (Docker, multi-user, loopback `127.0.0.1:49374`)
- Clients: **OpenCode 1.17.15** (fails) and **Claude Code** (works)
- MCP config generated by `ai-memory install-mcp --client opencode` (verified byte-for-byte against the official snippet — nothing custom)

## Reproduction

1. Configure OpenCode with the official ai-memory MCP (`type: remote`).
2. Ask it to read a specific page (e.g. after `memory_recent` returns a hit): "read page notes/onde-paramos.md".
3. OpenCode calls `memory_read_page` with `path=null` (and no `query`).
4. Server errors → model loops the same call indefinitely.

Trace (abbreviated):

```
memory_recent [limit=15]        -> OK, returns the page + path
memory_read_page [path=null]    -> error
memory_read_page [path=null]    -> error
... (repeats 60+ times)
```

## Cross-check (same server, same page)

From **Claude Code**, `memory_read_page { path: "notes/onde-paramos.md" }` returns the full body immediately. So this is client-dependent behavior triggered by the all-optional schema, not a server bug. Two different OpenCode models (deepseek-v4-flash-free and glm-5.2) fail identically, ruling out model capability.

## Suggested fix (pick one)

1. **Encode the constraint in the schema** — a JSON-Schema `oneOf`/`anyOf` requiring at least one of `path`/`query`, so clients that respect it are forced to send one.
2. **Split into two tools** — e.g. `memory_read_page_by_path` (required `path`) and `memory_search_page` (required `query`). Required args can't be dropped by clients; this is why `memory_query` (required `query`) never exhibits the problem.
3. At minimum, make the tool description explicitly state "you MUST pass either `path` or `query`" (current text says "Errors if ... neither argument is supplied" which the model ignores).

[...]

PR #195, opened 2026-07-17
feat(agents): add Kimi Code as a first-class supported agent

lucasliet:
## Motivation

Kimi Code CLI is Moonshot's coding agent CLI, with Claude-Code-style hooks and MCP support. Its config formats differ enough from the existing integrations that none of them apply out of the box: hooks live as `[[hooks]]` TOML entries inside the main `config.toml` (the same file that holds provider/model settings), and MCP servers live in a separate `mcp.json` where a bare `url` field already means streamable HTTP.

Docs: [hooks](https://www.kimi.com/code/docs/en/kimi-code-cli/customization/hooks.html) · [mcp](https://www.kimi.com/code/docs/en/kimi-code-cli/customization/mcp.html)

## Before

`ai-memory install-mcp --client kimi-code` / `ai-memory install-hooks --agent kimi-code` didn't exist. Users had to hand-craft both:

- MCP: `~/.kimi-code/mcp.json` with an `mcpServers` entry
- Hooks: `[[hooks]]` entries in `~/.kimi-code/config.toml`, carefully preserving the provider/model settings the same file holds (only `event`/`matcher`/`command`/`timeout` are allowed per entry — extra keys make the whole config fail to load)

## After

`kimi-code` is a first-class client/agent (alias `kimi`) alongside Claude Code, Codex, Gemini CLI, Antigravity CLI, etc.:

```bash
ai-memory install-mcp --client kimi-code --apply
ai-memory install-hooks --agent kimi-code --apply
```

Key differences handled automatically:

| Aspect | Claude Code | Kimi Code |
|---|---|---|
| MCP config file | `~/.claude.json` | `~/.kimi-code/mcp.json` (honors `$KIMI_CODE_HOME`) |
| MCP HTTP entry | `{ "type": "http", "url", "headers" }` | `{ "url", "headers" }` (bare `url` = streamable HTTP; `transport` is for legacy SSE only) |
| MCP URL written | `http://host/mcp` | `http://host/mcp?flavor=moonshot` (see note below) |
| Hooks config file | `~/.claude/settings.json` (JSON) | `~/.kimi-code/config.toml` (TOML `[[hooks]]` entries, merged preserving provider/model settings and third-party hooks) |
| Hook events | 9 events | Claude Code's 9 plus `PostToolUseFailure` — Kimi Code fires `PostToolUse` on successful calls only, so failed tool outcomes ride the failure event (same post-tool-use capture path) |
| Hook stdin payload | `{hook_event_name, session_id, cwd, tool_input}` | Same envelope, but `prompt` arrives as Anthropic-style content blocks (`[{"type":"text","text":...}]`) — the user-prompt title/body extraction flattens them |
| Handoff injection | `hookSpecificOutput.additionalContext` | Plain stdout at `SessionStart` (Kimi Code appends exit-0 stdout to context) |

**Moonshot schema flavor.** Kimi Code talks to Moonshot, whose tool-parameter validator ("moonshot flavored json schema") rejects any root-level `anyOf`/`oneOf`/`allOf` in a function schema with a 400 at `tools/list` time — that includes the issue-#155 `anyOf` on `memory_read_page`. `install-mcp --client kimi-code` therefore writes the MCP URL as `/mcp?flavor=moonshot`; requests carrying that marker get tool schemas with the three root combinators stripped (runtime arg validation and the field docs still enforce every contract, and nested combinators pass through untouched). Every other client keeps the exact upstream schema shape.

`setup-agent`, `uninstall`, and `memory_install_self_routing` (instruction file: `AGENTS.md`) handle kimi-code like the other first-class integrations.

## Files changed

- **Hook scripts**: `hooks/kimi-code/` — 9 `.sh` + 9 `.ps1` (same pattern as gemini-cli, with `agent=kimi-code`); `_lib.sh` header + `scripts/install-hooks.sh` agent list (kimi-code also fetches the two subagent scripts, matching its on-disk bundle)
- **CLI enums**: `cli.rs` — `KimiCode` in `AgentChoice` + `McpClient` (alias `kimi`); `AgentKind::KimiCode` in `ids.rs` + migration `V30__sessions_kimi_code_agent_kind.sql` for the `sessions.agent_kind` CHECK constraint
- **install-mcp**: `install_mcp.rs` — `~/.kimi-code/mcp.json` path (`$KIMI_CODE_HOME`-aware), `RootMcpServers` location, `{url, headers}` entry shape, with the `?flavor=moonshot` URL written via an idempotent helper
- **MCP server**: `server.rs` — `list_tools` override that strips root `anyOf`/`oneOf`/`allOf` from tool parameter schemas only when the request URL carries `?flavor=moonshot`; all other clients receive the upstream schemas byte-identically
- **install-hooks**: `install_hooks.rs` / `render_shared.rs` — TOML-aware idempotent merge of `[[hooks]]` into `config.toml` via `toml_edit` (same machinery as the Codex MCP upsert), `KIMI_CODE_EVENTS` mirroring the 9-event Claude profile; MCP→hook URL inference strips path+query so the flavor marker never leaks into `AI_MEMORY_HOOK_URL`
- **Hook payload**: `payload.rs` — the user-prompt title hint uses `extract_content`, so Kimi Code's content-block `prompt` flattens into the observation title (plain-string agents get identical values back — strict superset)
- **uninstall**: strips only ai-memory `[[hooks]]` entries and the `mcp.json` server entry, preserving third-party hooks and provider/model settings; the kimi-code MCP entry matches both the flavored and the plain URL
- **setup-agent / install-instructions / self-routing**: kimi-code routes to `AGENTS.md`
- **Docs**: README support matrix, `docs/install.md`, `docs/mcp-install.md`, `docs/usage.md`, `docs/windows.md`, CHANGELOG

## Test plan

- `cargo fmt --check` · `git diff --check` ✓
- `TAILWIND_SKIP=1 cargo test --workspace` ✓ — including new coverage: MCP entry shape + `KIMI_CODE_HOME` path tests, TOML merge preserve/idempotency tests, `.sh`/`.ps1` bundle parity, uninstall preserve tests (hooks + MCP-by-endpoint, including flavored-URL matching), routing-instructions test, `AgentKind` round-trip + the `sessions.agent_kind` CHECK test that iterates `AgentKind::ALL`, stateless-HTTP `tools/list` tests asserting the root `anyOf` is stripped with `?flavor=moonshot` and preserved without it, `moonshot_safe_tool_list` unit tests, the MCP→hook URL inference table, and the kimi-code content-blocks user-prompt title regression test
- `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings` ✓


akitaonrails (changes requested):
Post-audit result: the integration direction is sound, but this is not merge-ready yet.

Pros
- Correct KIMI_CODE_HOME-aware TOML/JSON locations and schema shapes.
- TOML-aware hook merge preserves provider/model configuration.
- The Moonshot tools/list flavor is isolated and covered through the HTTP transport.
- The branch compiles against a virtual merge with current main, and the 18 focused Kimi tests pass.

Blocking fixes
1. Default MCP uninstall is a false success. build_plan uses strip_mcp_json_client, so it reports the flavored `?flavor=moonshot` entry as removable, but apply_change dispatches RewriteOp::McpJson through strip_mcp_json with only the unflavored URL. Reproduced end to end: the CLI prints `would remove ai-memory`, then `no-op`, leaves the entry in mcp.json, and still prints `Wiring removed`. Route apply mode through strip_mcp_json_client too, and add an integration test with the exact URL install-mcp writes.
2. Kimi documents PostToolUse as successful calls only and exposes PostToolUseFailure separately. KIMI_CODE_EVENTS registers only PostToolUse, so failed tool outcomes are silently absent even though HookEvent already aliases PostToolUseFailure to PostToolUse. Add the failure event (it can reuse the post-tool-use script/command), update setup-agent to use the Kimi event set, and adjust the 9-event documentation/tests.

The branch also currently conflicts with main in CHANGELOG.md. Recommended next step: update from main, make the two bounded fixes above, rerun the focused Kimi tests plus the full repository gate, then request re-review.

lucasliet:
Thanks for the audit — both blocking fixes are in, and the branch is updated from current `main` (CHANGELOG conflict resolved in merge 3fa927b).

**1. Uninstall false success (2f0fad9)** — `apply_change` now dispatches `RewriteOp::McpJson` through the same `strip_mcp_json_client` candidate set that `build_plan` uses, so plan and apply can no longer diverge. Added `apply_change_removes_kimi_code_flavored_mcp_entry`, which writes a temp `mcp.json` with the exact `?flavor=moonshot` URL `install-mcp` produces and drives `apply_change` end to end (entry removed, unrelated servers preserved).

**2. Missing tool-failure capture (24c04a2)** — `KIMI_CODE_EVENTS` now registers `PostToolUseFailure` (confirmed against Kimi Code's hooks docs: `PostToolUse` fires on success only). It reuses the `post-tool-use` script/command as suggested; the server already aliases `PostToolUseFailure` → `PostToolUse` and reads the outcome from the payload. `setup-agent` now emits the Kimi event set instead of `CLAUDE_CODE_EVENTS`, with per-script dedup so the shared script is listed once. Docs/CHANGELOG/README moved to the 10-event list.

Gate after both fixes: `cargo fmt --check` ✓ · `git diff --check` ✓ · `TAILWIND_SKIP=1 cargo test --workspace` — 44/44 suites ✓ · `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings` ✓

Ready for re-review.

akitaonrails (approved):
Post-audit fixes verified: default Kimi uninstall now removes the flavored MCP URL in apply mode, and PostToolUseFailure is captured through the post-tool path. The final branch also includes focused unit/integration regressions, preserves unrelated config entries, passes the full local gate, and all GitHub checks are green.

PR #209, opened 2026-07-21
Add Atlas Cloud LLM provider

binyangzhu000-sudo:
## Summary
- add an `atlascloud` LLM provider backed by Atlas Cloud's OpenAI-compatible chat-completions endpoint
- wire `ATLASCLOUD_API_KEY`, Atlas provider aliases, default model selection, and `llm-test` CLI mapping
- add focused config/factory coverage for Atlas Cloud auth and provider resolution

## Validation
- `RUSTUP_TOOLCHAIN=stable cargo test -p ai-memory-cli atlascloud_provider_resolves_aliases_default_model_and_api_key`
- `RUSTUP_TOOLCHAIN=stable cargo test -p ai-memory-llm provider_choices_declare_current_auth_requirements`
- `RUSTUP_TOOLCHAIN=stable cargo check -p ai-memory-llm -p ai-memory-cli`
- `RUSTUP_TOOLCHAIN=stable cargo fmt --all -- --check`
- `git diff --check`
- confirmed Atlas Cloud live model catalog includes `qwen/qwen3.5-flash` and `deepseek-ai/deepseek-v4-pro`

No README changes; this is a code-level provider integration only.

### DIFF
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -1068,6 +1068,8 @@ pub enum LlmProviderChoice {
     Copilot,
     /// OpenCode Zen/Go cloud API.
     Opencode,
+    /// Atlas Cloud OpenAI-compatible API.
+    Atlascloud,
 }
 
 /// Arguments for `embed`.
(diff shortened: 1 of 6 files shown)


akitaonrails:
Thanks for proposing Atlas Cloud support and for mapping its endpoint, key, and model. We audited the integration and confirmed Atlas Cloud already fits ai-memory's shipped `openai-compat` transport without a distinct wire protocol or authentication flow. Adding a first-class provider would duplicate the existing client/factory path and create another public alias/default to maintain.\n\nI'm closing this implementation in favor of #216, which documents the exact Atlas Cloud base URL, model id, and `ATLASCLOUD_API_KEY` to `LLM_API_KEY` mapping. That gives users the requested capability without redundant provider code. We can revisit a first-class provider if Atlas later requires behavior that `openai-compat` cannot express.

PR #235, opened 2026-07-23
fix(hooks): deliver the pending handoff on a managed SessionStart

juniorgaudencio00-code:
## The bug

On a managed SessionStart (`ai-memory run`), `fetch_and_accept_handoff` renders
the workstream ledger and returns before ever reaching `latest_open_handoff`.
The handoff is a single-use slot, so nothing else picks it up: it stays `open`,
and the next managed session misses it too, and the one after that.

```rust
if let Some(raw_run_id) = query.managed_run.as_deref() {
    ...
    return Ok(combine_handoff_and_brief(rendered, brief_md));   // ← returns here
}
// ↓ never reached under a managed run:
//   latest_open_handoff(...)   → the pending handoff
```

v1.18.0 fixed half of this by resolving the brief inside the managed branch —
that part works. The handoff is still dropped, and unlike the brief (recomposed
every session) it has no second chance.

Handoffs still reach a managed run through `memory_handoff_begin`, so this is
not a dead code path: it is exactly how a curated "resume here" note is meant to
travel between sessions.

## Reproduction

1. End a session with `memory_handoff_begin`.
2. Start the next session with `ai-memory run`.
3. The ledger arrives; the handoff does not. `handoffs.state` is still `open`.

Observed in production on a self-hosted 1.17.3 server: three handoffs in a row
sat undelivered across four sessions, silently — the agent could not tell
"nothing was left for me" from "something was left and got swallowed".

## The fix

Move the managed branch into `fetch_managed_context()`, which returns `None`
instead of exiting the request — including on the anomalous paths (bad run id,
no active run, agent mismatch, context already delivered), where the pending
handoff still needs to get through. The request then continues into the normal
handoff + brief path and the three blocks are concatenated.

Ordering is handoff → ledger → brief. The handoff goes first because it is a
short curated pointer that should not be buried under a ledger that can run tens
of KB; the existing ledger-before-brief order is untouched, so
`managed_handoff_combines_portable_delta_and_project_brief` still holds.

`resolve_requested_session_brief` is removed: with the managed branch falling
through, the brief is resolved once by the existing call instead of twice.

## Tests

New regression test `managed_session_start_delivers_ledger_and_pending_handoff`
sets up a workstream with a delivered-once ledger plus a pending handoff, and
asserts both reach the response and that the handoff is marked accepted. It
fails on `main` (the handoff marker is absent).

`cargo fmt --check`, `cargo clippy --all-targets -D warnings` and
`cargo test -p ai-memory-hooks` all pass.


akitaonrails:
Good root-cause fix, @juniorgaudencio00-code — audited locally (full workspace suite green incl. the new regression test; auto-merges cleanly onto the new main with #236 in).

**What's right:** the early-return in `fetch_managed_context` really was dropping the single-use handoff slot, and your fix delivers handoff + ledger + brief in a justified order. Non-managed path is byte-for-byte equivalent. `accept_handoff`'s conditional `UPDATE … WHERE state='open'` through the single writer actor means racing sessions can't double-accept. Real regression test at `router.rs:5653` asserting both markers delivered and the slot consumed.

**Two asks before merge:**

1. **[Blocking] CHANGELOG entry** under `## [Unreleased]` → `### Fixed` — observable bug fix, required by the merge gate. #236 just landed with its own `[Unreleased]` entry, so **rebase onto main when you add it** to avoid the head-of-section race (your code changes themselves merge clean).
2. **[Should-fix] Side-effect ordering in `fetch_managed_context` (`router.rs:1190`)** — `accept_managed_run_context` now commits *before* the fallible `resolve_project_ids_inner`, `latest_open_handoff`, and `render_requested_session_brief` run. If any of those errors, the response 500s but the ledger is already flagged `context_delivered` — the continuity packet is silently lost on retry, which is the same class of bug this PR fixes. Please have `fetch_managed_context` return `(run_id, rendered)` and accept just before the final `combine`, after assembly succeeds.

Nit: consider a second `fetch_and_accept_handoff` in the test asserting non-redelivery explicitly — the current state assertion is nearly equivalent, so not blocking.

With those, this merges.

PR #236, opened 2026-07-24
refactor: code-audit remediation — coverage, dedup, crate boundaries

akitaonrails:
## Summary

Full code-audit remediation pass over the workspace (~92k LOC, 10 crates). The audit itself found the codebase healthy — all 15 documented architectural invariants verified compliant, zero TODO/FIXME markers, no panics reachable from external input. This PR fixes the drift it did find, in priority order.

### Test coverage (+52 tests)
- **store**: first *direct* writer-actor tests — FIFO ordering, SQL-error isolation (a failed command doesn't kill the actor), mid-batch rollback, clean shutdown, `WriterClosed` after shutdown. Invariants #2/#3 rest on this file.
- **store**: managed-workstream ledger lifecycle tests (heartbeat leases, idempotent finish, foreign-event batch rollback, `accept_context`, event ordering/limits, stale-lease expiry) — previously the largest untested SQL surface, covered only by a manual script.
- **store**: the AGENTS.md-mandated table-driven scope-resolution suite (partial scope / missing explicit scope / active-project precedence / cross-workspace isolation) + failed-migration rollback/recovery test.
- **consolidate**: curator unit tests (finding classification, pinned-page exclusion, `age_days` math, markdown rendering, byte-level read-only guarantee) — previously zero unit tests.
- **llm**: `auth_file` round-trip/corruption/0600-permission tests, opencode basics.

### Refactors (behavior-preserving)
- **Embedding backfill deduplicated**: one shared implementation in `ai-memory-consolidate::embed`; the scheduler tick and `/admin/embed` both call it. The scheduler now skips empty pages instead of embedding empty strings.
- **OAuth token types unified**: `OidcToken`/`OpenAiOAuthToken` are now specializations of `StoredOAuthToken<E>` with one shared `refresh_grant`. The `auth.json` on-disk format is unchanged (pinned by tests).
- **Atomic writes consolidated**: CLI `apply_shared` delegates to ai-memory-wiki's canonical `write_atomic`, gaining parent-dir fsync durability. The llm credential-file writer stays (0600 requirement, no heavy dep edge) with a doc comment tying it to the canonical one.
- **Server machinery moved out of the CLI crate** (thin-CLI rule): axum auth middleware → `ai-memory-mcp`, web-UI mounting/HTML injection → `ai-memory-web`, auto-improve scheduler orchestration → `ai-memory-consolidate`. `serve.rs` shrank ~3,800 → ~1,860 lines.

### Robustness
[...]

PR #237, opened 2026-07-24
feat(workstream): add Grok Build CLI managed harness

iagogfe:
## What changed

`ai-memory run grok` (alias `grok-build`) now manages Grok Build CLI as a full
cross-harness workstream member. Before: Grok had MCP + lifecycle-hook capture
only, with no managed resume and no context injection. After:

- Fresh sessions launch with a wrapper-generated `--session-id`; returning
  launches resume the linked session with `--resume <id>`. Explicit native
  selectors (`--resume`/`-r` including the bare picker, `--continue`/`-c`,
  `--session-id`/`-s`, `--fork-session`) always win, and utility subcommands
  (`agent`, `sessions`, `login`, `export`, `wrap`, ...) pass through untouched.
- The bounded workstream context packet is delivered through Grok's native
  `--rules` flag (documented alias of `--append-system-prompt`), since Grok
  ignores `SessionStart` stdout and its `UserPromptSubmit` hook is passive.
  Delivery is acknowledged only after the child spawns, matching the Crush
  contract. `--rules` is single-use in Grok's parser, so a natively supplied
  `--rules`/`--append-system-prompt` wins and the packet is redelivered on a
  later run that can accept it (stderr notes this).
- Transcript import reads `$GROK_HOME/sessions/*/*/chat_history.jsonl`
  read-only. Grok can rewrite the journal in place on rewind, so the source
  cursor stores a prefix hash and record ids are content hashes (the Kimi
  strategy, now shared via `journal_rewrites_in_place`): a rewrite replays
  from the start without duplicating imported events. `system` records,
  encrypted `reasoning` records, and the injected `<user_info>` block are
  excluded and surfaced as loss annotations. Sibling files (`events.jsonl`,
  `updates.jsonl`, `rewind_points.jsonl`) are never read as transcripts.
- Checkout discovery reads `summary.json`'s recorded `info.id`/`info.cwd` and
  never parses the URL-encoded bucket directory name. `GROK_HOME` is honored.
- Wrapper `--yolo` maps onto Grok's native `--yolo` and is not duplicated when
  either native spelling (`--yolo`, `--always-approve`) is already present.
- Grok stays out of the bare-mode automatic pool (explicit support first, per
  the managed-harness contribution protocol); the server accepts it as an
  explicit harness and rejects it in `available_agents`.

No defaults changed for existing harnesses; `AgentKind::Grok` already existed,
so no schema migration was needed.

## Why

Grok Build CLI satisfies the full managed-adapter contract (caller-supplied
session ids, documented resume flags, checkout-keyed session store with
recorded metadata, and a native pre-turn context mechanism), so managed
workstreams can extend to it without any partial-adapter compromises.

## Test plan

- [x] Manual test: deterministic acceptance
      (`AI_MEMORY_ACCEPTANCE_DETERMINISTIC_ONLY=1
      scripts/managed-workstream-acceptance.sh`) passes with the new Grok
      fake-mode fixture: fresh launch receives `--session-id` and no `--rules`
      on an empty workstream; the same-session return resumes via `--resume`
      with no packet; a fresh Grok session joining the established Kimi
      fixture workstream receives `--rules` carrying the prior harness
      sentinel; incremental import neither duplicates nor misses a round.
- [x] Manual test: real-harness pass with
      `AI_MEMORY_ACCEPTANCE_HARNESSES="claude grok"` against Grok Build CLI
      v0.2.111 (Linux/WSL2). Real Grok received the Claude round sentinel
      through `--rules` and replied with both tokens; its reply was imported
      from `chat_history.jsonl` (7 events).

New focused coverage: fresh/resume/selector-precedence/utility/yolo/env tests
in `harness.rs`; discovery-via-summary.json, private-record exclusion, and
rewrite-cursor-reset tests in `transcript.rs`; explicit-accepted /
automatic-rejected and direct-context-gate tests in
`ai-memory-hooks/src/workstream.rs`; `--rules` suppression detection in
`run.rs`.

## Commit attribution

## CHANGELOG (merge gate)

## Notes for reviewers

- Delivery lands in Grok's system prompt (`<human_rules>` block) rather than a
  user message; the rendered packet already labels historical tool activity as
  completed evidence, and the real-harness pass confirmed the model reads it.
- During the real-harness pass, the final return-to-Claude leg of the
  round-trip assertion failed for a reason outside this diff: the ~14KB
  context packet was surfaced to Claude Code as a truncated 2KB hook preview,
  and the acceptance prompt forbids tools, so the model could not quote the
  sentinel. The Claude injection path is untouched here; flagging it as a
  pre-existing large-packet limitation of the acceptance script.
- `run_context`/`accept` server gate widened from Crush-only to Crush|Grok;
  everything else rides the existing shared lease/cursor/import path.
- Known v1 limitation: `chat_history.jsonl` records carry no per-record
  timestamps, so imported Grok events have no `occurred_at` (ordering is
  preserved by file order, and candidate recency uses file mtimes).


akitaonrails:
Solid work, @iagogfe — audited locally: full workspace suite green on your head (1771 passed, incl. the 13 new Grok tests covering the hard cases: in-place journal rewind replay with stable ids, discovery contradiction, bare `--resume` picker, single-use `--rules`, and the Kimi→Grok acceptance fixture). fmt/clippy clean.

**What I liked:** faithful mirror of the merged Kimi pattern, and you *refactored* the Kimi-only special cases into a shared `journal_rewrites_in_place()` instead of copy-pasting — net dedup. Explicit-only pool membership (client `AUTO_HARNESSES` + server `AUTO_AGENTS` both excluding Grok) matches the contribution protocol.

**Three asks before merge:**

1. **[Blocking] Rebase onto main** — the PR is currently DIRTY. Good news: the only conflict is `CHANGELOG.md` (head-of-`[Unreleased]` append race with #236's entry, which merged today); keep both entries. I verified #236 touched zero of your 9 files, so the rebase should be purely textual.
2. **[Should-fix] Two doc-list gaps** — `docs/design-decisions.md` §15 (managed-harness list) and the `run` subcommand row in `docs/install.md` (~line 1174) still don't list Grok. This is exactly the omission class we had to fix post-merge for Kimi in d72f31b — let's not repeat it.
3. **[Confirm] Native-flag contract** — I can't verify `--session-id`/`--resume` behavior locally (no Grok CLI on this machine). Can you confirm in a reply that the acceptance run (`AI_MEMORY_ACCEPTANCE_HARNESSES="claude grok"` or equivalent) passed against a real Grok install, and which version? Your v0.2.111 claim in the PR body sounds like yes — just want it on the record.

Nit: `run.rs:153` — `user_supplied_grok_rules(&native_args)` is computed unconditionally for every harness; harmless, but it could live inside the Grok branch.

Rebase + the two doc lines and this is a merge.

iagogfe:
**Items 1 and 2, done.** Rebased onto main, on top of 6c1cb77. The only conflict was the CHANGELOG `[Unreleased]` head, and both entries are in, yours first. Both doc lists now name Grok Build CLI: `docs/design-decisions.md` §15 and the `run` row in `docs/install.md`. Head is 4032802.

Before the good part, a correction. The PR description called the failing acceptance leg a pre-existing limitation. That was wrong, and I only found out because you asked me to put the run on the record. Chasing it turned up a real bug in my adapter.

**The bug.** Grok keeps harness scaffolding inside `user` records: an environment block, then Claude-style `<system-reminder>` blocks carrying project instructions, the skills catalogue, and the connected MCP servers. I only skipped the environment block, so the rest went into the portable ledger as user input. One real session: 42KB of reminders against 270 bytes of actual input.

That also explains the flaky leg. `render_managed_context` caps a packet at 30,000 chars and one event at 6,000, so roughly 12.8KB of the budget went to reminders and pushed real conversation out. The next harness got a packet full of my scaffolding.

Measured on one captured session, same file through the old and new parser: 5 events / 42,984 bytes before, 2 events / 327 bytes after. What survives is the real `<user_query>` (270) and the assistant reply (57). Live `claude grok` round trip, end to end: 43,430 bytes of Grok events before, 1,017 after, no ledger event containing a reminder, and the round trip passes.

Kimi doesn't need this filter because it keeps the same scaffolding in `config.update`, which the adapter never reads. Grok has no equivalent marker, so prefix matching is the signal available. Fix plus a regression test built from the four real blocks is in d4585a5.

**Your item 3, on the record.** Grok Build CLI v0.2.111, Linux (WSL2), real model calls. `AI_MEMORY_ACCEPTANCE_HARNESSES="claude grok"` passes end to end after the fix: fresh session created through the wrapper's `--session-id`, packet delivered via `--rules`, Grok replied with both round tokens, its reply imported from `chat_history.jsonl`, and the returning launch resumed the same session through `--resume <id>`. I also ran `claude kimi` (kimi v0.29.1) as a control, which passes with a clean ledger, and `claude codex` on pristine main, which is how I established the failure wasn't pre-existing.

**The nit.** It doesn't compile inside the Grok branch. `native_args` is moved at run.rs:214 by the adoption path:

```
error[E0382]: borrow of moved value: `native_args`
   --> crates/ai-memory-cli/src/commands/run.rs:260:37
214 |                             native_args,
    |                             ----------- value moved here
```

Line 153 is what keeps that call site from needing another clone, so I left it. Say the word if you'd rather have the locality.

I also fixed `docs/managed-harness-contributions.md`, which still listed Grok among the integrations that don't become managed (4032802).

Two things I hit along the way that are in your code rather than mine, filed with the evidence: #241 and #242.

PR #238, opened 2026-07-24
prompts: emit wikilinks + mirror input language in consolidation

juniorgaudencio00-code:
## Motivation

Two gaps in the consolidation prompts, both observed running a self-hosted
server in production:

1. **No page is ever linked.** The wiki is a graph and the machinery for
   Obsidian-style wikilinks already resolves at three levels
   (same-project, cross-project, `_global`) — but no prompt instructs the
   model to *emit* links. In practice the corpus grows as disconnected
   islands: every page is an orphan, and `/api/v1/graph` stays empty of edges.

2. **Output language drifts to English.** Working in Portuguese, the
   consolidator would translate the user's vocabulary and emit English page
   titles, even though the input was entirely Portuguese. Code, identifiers
   and paths should stay verbatim, but prose and titles should mirror the
   input language.

## The change

Adds two short sections to each of the three consolidation system prompts
(`single_`, `batch_`, `bootstrap_consolidate_system.md`):

- **WIKILINKS** — connect related pages into the graph. Documents the three
  link forms and their exact semantics: `[[page-path]]` (same project, path
  not title), `[[project:page-path]]` (sibling project, only when named in
  the input — never invent project names), and `[[_global:page-path]]`
  (cross-cutting principle/convention/trap). Notes that a link to a
  not-yet-existing page is fine (pending link, resolves on appearance), and
  steers toward 2–5 well-chosen links rather than exhaustive linking.

- **OUTPUT LANGUAGE** — write all page titles and body prose in the dominant
  natural language of the input; keep code, identifiers, file paths, shell
  commands and error strings verbatim; JSON keys stay in English.

Prompt-only change — no code, no schema, no test changes. Split into two
commits (wikilinks + language, then the concrete cross-project example and the
"ALL titles" clarification) for reviewability.


akitaonrails:
Nice catch on both gaps, @juniorgaudencio00-code — audited locally (fmt/clippy/tests green; your branch is already on top of the merged #236, so no rebase needed for code).

**What I verified:**
- The wikilink syntax you teach matches the resolver exactly: bare `[[path]]` root-relative, `[[project:path]]` via `split_scope`, `_global` scope, and unresolved forward links do persist. Teaching the wrong syntax here would have been a silent graph-killer — you got it right.
- New sections sit *before* the JSON-shape instructions, and "JSON keys stay in English" preserves structured-output compliance.
- Language mirroring is safe downstream: title consumers use Unicode-safe folding, and slot pages are matched by fixed English *path*, never by title.

**Before merge, two asks:**
1. **[Blocking] CHANGELOG entry** under `## [Unreleased]` → `### Changed` — this alters generated wiki content (link density, page language) for every consolidation user, so the merge gate requires it. Since #236 just added its own `[Unreleased]` entry, **rebase onto main when you add it** to avoid the head-of-section race.
2. **[Should-fix] Prompt-marker assertion** — `consolidator.rs` already has a prompt-content contains-test (~line 862); extend it (and the bootstrap equivalent) to assert the `WIKILINKS`/`OUTPUT LANGUAGE` sections survive future prompt edits.

Nit: the OUTPUT LANGUAGE paragraph wraps mid-phrase ("...dominant natural language\nof the input") in all three prompt files — tidy while you're in there.

With those in, this merges.

PR #239, opened 2026-07-24
perf(mcp): throttle M8 access-counter bumps to once per page per minute

samirhvbr:
Picks up one of the follow-ups you listed as out-of-scope in #236:

> Unthrottled `spawn_access_bump` per search call (`mcp/src/server.rs`).

## Problem
`memory_query` calls `spawn_access_bump` on all three search paths, so every search spawns a writer-actor `bump_access` command for every hit. Repeated or overlapping queries return the same hot pages, so a burst of searches floods the single writer actor with redundant M8 reinforcement writes that compete with real work — one `tokio::spawn` + one writer command per search, unbounded.

## Change
A shared cooldown clock on the server (`access_bump_seen: Arc<Mutex<HashMap<PageId, Instant>>>`):

- A page is bumped **at most once per `ACCESS_BUMP_COOLDOWN` (60s)**.
- Aged-out entries are pruned in the same pass, so the map stays bounded by the recent working set rather than growing with every distinct page ever searched.
- When nothing is due, no task is spawned at all.
- The policy is factored into a pure `select_bumpable(seen, ids, now, cooldown)` helper, unit-tested deterministically (injected `now`) without a store or runtime.

No new dependencies — std `Mutex` / `Instant` / `HashMap` only.

## Semantic note (your call)
This changes M8 reinforcement from **once per search** to **at most once per page per minute**. For a signal feeding day/week-scale retention scoring that's ample resolution and the effect on decay ordering is negligible, while the writer-pressure reduction under load is large. If you'd rather preserve the exact per-search increment count, the alternative is to coalesce over a window and *sum* increments (`access_count = access_count + N`) instead of dropping them — happy to switch to that shape if you prefer.

## Gate (`TAILWIND_SKIP=1`)
- `cargo clippy --workspace --all-targets -- -D warnings` ✓
- `cargo test --workspace` ✓ — 0 failures; adds `access_bump_throttle_dedups_within_cooldown_and_reallows_after`

Branched off `main` (v1.18.0), independent of #236.


akitaonrails:
Thanks for picking up the `spawn_access_bump` follow-up from #236, @samirhvbr — audited locally and this is clean work.

**Audit results (fmt/clippy/full workspace suite green, merges cleanly onto the new main):**

Pros:
- Fixes exactly the right thing: one spawn + writer command per search call on all three search paths, now suppressed unless a page is due.
- Throttle state is an instance field shared across clones, not a global singleton (invariant-compliant); the map self-prunes in the same pass, so it's bounded by distinct pages searched per window.
- `select_bumpable` is a pure function with injected `now` — the unit test covers dedup, per-page independence, window expiry, and pruning. First sighting bumps immediately, so recency signal isn't deferred; `Entry::Vacant`-only insertion means a continuously-hot page still gets its once-per-window bump (no starvation).

One request before merge — **please add a CHANGELOG entry** under `## [Unreleased]` → `### Changed`. The throttle changes how `access_count`/`last_accessed_at` accumulate, which feeds the documented M8 decay math, so it's user-visible per our merge-gate rule. When you do, **rebase onto main first** — #236 just landed with its own `[Unreleased]` entry, and rebasing avoids the head-of-section race.

Optional, your call: a one-liner in `docs/ARCHITECTURE.md` (~line 101) noting the once-per-page-per-minute throttle. Also noted your sum-coalescing alternative in the PR body — I'm fine with the drop-based version as-is; the counters are a heuristic.

Once the entry's in, this is a merge.

samirhvbr:
Thanks for the audit — done, both of them.

- **Rebased onto current `main`** (`6c1cb77`, post-#236) first, as suggested, so the `[Unreleased]` entry lands under your `/admin/open-sessions` one instead of racing it at the head of the section.
- **CHANGELOG** entry added under `## [Unreleased]` → `### Changed`. I spelled out the user-visible part explicitly, since it feeds the M8 decay math: first sighting still bumps immediately and a continuously-hot page keeps earning one bump per window, so recency ordering is unaffected — only the *magnitude* of `access_count` grows more slowly under bursty search.
- **`docs/ARCHITECTURE.md`** — took the optional one-liner too, appended to the M8 reinforcement sentence at §6.

Agreed on keeping the drop-based version; the sum-coalescing shape would buy exact counts at the cost of carrying pending deltas across the window, which isn't worth it for a heuristic.

Re-ran the full gate on the rebased tree (`TAILWIND_SKIP=1`): `fmt --check` ✓ · `git diff --check` ✓ · `clippy --workspace --all-targets -- -D warnings` ✓ (0 warnings) · `cargo test --workspace` ✓ — 1809 passed, 0 failed. Squashed into the single commit so the PR stays self-contained.

PR #245, opened 2026-07-25
fix(hooks): simplify normalize_segments with ? operator

lhzapata:
## What changed

`normalize_segments` now uses `?` for the absolute-path branch instead of an explicit `else { return None; }`. Behaviour is unchanged; the code remains equivalent and becomes clean under newer Clippy versions that flag this pattern.

## Why

This is a preventive Clippy cleanup observed while validating with a newer secondary toolchain. The project still pins Rust 1.95, so this is not fixing a current pinned-toolchain failure.

## Test plan

- [x] Manual test: validated the change against Rust 1.95.0 and a newer Rust/Clippy environment where the lint is emitted.

## Commit attribution

## CHANGELOG (merge gate)

- [x] No `CHANGELOG.md` entry: this is an internal Clippy/style cleanup with no user-facing behaviour, flag, env var, endpoint, MCP tool, marker key, default, or observable bug fix.
- [x] No default changed.

## Notes for reviewers

Trivial mechanical cleanup only; the `strip_prefix('/')?` branch is equivalent to the previous explicit `return None`.

### DIFF
--- a/crates/ai-memory-hooks/src/capture_policy.rs
+++ b/crates/ai-memory-hooks/src/capture_policy.rs
@@ -777,10 +777,9 @@ fn normalize_segments(path: &str) -> Option<String> {
             format!("{}:/", path[..1].to_ascii_uppercase()),
             path[3..].split('/').collect(),
         )
-    } else if let Some(rest) = path.strip_prefix('/') {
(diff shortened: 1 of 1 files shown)


lhzapata:
Closing

PR #259, opened 2026-07-26
fix(cli): honor .ai-memory.toml scope in every client command

milesibastos:
## What

Only the lifecycle hooks read `.ai-memory.toml`. Every client command resolved `--workspace` to the literal `default`, so a checkout declaring `workspace = "acme"` had its captures land in `acme` while `run`, `bootstrap`, `search`, `write-page` and the rest resolved into `default` — one repository split across two scopes.

`ai-memory run` is the worst case: its managed workstream landed in `default/<repo>` while the hook-captured sessions went to `acme/<repo>`, leaving `default/<repo>` rows that look abandoned (`0 pages, 0 sessions, 0 observations`).

Purging one of those is how this was found. `workstreams` cascades from `projects` and `managed_runs` cascades from `workstreams`, so the purge deleted a **live** run's lease row: the wrapper then failed every heartbeat with `409 managed run lease is not active` for the rest of the session, and that workstream's portable ledger was lost — segments stranded under `raw/workstreams/<id>/` with no row pointing at them. The purge report counted pages, sessions, observations, handoffs and embeddings — none of which covered what it actually deleted.

## How

**Scope resolution.** The marker reader moves out of `commands::hook_capture` into a new `marker` module, shared by both entry points. `commands::resolve_scope` resolves each field independently: explicit flag → marker (`workspace`, and `project` or the main repo root under `project_strategy = "repo-root"`) → the previous fallbacks. The defaulted `--workspace` args become `Option<String>` so an explicit `--workspace default` is distinguishable from the default value. When the marker decides a field the command announces the resolved scope on stderr, and `AI_MEMORY_IGNORE_MARKER=1` restores the previous resolution for one invocation.

`serve` keeps its literal default on purpose: it has no caller cwd, and its `--workspace`/`--project` are the baked fallback for hook events without a usable one.

**Purge guard.** `purge-project` now refuses with a 409 naming the offending workstreams while a managed run holds a lease that has not expired, unless `--force` is passed. The report gained `workstreams_deleted`, `managed_runs_deleted` and the `raw/workstreams/<id>/` directories left orphaned on disk. `move-project`'s internal source purge maps the same conflict to a 409 naming how many pages were already copied, rather than a 500 on a half-applied merge.

## Commits

Two, kept separate on purpose:

- `3d8b9b4` — the fix.
- `3fe60ec` — what an adversarial review found in it. Most notably the guard originally treated `state = 'active'` as liveness without checking `lease_expires_at`, so an agent killed by SIGKILL/sleep/OOM would have blocked every future purge of its project forever; and `AI_MEMORY_IGNORE_MARKER` / `AI_MEMORY_PROJECT_STRATEGY` were read with `std::env::var` outside `Config::load()`, violating the config-read invariant.

## Tests

- `marker::tests` — parser and walk contract, moved with the code.
- `commands::tests` — the precedence table, per field.
- `crates/ai-memory-cli/tests/marker_scope.rs` — subprocess coverage for the env wiring and the `$HOME` bound.
- `ops.rs` — purge refuses with a live run, ignores an expired lease, and reports what the cascade removed.

## Gates

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test --workspace` all pass. `tests/hooks/test_lib.sh` is 29/29.

## Note

The two `CHANGELOG.md` entries still need this PR's `(#NNN)` reference — added in a follow-up commit once the number exists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   resolved into `default` — the same repository split across two scopes, with
   `ai-memory run`'s managed workstream stranded on the wrong side. Each field
   still prefers an explicit flag; when the marker decides one, the command
-  announces the resolved scope on stderr. `AI_MEMORY_IGNORE_MARKER=1` restores
-  the previous resolution for one invocation. `ai-memory serve` is unchanged:
+  announces the resolved scope on stderr, naming which half the marker
+  decided. `AI_MEMORY_IGNORE_MARKER=1` restores the previous resolution for
+  one invocation (client commands only — the hooks keep reading the marker).
+  `embed --force` without `--project` still fans out across the workspace and
+  no longer needs a derivable project name. `ai-memory serve` is unchanged:
   it has no caller cwd, and its `--workspace` / `--project` remain the baked
   fallback for hook events without a usable one.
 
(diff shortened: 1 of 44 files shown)


akitaonrails:
Maintainer audit follow-up completed in f69e896:

- aligned workspace-only CLI scope with hook basename semantics, including subdirectories/worktrees
- bounded marker discovery at HOME or checkout root across native, shell, PowerShell, and generated TypeScript integrations
- preserved marker discovery through Docker wrappers with a bounded read-only scope mount
- removed raw managed-workstream segment directories during purge and reported partial failures
- documented and tested that move-project --force never deletes a live managed run during copy-purge

Local verification passed: cargo fmt, git diff --check, workspace tests, workspace clippy -D warnings, cargo deny, companion fmt/test/clippy, hook shell tests, native packaging checks, and handoff smoke.

PR #291, opened 2026-07-28
feat: add pluggable content storage with outl adapter

### LINKED ISSUE #290: proposal: pluggable content storage, let the user choose where memory lives
**Problem this would solve**

Today page content is hardwired to the `wiki/` markdown + SQLite pipeline. Works great as a default, but memory is personal data. People already live in Obsidian, Logseq, org-mode, outliners. There is no way to make ai-memory persist pages inside the tool the user already uses, so memory ends up being one more silo to maintain instead of part of their knowledge base.

**Proposed solution**

Content storage adapters. I implemented this on my fork and it's running in production on my machine, with [outl](https://github.com/avelino/outl) (my outliner) as the first backend. The shape:

- a `ContentBackend` trait in `ai-memory-store` owning only the content write path: persist/delete/move/read page + `fs_root()`. Search, recency, graph, embeddings and decay keep reading the SQLite index in every mode. SQLite stays the derived index, exactly as today.
- selection via config, no new concepts for the operator:

```toml
[storage]
backend = "outl"   # "sqlite" is the built-in default

[storage.outl]
workspace_dir = "~/ai-memory-outl"
```

- adapters are feature-gated modules under `ai-memory-store/src/adapters/<name>/`. Each cargo feature (`adapter-<name>`) carries its extra deps, so the default build compiles zero of them.
- `[storage.<name>]` sections are opaque to the config layer (flatten map). A new adapter is a module + a feature + one line in the registry. Config code never changes.
- `AdapterContext` hands the factory the index writer/reader, the built-in fs backend (for shadow/mirror modes) and its raw TOML section.
- a backend with no on-disk root returns `fs_root() = None` and the fs watcher stands down.
- operational data (sessions, observations, users, handoffs) stays SQLite-only. Out of scope on purpose.

Some decisions from the outl adapter that generalize to any backend:

- **index-preserving**: the index keeps the ORIGINAL page body, the backend stores its own projection. Reads serve the index, so search results never drift with the target app's markdown dialect.
- **external edits**: pages carry a content-sha stamp and a reconciler re-indexes only what diverged (user edited the page in the other app). The same pattern covers an Obsidian vault with file hashes.
- **shadow mode**: fs stays primary, the adapter mirrors best-effort. Good migration path before flipping a backend to primary.
- each backend declares its consistency model: fs is atomic with rollback, append-only backends are SoT-first with the reconciler healing the index.

An Obsidian adapter on top of this is mostly "write `.md` into the vault + reconcile on edit". The trait already covers it.

**Alternatives considered**

- **Talking to each app over HTTP/API**: rejected for the first backend. Outl has no server, and pulling a network hop into every page write adds latency and failure modes for something that can be embedded. Nothing in the trait blocks an HTTP-based adapter later, it's just not required.
- **Hardcoding a second backend behind an if/else**: doesn't scale past two. The registry + feature-gate shape costs almost nothing more and makes third-party backends possible.
- **One-way export/sync into the target app**: memory would still live in the silo and exports drift the moment the user edits a page on the other side. Making the backend the source of truth with a reconciler solves the edit-conflict problem instead of ignoring it.

**Additional context**

Full implementation is on my fork (trait + registry + outl adapter as reference, integration tests + a perf harness included). Happy to send the PR: trait + registry + the outl adapter as reference implementation, or trait + registry only if you prefer keeping adapters out of tree. Makes sense to bring this upstream?

avelino:
## What changed

Before: page content always lands in the `wiki/` markdown tree + SQLite index, no way to change where memory lives.

After: `[storage] backend = "<name>"` in `config.toml` selects a content storage adapter. `sqlite` stays the built-in default and **nothing changes for existing installs** (no config = same behaviour, byte for byte). The PR ships:

- a `ContentBackend` trait in `ai-memory-store` owning the content write path (persist/delete/move/read page + `fs_root()`); search, recency, graph, embeddings and decay keep reading the SQLite index in every mode
- the historical fs pipeline extracted verbatim into `FsContentBackend` (same atomic write + rollback semantics)
- an adapter registry where each adapter is a feature-gated module (`ai-memory-store/src/adapters/<name>/`, cargo feature `adapter-<name>`); default build compiles zero extra deps
- the first adapter: `outl` (feature `adapter-outl`), persisting pages into an embedded [Outl](https://github.com/avelino/outl) workspace (crates.io `0.8`) as content source of truth, with a sha-stamped reconciler absorbing edits made inside the Outl app back into the index
- shadow mode (`mode = "shadow"`): fs stays primary, the adapter mirrors best-effort, migration path before flipping
- fs watcher stands down when the backend has no on-disk root (`fs_root() = None`)
- `docs/storage-adapters.md` with the contract and a how-to for new adapters

Also bumps rmcp `1.7` to `1.8`: `peer_info()` changed signature upstream and any fresh `cargo install` (which ignores the lockfile) was failing to compile.

Changed defaults: none. `[storage]` absent = `backend = "sqlite"` = today's behaviour.

fixed: https://github.com/akitaonrails/ai-memory/issues/290

## Why

Memory is personal data. People already keep their knowledge base in Obsidian, Logseq, org-mode, outliners. Hardwiring content to the internal wiki makes ai-memory one more silo instead of part of what the user already maintains. The adapter shape opens that up: an Obsidian adapter on top of this trait is mostly "write `.md` into the vault + reconcile on edit".

Proposed and discussed in #ISSUE_NUMBER.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes (ran with the stable toolchain; the pinned 1.95 on my machine has no clippy component)
- [x] `cargo test --workspace` passes, including 206 unit + 5 integration tests for the store with `adapter-outl` on (integration covers persist round-trip, external-edit reconcile convergence, idempotent delete, project-scoped removal and concurrent second opener falling back to an ephemeral actor)
- [x] Manual test: ran `serve` with `backend = "outl"` against my real data dir (498 pages). Boot logs `outl content backend attached`, watcher stands down. Wrote a page via `write-page`: it lands in the Outl workspace under the deterministic slug with the content projection and the FTS index serves the original body. Edited the page via the `outl` CLI (separate actor): reconciler absorbed it into the index in under 5s and converged. Perf harness included (`tests/outl_bench.rs`, ignored by default): 10.8ms per page write, reconcile scan of 300 pages in 16ms.

## Commit attribution

## CHANGELOG (merge gate)

- [x] I added a `CHANGELOG.md` `[Unreleased]` entry (Added: storage adapters + outl backend; Fixed: rmcp 1.8)
- [x] Any changed **default** is called out explicitly in "What changed" above (there are none)

## Notes for reviewers

- **Consistency contract changed per backend, on purpose**: the fs backend keeps atomic write + rollback; the outl backend is append-only, so a failed index write leaves the SoT ahead until the reconciler heals it. Documented in the trait and in `docs/storage-adapters.md`. This is the design decision I'd like the most scrutiny on.
- The index always keeps the ORIGINAL page body (the Outl side stores a projection), so search results never drift with the target dialect. That's why `read_page` serves the index in outl mode.
- Declared v1 limitations for non-fs backends: git checkpoints/restore-page stand down, `backup` doesn't include the external workspace, externally edited pages lose graph links until their next engine write.
- `[storage.<name>]` sections are opaque to the config layer (flatten map), so config code never changes for a new adapter. Unknown `backend` names load fine and fail at boot listing the registered ones (the config layer can't know which features were compiled in).


akitaonrails:
Maintainer audit result: I closed this rather than merging it.

The proposal has real value: keeping ai-memory pages visible in an existing knowledge tool, deterministic scope/path mapping, a shadow mode, and explicit adapter tests are all useful ideas. The current implementation cannot be repaired as a small follow-up, though, because it changes load-bearing storage and recovery guarantees:

- This repository's documented invariant is that Markdown-in-git is the source of truth and SQLite is derived. Primary adapter mode replaces that invariant and intentionally disables complete git checkpoint/restore and backup coverage.
- Some existing mutations bypass the selected backend. `restore_page_from_checkpoint` writes the filesystem plus SQLite directly, and auto-improvement approval uses the filesystem rollback path plus a specialized SQLite transaction. In primary Outl mode those operations split the advertised source of truth from the served index.
- Project/workspace purge commits the SQLite deletion before calling the adapter removal. An adapter failure therefore leaves external canonical content behind after the database identity and audit state have already been destroyed; there is no rollback or durable retry protocol.
- Switching an existing installation to primary mode has no lossless migration/verification contract. Reads continue to be served from SQLite, and the Outl projection is intentionally lossy for original Markdown/frontmatter/link state.
- The adapter is feature-gated in `ai-memory-store`, but `ai-memory-cli` enables that feature unconditionally, so every shipped CLI compiles and carries the full Outl dependency graph.
- The required dependency-policy gate fails on this exact head: `Unicode-DFS-2016` is not allowed by the current license policy, and the Outl graph adds the unmaintained `bincode 1.3.3` (RUSTSEC-2025-0141) and `smallstr 0.3.1` (RUSTSEC-2026-0215).
- The branch includes the unrelated managed-run heartbeat behavior change, and it is now conflicting with main.

Validation performed on `acd7b45e71dae0fd0368176046eab51aab73e1ec`:
- `cargo test -p ai-memory-store --features adapter-outl --test outl_backend`: 7 passed.
- matching clippy with `-D warnings`: passed.
- `cargo deny check`: failed for the license/advisory findings above.

A safer future design is a separate optional companion sync/exporter: Markdown-in-git remains canonical; changes from an external tool return through the existing Wiki mutation/admission boundary with explicit revision/conflict detection; backup, restore, reindex, checkpoints, and purges remain complete; and tool-specific dependencies do not enter the core binary. That design should start with a written consistency/migration/recovery contract and failure-injection tests before implementation.

Thank you for the substantial implementation and for responding to the earlier review comments. The rejection is about compatibility with this project's core durability model, not the quality or usefulness of Outl itself.

PR #294, opened 2026-07-29
fix(hooks): recognize antigravity-cli native tools and capture edit content

marcelomogami:
## Summary
- Antigravity CLI's native tools (`view_file`, `replace_file_content`, `multi_replace_file_content`, `write_to_file`, `list_dir`, `grep_search`, `read_url_content`, `read_resource`, `call_mcp_tool`) fell through `capture_policy::family()` to `ToolFamily::Unknown`, so `safe_tool_body()` discarded all content for every Antigravity file/search observation — session-end LLM consolidation had nothing to summarize.
- `family()` and `direct_paths()` now recognize these tools and their path argument keys (`TargetFile`, `SearchPath`, `DirectoryPath`).
- Fixing `family()` alone still left edit tools rendering `(no output captured)`: Antigravity's `post-tool-use` payload never sends a top-level `tool_response`/`output`/`result` field — the written content lives nested under `toolCall.args` instead. `extraction_candidates()` now also looks there, and `safe_tool_body()` recognizes the real argument keys, confirmed against live payloads: `CodeContent` (`write_to_file`), `ReplacementContent` (`replace_file_content`), and `ReplacementChunks` (`multi_replace_file_content`, an array of edit objects).
- Antigravity's read/search tools still summarize to `tool_family` + outcome only: its `post-tool-use` payload does not echo back read/search results, so there is nothing to capture for those.

## Test plan
- [x] `cargo test -p ai-memory-hooks` (190 passed), including new unit tests using real payload shapes captured from a live `antigravity-cli` session (`view_file`, `replace_file_content`, `list_dir`, `grep_search`, `write_to_file`, `multi_replace_file_content`)
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (crate-scoped, clean)
- [x] `cargo deny check` (advisories/bans/licenses/sources all ok)
- [x] End-to-end validated against a live `antigravity-cli` session running the fixed binary: confirmed via direct SQLite inspection that `write_to_file`/`replace_file_content`/`multi_replace_file_content` observations now contain the real written content instead of `(no output captured)`
- [ ] `cargo test --workspace` has 4 pre-existing, unrelated failures in `ai-memory-cli` (`commands::hook::tests::devin_*`) reproducible on a clean `main` checkout — not touched by this change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #296, opened 2026-07-30
test(cli): make Devin cwd-fallback fixtures hermetic

marcelomogami:
## Summary

Follow-up to #294, where I reported four failing `commands::hook::tests::devin_*` tests and you correctly noted they did not reproduce (`all 36 Devin-filtered CLI tests pass`). This explains why, and makes them environment-independent.

The four tests pass **relative** fake cwd values (`"process-project"`, `"env-project"`, `"payload-project"`). `marker::absolute_normalized()` falls back to joining a non-canonicalizable **relative** path onto the real process cwd:

```rust
path.canonicalize().unwrap_or_else(|_| {
    if path.is_absolute() { path.to_path_buf() }
    else { std::env::current_dir().map(|cwd| cwd.join(path))... }
})
```

So the marker walk-up starts inside the actual checkout. On a machine where an ancestor directory of the checkout has a real `.ai-memory.toml`, `find_marker` picks it up and `marker_query_suffix` appends an unexpected `&workspace=…`, failing all four assertions. CI has no marker above the checkout, so it is green there — the failure is deterministic wherever one exists, which is plausible for anyone who dogfoods ai-memory and clones it inside a marked workspace.

Switching the fixtures to absolute paths skips that fallback branch entirely: for an absolute path outside `$HOME`, `find_marker` bounds the walk at `checkout_root().unwrap_or(start)`, so it makes exactly one `.ai-memory.toml` check and returns `None` deterministically.

This matches the convention already established in the module that owns the function under test — `hook_capture.rs`'s marker tests use absolute fakes or `tempdir()`, e.g.:

```rust
fn query_suffix_without_marker_has_only_cwd() {
    let qs = marker_query_suffix("/nonexistent/path/xyz", None);
    assert_eq!(qs, "&cwd=%2Fnonexistent%2Fpath%2Fxyz");
}
```

The four `devin_*` tests were the outliers. Assertions are updated to the percent-encoded form for the same reason that precedent test does it; since `marker_query_suffix` encodes the caller's literal input string before any normalization, the expected values are platform-stable (that precedent already runs on the non-gating Windows job).

Test-only change — no production behavior and no CHANGELOG entry, per CONTRIBUTING's test-only exemption.

## Test plan
- [x] `cargo test --workspace` — full suite passes, including the four previously-failing `devin_*` tests
- [x] `cargo clippy -p ai-memory-cli --all-targets -- -D warnings`
- [x] `cargo deny check` (advisories/bans/licenses/sources ok)
- [x] Verified the failure reproduces on a clean `main` checkout in an environment with an ancestor marker, and stops reproducing with this change


akitaonrails:
Maintainer audit complete.

**What was right**
- The reported failure is real: relative fake cwd values can be resolved beneath the test process cwd, allowing an ambient ancestor `.ai-memory.toml` to change these assertions.
- The change is correctly test-only; no CHANGELOG or user documentation update is needed.

**Adjustment applied before merge**
- `/env-project`-style fixtures avoid the observed Unix failure, but root-relative path semantics are not fully portable to Windows.
- I preserved the contributor commit and merged current `main` normally, then added `68186bd` to extract the pure cwd-selection step. The tests now exercise payload/env/process precedence without invoking marker discovery or touching the host filesystem. Production URL/marker behavior is unchanged.

**Verification**
- `git diff --check`
- `cargo deny check`
- companion importer fmt/test/clippy

All passed locally. Recommendation: merge after the fresh hosted checks for `68186bd` pass.

PR #297, opened 2026-07-30
fix: supersede stale automatic handoffs

### LINKED ISSUE #293: Open handoffs are never garbage-collected: superseded auto-handoffs accumulate forever, and cwd specificity can deliver a stale one to a new session
**Problem this would solve**

Every non-managed `SessionEnd` creates one auto-handoff in state `open`, but each `SessionStart` accepts at most one (the selection winner). The losers stay `open` forever: the only transition of a handoff to `expired` in the codebase is the manual `memory_handoff_cancel`, which requires the exact `handoff_id` (and the docs frame it as a cleanup for mistakenly-created handoffs). Neither the retention forget sweep nor any other scheduled job modifies the `handoffs` table.

Two consequences, measured on one single-user workstation install:

1. **Unbounded accumulation.** Net growth of +69 open handoffs in 11 days (audit_log: 387 `insert_handoff` vs 318 `accept_handoff`), reaching 536 opens in under two months of use (first auto-handoff 2026-06-05, measured 2026-07-25). 98.5% of them (528) were superseded in practice: a newer `accepted` handoff existed in the same workspace/project. The briefing's pending-handoff count inflates accordingly.
2. **Stale delivery (the sharp edge).** Among auto handoffs, selection prefers the more specific cwd over recency (the intentional matching semantics from #114). A session started in a subdirectory that has an old open handoff receives that weeks-old handoff instead of the newer one from the parent scope. The only staleness cue is the `created` timestamp in the block header; nothing signals that a newer handoff exists in the parent scope, so the session consumes it as current context.

Cleaning up today requires going around the tool surface: the losers cannot be enumerated read-only through any tool or endpoint. Repeatedly calling `memory_handoff_accept` would surface them one at a time, but marks each one `accepted` (falsifying consumption data), after which `cancel` refuses them. My cleanup required reading the SQLite index directly (read-only) to enumerate ids and then 528 individual `memory_handoff_cancel` calls.

**Proposed solution**

Layered, in whatever subset makes sense (the third item is useful even if the first two are rejected):

1. **Expire-on-accept.** When a `SessionStart` accepts the winning handoff for a scope, mark older `open` AUTO handoffs of the same workspace/project (or, narrower, those whose cwd is within the accepted handoff's scope) as `expired`, in the same transaction. Event-driven, no new job, no new config; semantically the losers are obsolete by definition once a newer handoff was consumed. Manual handoffs would be exempt (they were created deliberately).
2. **TTL backstop in the retention sweep.** Include `open` AUTO handoffs older than a configurable age (e.g. 14 days) in the scheduled forget sweep, covering scopes that never see another accept. The sweep is already multi-scope (#173) and its cadence is persisted across restarts (#192), so the infrastructure for this is in place and reliable.
3. **At minimum, expose the ids: `GET /admin/handoffs?state=open`.** v1.19.0 added `GET /admin/open-sessions`; the same symmetry for handoffs would let operators audit and clean up without touching SQLite directly.

**Alternatives considered**

- **Manual cleanup with `memory_handoff_cancel`:** works (I ran 528 of them) but requires SQL access to enumerate ids, does not scale, and does not prevent recurrence.
- **Changing the selection order (recency over specificity):** would mask the stale-delivery symptom but changes the intentional matching semantics from #114 and does not address the accumulation; not proposed here.
- **Doing nothing:** storage cost is negligible (rows are small), but the stale-delivery edge and the inflated pending count remain.

**Additional context**

Reproduction (isolated instance, stock v1.19.0 binary; re-verified unchanged at v1.19.2, where the only handoff transition to `expired` is still the exact-id cancel and the sweep still does not touch the `handoffs` table; all synthetic):

1. Two sessions end in the same cwd, each leaving an auto-handoff; a new session starts in that cwd. The newest handoff is accepted; the older one stays `open`. States afterwards: `1 accepted, 1 open`: the open one is permanently orphaned (repeat N times: N-1 orphans).
2. `ai-memory forget-sweep --project <p>` right after: `candidates_evaluated` covers pages only; handoff states are byte-identical before/after. No retention path applies.
3. Stale delivery: an older auto-handoff from `<checkout>/sub` plus a newer one from `<checkout>`; a new session starting in `<checkout>/sub` receives the OLDER one (specificity beats recency). On the workstation above, the same shape delivered a weeks-old handoff to a fresh session in a rarely-used subdirectory.

After the manual cleanup the ledger reconciles exactly (561 expired = 528 cancelled + 33 earlier manual cancels; every expired row is an auto handoff with a newer accepted in the same scope), so the measurement is reproducible from the audit trail.

The `expired` state has existed in the schema since the handoffs table was created, but nothing automatic has ever reached it (verified with `git log -S` across the full history), so this is a feature request rather than a regression report. Related: #114 (selection semantics), #173 and #192 (the sweep infrastructure the TTL backstop would ride on), and the v1.19.0 `GET /admin/open-sessions` endpoint (the symmetry argument for item 3).

akitaonrails:
## Summary
- prefer the newest cwd-eligible automatic handoff while preserving manual precedence and path-boundary isolation
- expire prior automatic handoffs from the exact cwd on insertion, then expire older eligible automatic candidates atomically on acceptance
- preserve manual and sibling-directory handoffs and roll back cleanup when a managed startup claim or insertion fails
- document the updated handoff lifecycle and selection semantics

## Audit decisions
- no TTL: age-based expiry could erase legitimate context for a rarely used directory
- no new admin endpoint: insertion and acceptance now prevent the recurring orphan pattern without expanding the public/admin surface

## Tests
- `git diff --check`
- `cargo deny check`
- companion importer fmt/test/clippy gate

Closes #293


akitaonrails:
Maintainer PR audit complete.

**Pros**
- Fixes stale delivery by ranking eligible automatic handoffs by recency before cwd specificity.
- Bounds unattended same-directory growth at insertion and expires older eligible automatic candidates at acceptance.
- Preserves manual handoffs, sibling-directory work, workspace/project isolation, path-boundary matching, and atomic managed-startup claims.
- Adds rollback, selection, cleanup, sibling/manual isolation, and real SessionStart wiring regressions.
- Updates README, usage, architecture, design rationale, and CHANGELOG.

**Tradeoffs checked**
- No TTL was added because age alone cannot prove a rarely used handoff is obsolete.
- No admin listing endpoint was added because the recurring orphan pattern is now prevented without expanding the public surface.
- A genuinely abandoned unique-directory handoff can remain open until consumed or explicitly canceled; that is retained context rather than a superseded duplicate.

**Verification**
- Full mandatory local gate passed, including companion importer checks.
- All hosted CI/CodeQL jobs passed on exact head `eb9745019a6137e648df9a1642ef7f6f5c66e2cf`, including Linux/macOS/Windows tests, release builds, fresh source install, Docker smoke, audit/deny, clippy, rustfmt, gitleaks, and packaging.

Recommendation: merge.

PR #300, opened 2026-07-30
feat(llm): openai-compat embedder for self-hosted engines

samirhvbr:
## What changed

Adds a fifth embedding provider, `openai-compat`, for self-hosted engines that speak the OpenAI `/embeddings` shape — Ollama, LM Studio, vLLM.

- `AI_MEMORY_EMBEDDING_PROVIDER=openai-compat` requires `AI_MEMORY_EMBEDDING_BASE_URL`, `AI_MEMORY_EMBEDDING_MODEL`, and `AI_MEMORY_EMBEDDING_DIM` to be set explicitly — there is no safe default model or dimensionality for a self-hosted engine, so each is required rather than guessed. Each missing one fails at startup with a message naming the env var.
- Unlike the other providers it is **keyless**: a bearer token is sent only when `LLM_API_KEY` is present, for gateways that want one.
- Embeddings store under their own `provider="openai-compat"` identity.

**Changed default / behaviour note:** `EmbedderConfig.api_key` changes from `SecretString` to `Option<SecretString>`. No existing configuration changes behaviour — openai / voyage / google still require their key and fail identically when it is missing.

Switching an existing `openai` + custom-base-URL setup over to `openai-compat` changes the stored `{provider, model, dim}` triple, so those vectors stop matching at query time — run `ai-memory embed --force` to re-embed. This is documented in `docs/install.md`.

## Why

The existing OpenAI embedder could not express a keyless endpoint: `api_key` was a required `SecretString`. Local engines are the common case for people who don't want embeddings leaving their machine, and they run without auth.

The separate provider identity matters rather than reusing the `openai` label: embeddings are stored under a `{provider, model, dim}` triple, and reusing the label would silently mix vectors from two different models in one index.

## Test plan

- [x] Manual test: started `ai-memory serve` with `AI_MEMORY_EMBEDDING_PROVIDER=openai-compat`, a localhost base URL and no API key set — server starts and reports the embedder as `openai-compat`. Verified each missing-setting case (model, dim, base URL) aborts at startup with the env var named in the error.

New unit tests cover the keyless path, the `LLM_API_KEY`-as-gateway-token path, and all three fail-closed cases; `crates/ai-memory-llm/tests/openai_compat_embedder.rs` covers the request/parse round trip.

## Commit attribution

## CHANGELOG (merge gate)

## Notes for reviewers

The shared request path was extracted into `openai_style_embed` so the compat client and the existing OpenAI client cannot drift on retry/truncate/normalise behaviour. The existing providers' behaviour through that helper is unchanged — worth a look to confirm.

This is the first of a stack of 7 independent features; the rest follow as separate PRs once this lands, so each stays reviewable on its own.


akitaonrails:
Maintainer audit: **approve after adjustment**.

Pros:
- The feature matches the documented OpenAI-compatible `/v1/embeddings` contracts exposed by Ollama, LM Studio, and vLLM.
- A distinct `provider="openai-compat"` identity preserves the stored `{provider, model, dim}` mismatch guard.
- Requiring an explicit base URL, model, and dimension avoids guessing across self-hosted engines.
- The shared OpenAI-shape request helper keeps truncation, retry, parsing, dimension checks, and normalization aligned with the existing OpenAI embedder.
- Keyless requests omit `Authorization`; an explicitly configured `LLM_API_KEY` still supplies gateway bearer auth.

Adjustments made before merge:
- Merged current `main` normally, preserving contributor history.
- Kept `EmbedderConfig.api_key: SecretString` and `default_embedding_dim(...) -> u32` source-compatible instead of changing both public Rust APIs for this additive provider. A new `try_default_embedding_dim` exposes the no-default case cleanly.
- Added a factory-level zero-dimension rejection.
- Added regression coverage for keyless requests, bearer requests, missing base URL, zero dimensions, distinct provider identity, and the existing dimension API.
- Updated the README support matrix, architecture config reference, design decision, install guide, and CHANGELOG attribution `(#300)`.

Local verification passed:
- `git diff --check`
- `cargo deny check`
- companion importer fmt/test/clippy

Hosted checks are still completing; merge remains gated on the full matrix.

PR #310, opened 2026-07-30
feat(multiuser): scope handoffs, sessions and slots to their owner

djalmajr:
## What changed

On a server shared by several people, a pending handoff was delivered to
whichever session started next, whoever it belonged to, and one operator's
memory slots were injected into everyone's session brief. This adds per-operator
ownership to the state that is per-operator by nature — handoffs, sessions,
slots, auto-improve proposals, access reinforcement — and a way for an
authenticating proxy to say who the caller is.

**Nothing changes at default config.** Every new setting is off (`[slots]
per_user`, `[auth].actor_proxy_secret`) or zero (`[decay] breadth_weight`), and
the single rule that makes this adoptable is *absent = shared*: a NULL owner, or
a slot with no namespace, stays visible to everyone. That is the shape of
everything already stored, so turning a flag on cannot hide or reinterpret
existing data, and turning it back off cannot orphan what was written while it
was on. Regression tests assert the default-config path explicitly rather than
leaving it inferred.

Observable behaviour, before → after:

- A pending handoff went to the next session to start → it goes to its owner;
  unowned handoffs still go to anyone, as before.
- Every operator's slots reached every brief → personal slots
  (`_slots/<user>/…`) reach their owner, shared slots still reach everyone.
- One pending auto-improve proposal per page, server-wide → one per page **per
  operator**, where the deployment distinguishes operators.
- `hard_delete_decayed_pages` ran `DELETE FROM pages` with no scope → it is
  scoped to the swept project.
- New: `GET /api/v1/workspaces/{ws}/projects/{p}/handoffs`, the `forget_sweep`
  / `handoff_begin` / `handoff_accept` / `handoff_cancel` admission ops, and the
  `X-Memory-Actor-Proxy-Secret` identity overlay.

**Changed defaults: none.** Two behaviours change only once a new flag is set,
and are called out because they are the ones a reviewer should push on:

- With `[auth].actor_proxy_secret` configured, a proxied caller the proxy names
  drops from root to user tier, and `[auth].root_username` becomes **required**
  — the server refuses to start without it, because otherwise nothing could ever
  reach `/admin/*` or create the first DB user.
- With `[slots] per_user = true`, a slot write is placed into the writer's own
  namespace rather than the shared slot.

## Why

The engine already had the state that makes a shared server work — batons,
sessions, slots — but no notion of whose it was. On a single-operator install
that is exactly right and should stay free. Once two people share an instance,
the same state silently crosses between them, and the failure is quiet: you get
somebody else's context, not an error.

`OwnerFilter` and *absent = shared* are the whole design. One enum answers "may
this actor see this row", one idiom means legacy rows and single-user rows are
the same case, and the migrations are additive so a populated database needs no
backfill and no downtime.

## Test plan

- [x] `cargo test --workspace` passes — 52 suites, 2114 tests, 0 failures
      (run as `--all-targets --no-fail-fast`; without `--no-fail-fast` cargo
      stops at the first failing binary and reports a fraction as if it were all)
- [x] `RUSTFLAGS="-D warnings" cargo build --release`, `cargo deny --all-features
      check`, and the separately-manifested `companions/ai-memory-importer` crate
- [x] **Manual test: two operators behind one proxy, on a real server.**
      `docker/multiuser-test/` brings up the server behind an SSO-terminating
      nginx and drives it as three people. 14 assertions, 0 failures:
      slot writes re-home per operator; a session brief carries its own
      operator's slot body and never another's; the shared slot still reaches
      everyone; one operator cannot consume another's baton; a proxied non-root
      caller is denied the sweep; an operator named by OIDC **subject claim
      alone** gets a real namespace and sees their own data; a forged actor
      header without the shared secret is ignored; a duplicated one fails closed
      with 400.
- [x] **Upgrade against a populated production database.** A snapshot of a live
      v1.14.0 deployment (~850 pages, ~29k observations, 251 handoffs, 3
      workspaces, 25 projects) migrated V28 → V41 on boot with zero row loss
      across every table. Every legacy handoff and session carries a NULL owner
      afterwards, and legacy handoffs are still delivered — both to an
      unattributed caller and to an attributed one, which is the case where an
      ownership filter could hide them.

Two limits worth stating rather than leaving implied:

- That database happened to contain **no slot pages**, so the "a pre-existing
  nested slot must keep reaching every brief" regression is covered by unit test
  only, not by the upgrade run.
- Rollback is not in-place: an older binary opening the migrated schema gets
  `DataSchemaAhead` and refuses to start (correctly — it does not corrupt).
  Rolling back means restoring a backup.

Note for anyone reproducing locally: `crates/ai-memory-cli/tests/autoscope_env.rs`
is load-flaky — it spawns processes and treats "alive after 5 s wall-clock" as a
verdict, so it can go red under a parallel full-workspace run and passes
serially. It is unchanged by this branch.

## Commit attribution

- [x] I verified the name and email on every commit in this PR and corrected any
      unintended identity before requesting merge.

## CHANGELOG (merge gate)

- [x] Any changed **default** is called out explicitly in "What changed" above —
      there are none; the two flag-gated behaviour changes are listed there.

## Notes for reviewers

**SQLite specifics worth a second pair of eyes.** NULLs are DISTINCT in a UNIQUE
index, so preserving the single-user "one pending proposal per page" invariant
while adding a per-author bucket needs `COALESCE(col, x'00')` rather than a plain
composite index. And `*`, `?` and `[` are GLOB metacharacters: a username used as
a slot namespace segment is validated against them, because `a*` would otherwise
match every other operator's slots.

**Two places where the rule is deliberately NOT applied**, both easy to "fix"
into a bug:

- `asserts_root_identity` compares `[auth].root_username` against the asserted
  *username*, not the general identity key. Keying it on the OIDC subject would
  let a subject string equal to `root_username` acquire root capability.
- Ownership is stamped only where the deployment has somebody to separate. On a
  single-operator server that merely names its operator via `root_username`,
  stamping the one name would separate that operator's *transports* — HTTP
  carries the name, the stdio MCP transport and the local CLI carry no actor —
  and each would stop seeing what the other wrote.

**Where I'd look hardest.** The recurring failure mode while building this was a
rule enforced at one door and not its sibling: the MCP tool guarded while the
engine's own write path was not, one capability gate fixed while its
`route_layer` twin was missed, an HTTP response carrying a field the CLI struct
silently dropped. Each of those is fixed, but the shape is worth hunting for
rather than trusting.

**Migrations** V37–V41 are additive (`ALTER TABLE ADD COLUMN` with NULL
defaults, plus new tables and indexes). None rewrite a table, so they are O(1)
on a populated database — confirmed against a populated snapshot, see the test
plan. They have been renumbered twice to sit above migrations `main` added while
this branch was open; note that such a collision produces **no merge conflict**,
because the filenames differ, and `refinery` only rejects the duplicate at
startup.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,11 +63,12 @@ Many projects use CLAUDE.md for Claude Code and
 AGENTS.md for Codex / OpenCode / Cursor / Gemini CLI / Grok Build CLI / Kimi Code,
 but if the project says one file is canonical, use that file.
 
-If the rule is a standing *user/team* preference that should apply to
-every project (tech choices, code style, personal conventions), save it
-to ai-memory's reserved global scope instead — the durable-pages skill
-covers how. Default memory reads surface global-scope pages in every
-project automatically.
+If the rule is a standing preference that should apply to every project
+(tech choices, code style, durable conventions), save it to ai-memory's
+reserved global scope instead — the durable-pages skill covers how.
+Default memory reads surface global-scope pages in every project
+automatically. That scope is server-wide: on an instance shared with
+other people, a rule saved there replaces the one they see too.
 
 ### Refreshing this snippet
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,460 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Memory slots may be namespaced per operator (`_slots/<user>/…`). Slots are
+  injected into every session start for their project, so on a shared server one
+  person's working context silently became everybody's. A slot with NO namespace
+  is shared and stays visible to everyone — the shape of every slot written
+  before this — so enabling it hides nothing already stored. `[slots] per_user`
+  (default off) governs the whole regime: with it on, the engine namespaces the
+  slots it writes, session briefs and consolidation prompts carry the shared
+  slots plus the requesting operator's own, and a write into another operator's
+  namespace is refused (admins may still curate). With it off, a nested slot path
+  carries no ownership meaning and every slot reaches every brief exactly as it
+  did before, so turning the flag back off restores the old behaviour in full
+  ([#286]).
+- Auto-improvement proposals record who staged them, and the one-pending-per-
+  target rule is scoped to that person (migration V42). One operator's pending
+  suggestion for a page used to block everybody else's. Correct in both modes
+  from a single index, with no runtime switch: the operator is folded in through
+  `COALESCE(staged_by_actor_user, '')`, so unattributed rows (single operator, or
+  any caller the server cannot name) collapse into one bucket and keep the
+  original invariant exactly. A plain `UNIQUE (…, staged_by_actor_user)` would
+  not — SQLite treats NULLs as distinct, so every existing single-operator
+  database would silently start accepting unlimited pending proposals per page.
+  The column holds the actor username rather than a `users(id)`, matching
+  `handoffs.owner_user` / `sessions.actor_user` / `page_access.actor`: operators
+  named by a trusted proxy have no row, and keying on the id would leave every
+  one of them NULL and collapse them back into a single bucket. A caller is
+  recorded only where the deployment actually distinguishes operators — the same
+  question the admin gates ask — because the scheduler and the report handlers
+  stage unattributed by design, so bucketing an interactive call by
+  `[auth].root_username` on a single-operator server would leave two proposals
+  pending for one page ([#286]).
+- Page reinforcement is recorded per operator in a new `page_access` table
+  (migration V43), alongside — not replacing — `pages.access_count`. The scalar
+  cannot distinguish "50 reads by one person" from "one read by each of 50
+  people", though only the second says a page is load-bearing for a team. The
+  retention formula gains an optional breadth term weighted by `[decay]
+  breadth_weight`, which defaults to `0.0` and is provably identity: at the
+  default weight, and at 0 or 1 distinct readers under any weight, scores are
+  unchanged. So there is no eviction cliff and no backfill. The forget sweep
+  reads the distinct-operator count for its candidates in one grouped query —
+  and skips that query entirely while the weight is `0.0`, so a deployment that
+  never enables the term never pays for it. A read whose page was deleted in the
+  meantime records nothing for that page and leaves the rest of the batch's
+  reinforcement intact ([#286]).
+- Zed editor as an MCP-only client. `install-mcp --client zed` renders,
+  and `--apply` idempotently merges, a native remote HTTP entry under the
+  top-level `context_servers` map in Zed's platform user `settings.json`,
+  with optional bearer headers. The JSONC-aware apply and uninstall paths
+  preserve user comments, trailing commas, unrelated settings, and sibling
+  servers while changing only the matching ai-memory entry. Zed does not
+  provide lifecycle hooks or managed-workstream continuity. (#321)
+- Entity-match retrieval as a fourth RRF stream (V38 `entities` +
+  `entity_page_links`). Consolidation emits up to 10 normalized technologies,
+  components, services, files, or domain nouns per page into frontmatter;
+  manually edited `entities` use the same index path, and reindex rebuilds the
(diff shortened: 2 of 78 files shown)


akitaonrails:
Draft audit against current `main` (`3910fdf`) stopped before checkout/execution, as required for an explicitly unfinished PR. The direction is useful, but this head is not mergeable as one unit yet.

Concrete blockers from the static pass:

1. **Migration collision:** current `main` already owns V36 (`V36__page_expiry.sql`), while this draft starts at `V36__handoff_ownership.sql`. Refinery will reject duplicate versions at startup. The branch needs a fresh normal merge from current main and a new contiguous migration allocation; do not rewrite the existing contributor history.
2. **Identity-key collision/instability:** `ActorContext::identity_key()` stores an unqualified raw username or raw OIDC subject in the same TEXT namespace. A username equal to another identity's subject aliases ownership, and a sub-only proxy that later starts providing a username re-buckets that operator despite the stability rationale. Ownership keys need an explicit, stable, namespaced identity contract before they can protect handoffs or slots.
3. **Product-contract change needs separation:** shipped multi-user behavior is documented as single-tenant attribution, not per-user data isolation. Handoff/session ownership, proxy assertion, personal slot injection, proposal ownership, and access-breadth retention are distinct security/product changes and should not share one 14,638-line review unit.
4. **Confirmed independent bug should be extracted:** current `hard_delete_decayed_pages` is unscoped even though `run_sweep` targets one `(workspace, project)`. That can permanently delete eligible tombstones belonging to sibling projects. This merits a small focused fix and regression independently of the multi-operator feature.
5. **Current-main composition is unaudited:** this draft predates merged TTL and consolidation-preference behavior and changes the same sweep, consolidator, wiki, routing, docs, and migration surfaces. Those interactions need to be integrated and tested in the smaller PRs.

Recommended split order:

- scoped retention hard-delete bug fix;
- trusted-proxy identity contract and authorization tests;
- handoff/session ownership and recovery semantics;
- optional per-user slot injection;
- proposal ownership/collision reporting;
- opt-in access-breadth retention.

Each security boundary should carry root, DB-user, proxy-user, anonymous/no-auth, cross-project, cross-workspace, legacy-row, disable/re-enable, and migration-upgrade tests. This draft remains open and unmodified; no contributor code was executed.

djalmajr:
Split per the audit. Status of the five blockers:

1. **Migration collision** — resolved here by successive merges from `main`
   (this head is current with `main` and CI-green), and structurally avoided in
   the slices: each allocates against current `main`, so the handoff slice
   takes V39–V41 and the proposals/breadth slice V42–V43.
2. **Identity-key collision/instability** — addressed in #333: ownership keys
   are qualified (`sub:<subject>` / `user:<name>`), so a username equal to
   somebody else's subject is unrepresentable as the same identity; `sub`
   outranks `user` per OIDC's stability contract, with the one remaining
   re-bucket path (username-only ingress later adding `sub`) documented in
   `docs/users.md`; wiki path segments use an injective filesystem-safe
   derivation rather than raw values.
3. **Separation into review units** — this PR is superseded by:
   - #333 — trusted-proxy identity contract and authorization (no migration)
   - #334 — handoff/session ownership and recovery semantics (V39–V41)
   - #335 — optional per-user slot injection (no migration; includes the
     two-operator proxy harness under `docker/multiuser-test/`)
   - #336 — proposal ownership/collision reporting + opt-in access-breadth
     retention (V42–V43)

   Recommended merge order: #333 → #334 → #335 / #336. #336's migrations must
   land after #334's, or refinery's out-of-order rule rejects V39–V41 on a
   database already at V43.
4. **Scoped retention hard-delete** — fixed independently on `main`
   (`6ade07e`), whose version is a strict superset (same predicate, plus the
   transaction-scoped entity cleanup); the duplicate this branch carried was
   dropped rather than merged.
5. **Current-main composition** — the slices are cut against current `main`
   and each carries its own suite; this branch additionally validated the
   composed behaviour end to end: a two-operator run behind an
   SSO-terminating proxy (slot placement/visibility, handoff ownership, admin
   gating, header forgery, sub-only ingress), and an upgrade of a populated
   production snapshot (~850 pages, ~29k observations, 251 handoffs) with
   zero row loss and legacy NULL-owner rows still delivered to attributed and
   unattributed callers.

Closing in favour of the slices; this branch stays as the integrated
reference for that end-to-end evidence.

PR #317, opened 2026-07-30
feat(store): explain mode for memory_query

samirhvbr:
## What changed

`memory_query` gains an optional `explain: true`. Every hit then carries a `score_details` object:

- its 1-based rank and raw score in each retrieval stream — FTS5 bm25 score, embedding cosine, and graph-neighbour provenance naming the seed page and the link direction (`incoming`/`outgoing`);
- the per-stream RRF contributions and the fused total;
- the bounded page-authority multiplier applied after fusion (from #269).

The response also gains a top-level `streams_active` list, so degradation is visible: an embedder failure reads as `["fts","graph"]` instead of silently looking like a thin index.

**Changed defaults / behaviour:**

- Without `explain`, the response shape is unchanged.
- `hybrid_search_explained` is the new engine entry point; `hybrid_search` stays as a thin delegation, so existing callers are untouched.
- Equal adjusted scores now tiebreak on path. Previously they fell through to `HashMap` iteration order, so a tie could reorder between runs.
- No migration.

## Why

"Why did this page rank third?" had no answer short of reading the fusion code. Every input to that answer was already computed inside hybrid search and then thrown away — the FTS rank and score, the cosine, and the graph stream's `seed_ord`, which already encoded both the seed and the direction. This keeps them.

**The authority multiplier is reported deliberately, and it is the part worth scrutinising.** The returned `rank` is the fused score *after* that adjustment, so publishing `fused` alone would produce a surface that cannot account for the order the caller sees — precisely the confusion explain exists to remove. Two tests assert `rank == -(fused × authority)` rather than the pre-#269 `rank == -fused`.

## Test plan

- [ ] Manual test: **not re-run against a live deployment on this rebase.** The surface was exercised end-to-end earlier against a local `ai-memory serve` (confirming `score_details` appears with `explain: true` and that the default response stays byte-identical), but that predates the rebase onto the merged #309. Coverage here is automated only.

## Commit attribution

## CHANGELOG (merge gate)

## Notes for reviewers

**This PR also fixes a pre-existing bug, and that fix is load-bearing for the feature.** `memory_query` branched to plain FTS5 whenever a query vector was absent, so the graph-neighbour stream never ran in the zero-LLM default — the documented default — despite `ARCHITECTURE.md` describing FTS + graph. Explain would have faithfully reported that as a single active stream. Both paths now run the same code; a stream with no input contributes nothing, exactly as the vector stream already did. If you would rather see that as its own commit, say so and I will split it.

Rebased onto `main` after #309 merged, including your `fix(ttl): preserve query and deletion invariants` — one new `QueryArgs` fixture in the TTL test you added needed the new field.

Follows #309. Independent of #316.


akitaonrails:
Maintainer audit completed on exact head `8aafb67`, then current `main` was merged normally and correction `69fce37` was added.

Corrections before merge:
- fixed zero-vector `streams_active` to report `fts` + `graph`;
- kept explain-only provenance bookkeeping off ordinary queries;
- preserved explanations through explicit multi-scope and reserved-global-scope results, while documenting the distinct FTS-only `global=true` behavior;
- added deterministic graph and multi-scope tie ordering;
- added the missing store/MCP/schema/prompt regression tests, including vector, graph provenance, authority reconciliation, no-explain compatibility, global behavior, and scope unions;
- moved the changelog entry back under `Unreleased` and synchronized README, usage, architecture, routing skill, base snippet, AGENTS, and stale zero-LLM log/comments.

Security/static review found no dependency, workflow, executable-mode, unsafe-code, process-execution, network, or secret-handling additions. Full local project gate passed: workspace tests, workspace clippy with warnings denied, cargo-deny, companion fmt/test/clippy, native packaging, and cargo-audit.

PR #318, opened 2026-07-30
feat(mcp): memory_feedback records page quality signals (V37)

samirhvbr:
## What changed

A 17th MCP tool, `memory_feedback`, records how useful a recalled page actually was. Call it with an exact `path` and one of four signals:

- `helpful` / `not_helpful` step a new per-page `pages.salience` column (V37) in bounded 0.25 increments, clamped to `[0.25, 2.0]`. Salience now scales the retention formula's time term, where a single global `salience_default` used to.
- `stale` / `wrong` floor the salience **and** surface the page as a `feedback_flagged` finding in the next `memory_lint` report.

Signals land in a new append-only `page_feedback` table with an optional sanitized reason and a full audit-log entry.

**Changed defaults / behaviour:**

- `salience` stays `NULL` until feedback moves it, and `NULL` scores exactly as the previous global default — so an untouched database behaves identically.
- Nothing here ever deletes a page. The strongest action is "floor the confidence and ask a human to look".
- New V37 migration: one nullable column plus one new table. Additive, no backfill.

## Why

The M8 decay formula's only reinforcement signal was the access counter a search hit bumps, which cannot distinguish "this page answered the question" from "this page surfaced and wasted a read". This is the *finer-grained reinforcement beyond access counts* item from `prior-art-implementation-findings.md`.

**Why there is no `resolved_at` column.** Signals attach to the page *version* they were left on, so the lint query joining `is_latest = 1` drops the finding the moment the page is rewritten. Fixing the page **is** the resolution; a separate dismissal state could only ever disagree with it.

**Why a new tool on a surface that is narrow on purpose.** It cannot ride on a read tool without conflating read and write semantics, and the counter it supplements cannot express the distinction. The rationale is recorded in `ARCHITECTURE.md` next to the other tools that post-date the original cut — the discipline still holds, the count is now 17.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo test --workspace` passes — 51 suites, 0 failures
- [ ] Manual test: **not re-run against a live deployment on this rebase.** Exercised end-to-end earlier against a local `ai-memory serve` — salience moved 1.0 → 1.25 → 0.25 across signals and a `stale` signal produced a `feedback_flagged` finding in the following `memory_lint` — but that predates the rebase onto the merged #309 and #317. Coverage here is automated only.

The decay property tests were extended to cover per-page salience rather than only the global default.

## Commit attribution

[...]

PR #319, opened 2026-07-30
feat(llm): optional LLM reranker after RRF fusion

samirhvbr:
> **Depends on #317 and #318.** This branch carries those commits as its base — **review only the third commit**, `feat(llm): optional LLM reranker…`. It collapses to a single commit once they merge.

## What changed

An opt-in relevance pass after fusion. Set `AI_MEMORY_RERANKER=llm` (which also requires `AI_MEMORY_LLM_PROVIDER`) and `memory_query` over-fetches `limit * 3` candidates from the RRF + authority stage — capped at 30 — then has the configured LLM score each one against the query.

It runs through `complete_structured`, so it is JSON-schema output on the existing provider trait: no new provider dialect, and all eight providers work unchanged. With `explain: true` each hit's `score_details` carries its `rerank_score`.

**Changed defaults: none.** Unset, the query path is byte-identical to today. When set:

- `memory_query` becomes LLM-latency-bound. That is the trade being bought and it should be stated plainly.
- Degradation is silent and total: a 20-second timeout, a provider error, an unparseable judgement, or a response covering fewer than half the candidates all return the untouched RRF+authority order rather than failing the query — the same contract the embedder already has.
- An unknown `AI_MEMORY_RERANKER` value, or `llm` with no provider configured, **fails at startup**.

No migration.

## Why

RRF fuses streams without ever reading the query against a candidate's actual content, so the right page can land seventh. This adds a stage that does read it, for deployments willing to pay an LLM call per search.

Failing closed at startup is deliberate: a typo'd env var that silently disabled reranking would read as "reranking is on" in the operator's head while every query quietly ran plain RRF.

## Test plan

- [ ] Manual test: **not performed against a live LLM provider.** The degradation paths and ordering are covered by tests with a stub reranker, and the startup-validation cases were exercised against a local `ai-memory serve` — but no real provider has scored real candidates through this. That is the scenario it exists for, and it deserves a reviewer's scepticism.

## Commit attribution

## CHANGELOG (merge gate)

- [x] Any changed **default** is called out explicitly in "What changed" above — there are none.

## Notes for reviewers

**The coverage floor is the subtle part.** If the model returns scores for only a handful of candidates, using them would let a couple of low scores outrank candidates it never judged. Below half coverage the whole response is discarded and RRF order stands. Worth confirming you agree with the threshold.

**Ordering, stated explicitly:** RRF → authority adjustment (#269) → rerank. The reranker reorders positions that authority has already adjusted, not the raw fused scores, and `rerank_score` in the explain documents that. If you would rather it run before the authority multiplier, that inverts which signal wins ties and is worth deciding deliberately.

Rebased onto `main` after #309 merged, including your `fix(ttl): preserve query and deletion invariants`.


akitaonrails:
## Maintainer audit: adjusted before merge

Audited original head: `ce8ca8c0d91d8c56f02074133547faf4d17863d5`
Final audited head: `978ade15e1d88df861b6eed6bbb9cb62fd76a1d5`

Trust/hostile-change gate: clear. I found no malicious code, hidden execution, credential access, dependency/lockfile change, workflow or release expansion, executable-mode change, persistence mutation, or authorization/scope bypass. Contributor-controlled prose and prompt contents were treated as untrusted claims/data.

The feature is useful, but the original head was not mergeable as-is. Maintainer commit `978ade1` corrects these blockers without rewriting contributor history:

- reranks once after explicit multi-scope fusion instead of making one sequential provider call per scope;
- preserves requested limits above the 30-candidate judging cap;
- requires a complete, unique, finite `[0,1]` score set and otherwise preserves local order;
- uses page-version IDs, deterministic cross-scope tie-breaking, and keeps the unjudged tail in place;
- JSON-encodes bounded untrusted query/title/snippet input and states the prompt-injection boundary explicitly;
- caps provider work at one 20-second call per eligible query and four concurrent calls, degrading immediately on saturation;
- excludes the distinct `global=true` path and supplemental `_global` preference results, matching actual behavior;
- fails startup on an unknown reranker or `llm` without an LLM provider;
- adds privacy/cost documentation in README, SECURITY, architecture, install/usage, config template, and changelog;
- replaces the PR's mapping-only coverage with behavior tests for reorder, large limits, malformed/partial/provider-error/timeout fallback, saturation, one-call multi-scope fusion, global exclusion, prompt injection encoding, byte budgets, and startup/config validation.

Claim ledger:

- "Optional/off by default": confirmed.
- "Post-RRF relevance pass": confirmed for project and explicit-scope queries; corrected docs exclude global and supplemental-global paths.
- "Degrades on failure": confirmed after strict fallback fixes.
- "Stub reranker proves degradation and ordering": unsupported on the original head; confirmed by the added server-level tests on the final head.
- "Works through existing providers": confirmed through the existing typed `LlmProvider::complete_structured` boundary; no provider dialect or dependency was added.

Local gates on the final tree: fmt, `git diff --check`, full workspace tests, clippy `-D warnings`, cargo-deny, cargo-audit, importer fmt/clippy/tests, native packaging, 44 hook shell tests, security-surface review, and current-tree gitleaks all pass. Full-history gitleaks still reports six pre-existing synthetic sanitizer fixtures from old commits; the hosted gitleaks configuration is green and this PR does not touch those files.

Recommended action: merge only after every hosted check on exact head `978ade1` is green.

PR #320, opened 2026-07-30
feat(store): entity matches as a fourth RRF stream (V38)

samirhvbr:
> **Depends on #317, #318 and #319.** This branch carries those commits as its base — **review only the fourth commit**, `feat(store): entity matches…`. It collapses to a single commit once they merge.

## What changed

A fourth retrieval stream. Consolidation now emits an `entities` list per page — the specific technologies, components and domain nouns the content is *about* — which lands in the page's frontmatter and is indexed into two new tables (V38 `entities`, `entity_page_links`). `memory_query` matches query tokens against those names and fuses the result into the same RRF as FTS, vector and graph.

Matching is lexical: exact name, name-prefix, or a whole word inside a multi-word entity, weighted by inverse entity frequency so a rare noun outranks a ubiquitous one. With `explain: true` each hit reports `entity_rank` and `matched_entities`.

**Changed defaults / behaviour:**

- Extraction happens **only where an LLM already runs** — inside consolidation. No new LLM call, and none at query time.
- A project that never consolidated has an empty table, and an empty stream contributes nothing — exactly like the vector stream with no embedder. The zero-LLM default path is unchanged.
- New V38 migration: two new tables. Additive.

## Why

A page whose body says "the streaming broker" never answers a query for "jetstream". Lexical search has no shared token and vector search needs an embedder the default deployment does not have.

**Markdown stays the source of truth.** Entities live in frontmatter first; the tables are derived, so a reindex rebuilds them and a hand-edited page flows through the same watcher path. No spaCy, no NLP dependency, no graph database.

`entities.id` is deliberately shaped as the stable anchor a future temporal-triples table could reference. That P2 item stays deferred — this is the foundation, not the speculative half.

## Test plan

- [ ] Manual test: **not re-run against a live deployment on this rebase.** Exercised end-to-end earlier against a local `ai-memory serve` — a page whose body never contains "jetstream" but declares the entity `nats jetstream` was returned for that query, with no embedder and no LLM in the query path — but that predates the rebase onto the merged #309. Coverage here is automated only.

The `recall_eval` harness gained an `+entity` variant alongside the existing retrieval comparisons.

## Commit attribution

## CHANGELOG (merge gate)

## Notes for reviewers

**SQLite specifics worth a second pair of eyes.** `LIKE` patterns are escaped so a `_` inside an entity name cannot act as a single-character wildcard and match unrelated entities. Names are indexed in both hyphenated and spaced form, because a query for `writer-actor` should reach an entity recorded as `writer actor` and vice versa.

**Where the design could be wrong:** inverse-frequency weighting is a heuristic, not a tuned one. On a project where consolidation tagged everything with the same three nouns, the stream degrades to noise rather than silence — it would still contribute rank, just uselessly. A floor on entity rarity would fix that; I did not add one because I have no data on what the threshold should be.

The stream feeds the same candidate window the other three use, so its hits are eligible for the bounded authority adjustment rather than bypassing it.

Rebased onto `main` after #309 merged, including your `fix(ttl): preserve query and deletion invariants`.


akitaonrails:
Maintainer audit completed against the frozen contributor head `0a02b3ab79ea139e9b9fa5cdd88d6172ff0bf29c`. The contributor branch was fast-forwarded to the audited head `e506db221de4fc5c7d251a3b665e44bf26d716ef`; history was preserved.

**Claim verification**

- Confirmed: V38 adds a project-scoped entity index, consolidation and wiki reindex populate it, and `memory_query` fuses it as a fourth RRF stream without requiring an embedder or query-time LLM.
- Corrected: the original matching code did not fully support the stated hyphen/spaced word-prefix behavior. Matching now covers exact names, name prefixes, and word prefixes after spaces, hyphens, or literal underscores with escaped parameters.
- Corrected: the recall change adds an entity-only regression probe, not a new benchmark variant; documentation no longer overstates it.
- Removed: speculative statements that the schema was already a stable anchor for a future triples table.

**Maintainer corrections before merge**

- Added bounded typed entity normalization: 10 names/page, 64 characters/name, control-character rejection, de-duplication, and adversarial oversized-input coverage.
- Enforced workspace/project pairing and same-project entity/page links with migration triggers and upgrade tests.
- Kept entity link writes, supersession, delete cleanup, and reindex recovery transactional and aligned with immutable page versions.
- Scoped retrieval and inverse-frequency counts to the requested project, excluded expired pages by default, bounded query tokenization, escaped LIKE metacharacters, and handled extreme limits safely.
- Made explain output deterministic and lossless for names containing commas; it now reports `entity_rank`, raw `entity_weight`, `matched_entities`, and RRF contribution.
- Preserved current-main behavior from PR #319 by removing stale reranker/feedback tests reintroduced by the old branch base.
- Updated README, CHANGELOG, SECURITY, architecture, usage, deployment, install, design, prior-art, vector-policy, and retrieval-skill documentation.

**Security review**

No new dependency, workflow, build/release, executable-mode, process-launch, outbound-network, credential, auth, or permission surface was added. SQL is parameterized, inputs and result sizes are bounded, migration triggers fail closed on scope mismatch, expired rows do not influence ranking, and cross-project isolation has explicit decoy tests. Gitleaks found no secrets, Cargo audit/deny passed, and the security-surface scan found no feature-scope malicious or exploitable behavior.

**Verification**

- `git diff --check origin/main...HEAD`
- `cargo deny check`
- `cargo audit`
- Standalone importer fmt/clippy/tests: 11 passed
- Native packaging check
- Hook shell tests: 44 passed
- Handoff e2e smoke: 9 passed
- Focused entity migration, scope, expiry, cleanup, reindex, explain, and recall regressions all passed

Recommendation: merge after hosted checks complete successfully for exact head `e506db221de4fc5c7d251a3b665e44bf26d716ef`.

PR #333, opened 2026-07-31
feat(auth): qualified identity keys and trusted-proxy identity assertion

djalmajr:
## Summary

- Add a dedicated trusted-proxy bearer credential that is distinct from the root bearer.
- Accept proxy identity as either a username or the complete OIDC issuer/subject pair.
- Use qualified identity keys consistently for hook and MCP active-project routing.
- Treat trusted-proxy deployments as multi-operator for `/admin/*` and `memory_forget_sweep` authorization.

## Security contract

- `[auth].actor_proxy_bearer_token` requires a separate nonblank `[auth].bearer_token`; equal credentials fail startup.
- Proxy requests fail closed when identity is missing, partial, duplicated, or comma-folded.
- OIDC identity is keyed by `(issuer, subject)`, not `sub` alone.
- Proxy usernames never grant root. Proxied root access requires an exact configured `root_issuer` + `root_subject` match.
- Raw actor headers on root and DB-user requests remain ignored.
- The proxy must replace, never append, client-supplied `X-Memory-Actor-*` headers.

## Documentation

Updated the generated config template, README, changelog, multi-user guide, and auto-scope guide.

## Verification

- `git diff --check`
- `cargo deny check`
- focused proxy auth, admin capability, MCP capability, auto-scope stress, and hook tests
- differential gitleaks scan

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   value, and headers without the secret are ignored as before. The server
   refuses to start with a proxy secret but no `root_username`, because no
   proxied request could ever reach a root-only capability again. Unset — the
-  default — nothing changes (#310).
+  default — nothing changes (#333).
 - Qualified identity keys: every ownership decision resolves a request to
   `sub:<subject>` or `user:<name>` via one accessor
   (`ActorContext::identity_key`), never a bare string, so a username equal to
@@ -25,12 +25,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   filesystem-safe `path_segment` derivation ships alongside for per-operator
   wiki paths. `OwnerFilter` / `owner_stamp` carry the read and write sides of
   the same contract, with "absent = shared" as the compatibility rule
-  (#310).
+  (#333).
 - The `/admin/*` route layer and the MCP `memory_forget_sweep` tool now ask
   "does this deployment distinguish operators" instead of "do `users` rows
   exist", so a trusted-proxy deployment — which never writes a `users` row —
   gets root-only admin gating instead of waving every proxied caller through
-  the single-operator escape hatch (#310).
+  the single-operator escape hatch (#333).
 
 ## [1.21.0] - 2026-07-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,24 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Trusted-proxy identity: a proxy that terminates SSO can name the real end
-  user by echoing `[auth].actor_proxy_secret` in
-  `X-Memory-Actor-Proxy-Secret` beside the root bearer. Asserted callers drop
-  to the user tier (only the configured `root_username` keeps root), a
-  duplicated actor header is refused with 400 rather than resolved to either
-  value, and headers without the secret are ignored as before. The server
-  refuses to start with a proxy secret but no `root_username`, because no
-  proxied request could ever reach a root-only capability again. Unset — the
-  default — nothing changes (#333).
-- Qualified identity keys: every ownership decision resolves a request to
-  `sub:<subject>` or `user:<name>` via one accessor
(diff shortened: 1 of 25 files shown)


samirhvbr:
**Cross-PR: this and #310 define contradictory ownership keys**

I reviewed this alongside #310 (I'm not the author of either; I've been contributing the retrieval/TTL work that just landed). The two PRs are individually consistent and individually well-tested, which is exactly why this is easy to miss — nothing in CI compares two open PRs against each other.

**Both define the same ownership vocabulary, in different modules, with inverted precedence:**

| | #310 (`core/src/handoff.rs`) | #333 (`core/src/actor.rs`) |
|---|---|---|
| `ActorContext::identity_key` | `Option<&str>` | `Option<IdentityKey>` |
| precedence | `user` before `sub` | `sub` before `user` |
| persisted value | `alice` | `user:alice` / `sub:alice` |

The test names assert opposite contracts: `identity_key_prefers_user_over_sub` (#310) vs `identity_key_prefers_sub_over_user` (this PR).

**The ordering hazard.** `OwnerFilter::admits` here is an exact string comparison (`me == owner`), and this PR carries no migration. So: merge #310 → any deployment runs it and stamps `owner = "alice"` → merge this PR → the filter now derives `"user:alice"`, which no longer matches the stored `"alice"`.

To be precise about the blast radius: those rows are **not lost**. #310's `all_owners` / `any_owner` recovery switches still reach them across the MCP, admin, web and CLI surfaces. But in normal operation the row silently stops being visible to its own owner, and recovering it requires knowing the switch exists — on a feature whose entire purpose is per-operator isolation.

I don't have an opinion on which precedence is right — that's your call and #310's. Flagging that both exist, that the merge order determines whether a re-key migration is needed, and that this PR's own rationale (`a_username_never_aliases_an_equal_subject`) reads as an argument against #310's raw-string key.

**Separately, on this PR alone:** `IdentityKey`, `OwnerFilter`, `owner_stamp`, `storage_key` and `path_segment` are exported from `core` with no production caller — only tests inside `actor.rs`. `AGENTS.md` is explicit: *"no speculative abstractions or new public surface without a shipped caller"* and *"No dead code or half-built public surface."* They do have callers in #310, so this may just be a sequencing artifact worth stating in the PR body.

The trusted-proxy half looks solid to me: `ct_eq` for the secret, and `asserts_root_identity` comparing the **username** rather than the identity key — which correctly closes the "subject equal to `root_username` acquires root" path.

samirhvbr:
**Correction — I posted this against a stale read of #310.**

#310 was closed and split into #334/#335/#336 about 35 minutes before I commented; I checked its state at the start of my review and not again before posting. That makes the headline of my comment wrong on two counts: there is no merge-order hazard, because #310 will not merge, and the `sub`-over-`user` precedence here is the deliberate resolution of that exact blocker, not a contradiction I found. Your closing comment on #310 says so explicitly, and `docs/users.md` already documents the one residual re-bucket path. Apologies for the noise.

Two points from the original comment do still stand, for what they're worth: `IdentityKey`, `OwnerFilter`, `owner_stamp`, `storage_key` and `path_segment` land here with no production caller — the callers are in #334/#335/#336 — which is worth a line in the PR body given the `AGENTS.md` rule; and the trusted-proxy half reads as solid to me (`ct_eq`, and `asserts_root_identity` keyed on the username rather than the identity key).

djalmajr:
Thanks for the careful read — and for the correction, no noise taken; a
cross-PR comparison is exactly the kind of check CI can't do.

The standing point is accurate, and I've added a sequencing note to the PR
body: `IdentityKey` / `OwnerFilter` / `owner_stamp` / `storage_key` /
`path_segment` are exercised only by their contract tests in this slice —
verified, zero production call sites here; the callers are #334 (owner
columns), #335 (path segments) and #336 (staging buckets). The split was made
so the contract is reviewed once, before its consumers, instead of re-derived
inside each feature PR. If the `AGENTS.md` rule should be read strictly,
#333+#334 merge cleanly as a pair — the contract plus its first consumer.

On precedence: right — `sub` over `user` here is the deliberate resolution of
the identity-key blocker from the #310 audit, not a divergence from it. The
one residual re-bucket path (username-only ingress that later starts
forwarding `sub`) is documented in `docs/users.md` with the "forward `sub`
from day one" guidance.

PR #343, opened 2026-08-01
fix(run): resolve Windows harness executables through PATHEXT

mobnix:
## What changed

On Windows, `run` reported a harness as installed and then failed to start it.

npm-style installs drop three files side by side — `opencode`, `opencode.cmd`, `opencode.ps1`. The availability probe accepted the extension-less shell script, which exists for Git Bash and which `CreateProcess` cannot execute, so the check passed and the launch died with `program not found`.

Availability now resolves to a concrete file: only `PATHEXT` matches count, or a path whose extension was given explicitly. The spawn uses that resolved path instead of the bare name, so the check and the launch can no longer disagree.

Unix behaviour is unchanged — `resolve_candidate` returns the file itself.

## Why

Reproduced on Windows 11 with OpenCode installed through npm. `ai-memory run opencode` passed `ensure_executable_available` because `%LOCALAPPDATA%\...\opencode` (no extension) exists, then `CreateProcess` refused it. The user sees a harness advertised as available and a launch that cannot happen.

## Security side effect

Spawning the resolved path also closes a binary-planting vector. `CreateProcess` searches the application directory and the **working directory before `PATH`**, and the launcher runs the child with `current_dir` set to the checkout. A `claude.exe` or `codex.exe` dropped into an untrusted repository would therefore run instead of the real harness. Resolution now walks `PATH` entries only, so the working directory is never consulted.

This is a real change in behaviour on Windows, so it is called out explicitly rather than buried in the fix.

## Test plan

- [x] Manual test: on Windows 11, `ai-memory run opencode` now starts the harness; before the change it failed with `program not found` while `ensure_executable_available` reported success.

New regression tests:

- `windows_resolution_skips_the_extension_less_shim` (Windows only) — writes an extension-less `faux-harness` beside `faux-harness.cmd` in a temp dir and asserts the resolution returns the `.cmd`, and returns `None` when only the extension-less file exists.
- `unix_resolution_returns_the_file_itself` (non-Windows) — the file is the answer, no `PATHEXT` involved.
- `executable_available_accepts_an_existing_absolute_path` — the branch `--executable` relies on.
- `executable_available_rejects_a_program_that_is_not_installed`.

## Commit attribution

## CHANGELOG (merge gate)

- [x] Added a `[Unreleased] / Fixed` entry.
- [x] The changed behaviour on Windows (spawning a resolved path rather than a bare name) is called out above.

## Notes for reviewers

`(#NNN)` in the CHANGELOG entry is a placeholder — happy to push the real PR number, or fold it in on merge, whichever you prefer.

Two follow-up PRs build on this branch: the `ai-memory show` picker (which reuses this resolution to filter its harness menu) and Antigravity CLI managed-harness support. They target `main` as well, so their diffs include this commit until it merges.


akitaonrails:
Maintainer audit adjustment: the Windows npm-shim bug is valid, but the original commit's binary-planting explanation was not. Rust removed current-directory executable lookup from `std::process::Command` on Windows in Rust 1.58, so this PR should not claim to close that vector. The merge commit preserves the contributor commit without rewriting history, corrects the changelog to describe only the verified PATHEXT behavior, and extends the Windows regression to actually spawn the resolved `.cmd` wrapper.

Local final-head gates passed: formatting, whitespace, full workspace tests, clippy with warnings denied, and `cargo deny check`. Windows CI remains the required platform proof for the cfg-gated spawn regression.

PR #344, opened 2026-08-01
feat(cli): add `ai-memory show` project and harness picker

mobnix:
Implements #342.

## What changed

A new `ai-memory show` subcommand: pick a project, pick a harness, and it enters that directory and delegates to `ai-memory run`.

Project scope is resolved from the working directory, so opening an agent from a parent directory holding many checkouts resolves all of them to that parent's basename and piles unrelated sessions into one scope. `show` inverts the order so the scope is settled before the agent starts. The rationale is in #342.

```
  Project
  > + New project              create the directory and its agent context files
    default/ai-memory          38 minutes ago  ·  2 pages
    default/simulai            57 minutes ago  ·  5 pages
    personal                   not tracked yet
  ↑/↓ move · pgup/pgdn page · enter select · esc cancel   [1/26]
```

**Two sources.** Tracked projects come from `/api/v1/projects`, newest activity first, with page counts. A depth-1 scan of the current directory then adds anything carrying a project marker, so the picker works on a machine where ai-memory has never run. Dependency/build and dot-directories are skipped; the cost is one `exists()` per marker per candidate, never a walk. `--no-scan` disables it.

**Ordering.** Both halves are newest-first. Scanned entries use directory mtime; unreadable timestamps sort last and ties fall back to path order, so the list is stable across runs.

**New-project entry.** Asks for a name, creates the directory, pins both scope halves in `.ai-memory.toml`, and delegates to `install-instructions` for the routing block and managed Agent Skills — `CLAUDE.md` for Claude Code, `AGENTS.md` otherwise. Both marker keys are pinned because the defaults follow the directory name, so a rename or a subdirectory `cd` would otherwise fork the memory into a second, empty project.

**Scope by source.** A tracked pick pins `--workspace` and `--project`, so a marker in the target tree cannot retarget a launch the listing already resolved. A scanned or created pick passes neither, so `run` derives the scope from the checkout exactly as the lifecycle hooks would.

**Harness menu** is filtered to what resolves in `PATH`, reusing the lookup `run` enforces at launch. A failed listing degrades to scan-only with a warning. With stdin or stdout redirected, the same candidates print as tab-separated lines instead of drawing a menu.

## Changed default / response shape

`ProjectSummary` gains `repo_path`, so `GET /api/v1/projects` rows now carry the absolute checkout path. The picker needs it — a row with no checkout path has no directory to enter.

**This is worth a decision from you.** That route is public and read-only, and `docs/https-via-proxy.md` contemplates exposing `/web` behind a proxy, so the field discloses host filesystem paths (including the OS username) to anyone who can reach the API. `/admin/projects` already serves the same `ProjectSummary` under a root-only gate in multi-user mode. I kept it on the public route because that is where the CLI's other read-only listings come from, but I am happy to move the picker to the admin route, or split a DTO so `repo_path` only appears there. Say which you prefer and I will push it.

## Security

The name typed into the new-project prompt is validated against the server's own `^[a-z0-9][a-z0-9._-]*$` rule and against the parent directory before anything is written, so an invalid scope is corrected by typing rather than surfacing later as a hook warning on a session that already started. It also keeps `..`, `/` and `\` out of the joined path.

During development the `--workspace` argument reached the `.ai-memory.toml` template unvalidated while the prompted name was checked. Since both land inside a quoted TOML string, a value like `default"\nproject = "hijacked` could close the quote and append its own key, silently retargeting the project's memory — and marker files are often committed, so a forged value would propagate. Both halves are now validated at the same boundary, before `create_dir`, so a rejected scope leaves nothing behind. Regression test: `workspace_names_are_validated_before_reaching_the_marker`, written before the fix and confirmed failing against the unvalidated version.

Directory creation uses `create_dir`, not `create_dir_all`, so an existing directory is never adopted — it may hold an unrelated project whose marker the creation would replace.

## Test plan

- [x] `cargo deny check` passes (advisories, bans, licenses, sources)
- [x] Manual test: ran the picker from a directory holding 23 checkouts on Windows 11; created a project through the new-project entry and confirmed `.ai-memory.toml`, `CLAUDE.md` and `.claude/skills` land in it and that the launched agent resolves to that project rather than the parent directory. Also ran it with output redirected and confirmed the tab-separated listing.

25 unit tests cover the picker: scope-name validation, the rejection messages, marker contents, directory creation and refusal over an existing directory, harness-to-instruction-file mapping, scan inclusion/exclusion rules, recency ordering and its stable tiebreak, path normalisation for dedup, viewport scrolling maths, and truncation width.

Two notes on how they are written: `sort_by_recency` takes the timestamp provider as a parameter so ordering is tested against fixed values instead of `sleep`-ing between `mkdir`s, and the terminal-drawing helpers are tested through their pure parts (`scroll_offset`, `truncate`) since the raw-mode loop itself needs a TTY.

## Commit attribution

- [x] Verified on every commit in this PR.

## CHANGELOG (merge gate)

- [x] `[Unreleased] / Added` entry included.
- [x] The `/api/v1/projects` response-shape change is called out above.

## Notes for reviewers

**Stacked on #343** (the Windows `PATHEXT` fix), whose commit appears in this diff until that merges — `show` filters its harness menu with the resolution introduced there.

Two things I would like your call on, beyond the `repo_path` question above:

1. **New dependency.** `crossterm` is added for raw-mode key reading. The CLI had no terminal crate before; `anstyle` arrives transitively through clap but only styles text. `run` already has an interactive picker in `choose_native_session` using a numbered prompt on stderr with injectable `input`/`output` — a very different style, and testable in a way an arrow-key menu is not. I went with arrow keys for a 25-item list, but reusing the numbered prompt and dropping the dependency is a reasonable call and I will make the change if you prefer it. If `crossterm` stays, `0.29` would avoid a duplicate `rustix 0.38` alongside the workspace's `1.x` (`cargo deny` does not flag it today).

2. **`(#NNN)` placeholders** in the CHANGELOG entry — tell me whether to push the real numbers or fold them in on merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,44 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `ai-memory show`: an interactive launcher that lists projects, asks which
+  harness to start, enters that project's directory, and delegates to
+  `ai-memory run`. It exists because project scope is derived from the working
+  directory, so opening an agent from a parent directory holding many checkouts
+  resolved all of them into that parent's basename and piled unrelated sessions
+  into one scope. The menu merges two sources: projects the server already
+  tracks (newest activity first, with page counts, from a new `repo_path` field
+  on `/api/v1/projects` rows), and a depth-1 scan of the current directory for
+  directories carrying a recognised project marker — so the picker is useful on
+  a machine where ai-memory has not run yet, instead of staying empty until
+  every project has been opened the long way once. Dependency, build, and
+  dot-directories are skipped, and the scan is disabled with `--no-scan`. A
+  scanned directories are ordered by mtime, newest first, matching the tracked
+  half's activity ordering — alphabetical order buried a just-created project
+  under every checkout whose name sorted earlier. A
+  tracked pick pins both scope halves; a scanned pick passes neither, so `run`
+  derives the scope from the checkout exactly as the lifecycle hooks would. The
+  The list always leads with a **new-project** entry that asks for a name,
+  creates the directory, pins both scope halves in a `.ai-memory.toml` marker,
+  and installs the routing block plus the managed Agent Skills into the
+  instruction file the chosen harness actually reads (`CLAUDE.md` for Claude
+  Code, `AGENTS.md` otherwise) by delegating to `install-instructions`. The name
+  is validated against the server's own `^[a-z0-9][a-z0-9._-]*$` rule and
+  against the directory before anything is written, so a rejection is corrected
+  by typing instead of surfacing later as a hook warning on a session that
+  already started. Both marker keys are pinned because the defaults follow the
+  directory name, so a rename or a subdirectory `cd` would otherwise fork the
+  memory into a second, empty project. The
+  harness menu is filtered to the harnesses actually present in `PATH`, reusing
+  the same lookup `run` enforces at launch, so the picker cannot offer an agent
+  that would fail to start once a project has already been chosen. A failed
+  listing degrades to scan-only results with a warning rather than aborting. With stdin or stdout redirected it prints the same candidates as
+  tab-separated `label`/`path`/`detail` lines instead of drawing a menu, so the
+  list stays greppable. `--workspace` narrows the tracked list, and `--yolo`,
+  `--fresh`, and trailing native arguments are forwarded to `run` unchanged.
+  (#NNN)
+
 ### Fixed
 - `run` on Windows no longer reports a harness as installed and then fails to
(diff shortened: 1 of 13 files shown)


akitaonrails:
Superseded by #347, which preserves Vivaldo Junior’s authored feature commit (`ef74497`) and adds the maintainer audit/hardening commit. The replacement keeps checkout paths client-local rather than exposing server-host `repo_path` through the public projects API, removes process-global cwd mutation, adds an explicit JSON contract, validates/locks the local registry, stages new projects atomically, and routes both `run` and `show` through the host-native Docker wrapper. All Linux, macOS, Windows, release-build, Docker, dependency, secret-scan, and CodeQL checks passed before merge.

PR #345, opened 2026-08-01
feat(workstream): manage Antigravity CLI as a launcher harness

mobnix:
## What changed

`ai-memory run antigravity` (aliases `antigravity-cli`, `agy`) joins an Antigravity CLI session to the same workstream as the Claude Code or Codex sessions on the same checkout. Antigravity already had MCP config and lifecycle hooks; it was missing from the managed-launcher side.

**Launch planning.** `agy` accepts no caller-chosen id for a new conversation — same shape as Kimi — so a fresh launch injects nothing and the id is linked by the hooks or discovered afterwards. A linked resume passes `--conversation <id>`. `--continue` / `-c` resumes the most recent conversation without naming one; that is still an explicit user choice, so nothing is injected over it. `--yolo` maps to `--dangerously-skip-permissions`. `--print` / `-p` / `--prompt` block adoption; `-i` (`--prompt-interactive`) does not, since it seeds a prompt and keeps the session open. Utility subcommands (`models`, `plugin`, `update`, `agents`, `install`, `changelog`, `help`) pass through without a selector.

All of the above was read off `agy --help` rather than assumed.

**Conversation discovery.** Antigravity keeps one SQLite database per conversation at `~/.gemini/antigravity-cli/conversations/<conversation-id>.db`, so the id is the file name and locating one needs no scan. The workspace a conversation was opened on comes from `trajectory_metadata_blob`: a protobuf message whose field 1 holds a nested message whose field 1 is the workspace `file://` URI. Only those two fields are read.

The field walk is a small bounds-checked reader — every access goes through `get()` / `checked_add()`, the varint is capped at 10 bytes, unknown wire types end the walk instead of guessing a length, and the cursor always advances. A database that does not carry those fields, or is not a database at all, is skipped rather than failing the listing, so a store written by a different `agy` version cannot take the whole picker down.

Verified against five real conversation databases: the walk recovered the correct workspace path from each, including two different projects and a home-directory session.

**What is deliberately not implemented.** Step payloads are undocumented, unversioned protobuf whose step-type enum (values 5, 8, 14, 15, 21, 23, 90, 91, 98, 101, 132 in one sample) is not published. Decoding conversation text would mean guessing at a schema that changes between releases, so `export_transcript` returns a clear error for this harness and the visible-event ledger comes from lifecycle-hook capture instead. I would rather ship the half that is stable than a parser that silently rots.

Antigravity is not added to the no-argument auto-detection pool — same treatment as OMP and Grok.

**Server-side allowlist.** `prepare_run` validates the agent kind independently of the client, so `AgentKind::AntigravityCli` had to be accepted there too. Without it the launch is rejected with a 400 after the user has already chosen a project and a harness. The direct-managed-context gate is deliberately left alone: that path exists for Crush and Grok, which cannot receive context through hooks, while Antigravity injects via `PreInvocation`.

## Why

Antigravity sessions could read and write memory through MCP and hooks, but stayed outside managed workstreams, so switching to it from Claude Code started a disconnected line of work.

## Test plan

- [x] Manual test: `ai-memory run antigravity` from a checkout with an existing workstream, on Windows 11 with `agy` installed; verified the launch joins the workstream and that the 400 above reproduced before the allowlist arm was added.

New tests:

- `antigravity_names_parse_to_one_variant`, `antigravity_fresh_launch_injects_no_selector`, `antigravity_resume_appends_conversation_selector_after_user_arguments`, `antigravity_explicit_selector_always_wins`, `antigravity_utility_subcommands_are_passed_through`, `antigravity_print_blocks_adoption_but_interactive_prompt_does_not`, plus the harness added to the existing `--yolo` mapping table.
- `antigravity_lists_only_conversations_from_this_workspace` — builds real SQLite databases with the two-layer protobuf metadata and asserts only the current workspace's conversation is offered.
- `antigravity_skips_conversations_without_readable_metadata` — a database with no metadata table and a file that is not a database at all; the listing is empty, not an error.
- `antigravity_transcript_export_explains_why_it_is_unavailable`.
- `file_uris_decode_on_both_platform_shapes` — Windows drive form, POSIX form, percent-decoding, and a stray `%` that is not an escape.
- `protobuf_walk_skips_other_fields_and_wire_types` — including truncated input.
- `antigravity_is_accepted_explicitly_but_not_in_the_automatic_pool` — the server-side allowlist, mirrored on the existing Grok test.

## Commit attribution

- [x] Verified on every commit in this PR.

## CHANGELOG (merge gate)

- [x] `[Unreleased] / Added` entry included.

## Notes for reviewers

**Stacked on #343 and #344**, whose commits appear in this diff until they merge. The Antigravity work itself is independent of both — it touches `ai-memory-workstream`, `ai-memory-hooks`, and the harness enum in the CLI. Happy to rebase it directly onto `main` if you would rather review it first.

The contract was verified against the `agy` build currently shipping on Windows. If you want a version number recorded in `docs/managed-workstreams.md` the way Grok's is, tell me which field you consider authoritative — `agy --help` does not print one and `agy changelog` output is not stable to parse.

`(#NNN)` in the CHANGELOG is a placeholder, same question as the other two PRs.


akitaonrails:
Superseded by #346, which preserves Vivaldo Junior’s authored Antigravity commit while cleanly excluding the unrelated stacked #344 changes. The replacement was audited, hardened, documented, tested with deterministic and real Antigravity acceptance coverage, and merged with full hosted checks green. Closing this original PR without rewriting its history.

Issue #212, opened 2026-07-21
Erro Permission denied (os error 13) at path


Marcelo7889:
Ola Pessoal, eu estava com este erro ao rodar o comando ai-memory install-mcp   --client claude-code --apply
aqui no meu OpenSuse Tumbleweed, ele sempre dava:

marcelo@dev-station ~/.l/bin> ai-memory install-mcp   --client claude-code --apply
2026-07-21T18:26:40.213705Z  INFO ai_memory: ai-memory starting version="1.17.1" server_url=http://127.0.0.1:49374 data_dir=/data bind=127.0.0.1:49374
Error: creating tempfile next to /home/marcelo/.claude.json

Caused by:
    Permission denied (os error 13) at path "/home/marcelo/.ai-memory-apply-tmp.xmHKq9"

Não sei se é alguma configuração errada do meu Docker, ou bloqueio do Selinux da propria distro,  mas com a ajuda do claude consegui identificar e resolver, removendo a imagem e adicionando uma flag na criação da nova, rodando o comando docker da seguinte forma:

docker run -d --name ai-memory --security-opt label=disable \
                                      --restart unless-stopped \
                                      -p 127.0.0.1:49374:49374 \
                                      -v ai-memory-data:/data \
                                      -e AI_MEMORY_LLM_PROVIDER=anthropic \
                                      -e ANTHROPIC_API_KEY=sk-ant-... \
                                      -e AI_MEMORY_EMBEDDING_PROVIDER=openai \
                                      -e OPENAI_API_KEY=sk-... \
                                      akitaonrails/ai-memory:latest

Caso seja algum bug da aplicação, pedi ao claude para gerar um pequeno report aqui para ajudar a identificar, segue abaixo:

Version
Output of ai-memory --version:

ai-memory 1.17.1

Wrapper: bin/ai-memory from main, image akitaonrails/ai-memory:latest (sha256:d44b5172cff36adb697959221034a38e0f77ac504e9eb9ce887a8aa396b2fafc)

What happened

On an SELinux-enforcing distro (openSUSE Tumbleweed), every wrapper subcommand that writes host files fails with Permission denied:

$ ai-memory install-mcp --client claude-code --apply
2026-07-21T18:26:40.213705Z  INFO ai_memory: ai-memory starting version="1.17.1" server_url=http://127.0.0.1:49374 data_dir=/data bind=127.0.0.1:49374
Error: creating tempfile next to /home/marcelo/.claude.json

Caused by:
[...]

PR #388, opened 2026-08-09
fix(wrapper): detect rootless and SELinux under podman

rodrigopalhares:
## What changed

Before: on podman-based hosts every `bin/ai-memory` command that touches host files failed with `Permission denied (os error 13)`. After: those commands get the same `-u 0:0` remap and `--security-opt label=disable` they already got under rootless Docker.

Two independent fixes:

- **Engine detection.** Both gates were decided from `docker info --format '{{.SecurityOptions}}'`, a Docker-only field. Podman — including through the `podman-docker` `docker` shim — cannot evaluate it: the template fails and the command exits 125. The wrapper's `|| true` swallowed that into an empty string, so both gates read as "rootful, no SELinux" exactly where they were needed. Podman's `.Host.Security.Rootless` and `.Host.Security.SELinuxEnabled` are now consulted when the Docker probe comes back empty.
- **`bootstrap` coverage.** The host-file list only covered writers. `bootstrap` only *reads* — the repo bind-mounted at `/work` and the `.ai-memory.toml` markers under `$HOME` — but an unmapped UID and a confined label block reads just as hard. Its failure was two-stage and misleading: it degraded silently to `no .git found at /work; bootstrapping from README/docs/rules only`, then died with `Permission denied`. `WRITES_HOST_FILES` is now `TOUCHES_HOST_FILES` (readers ∪ writers).

**No changed defaults.** Docker hosts are unaffected: the podman probe only runs when the Docker probe returns nothing. Thin-client commands (`status`, `search`, …) stay confined — they only touch the named volume, which is not host-visible.

## Why

Bug fix. Fedora, RHEL and openSUSE ship rootless podman with SELinux enforcing by default, and the `podman-docker` package installs a `docker` shim — so the wrapper believes it is talking to Docker. That makes every host-file command (`install-mcp`, `install-hooks`, `setup-agent`, `install-instructions`, `install-skills`, `uninstall`, `backup`, `bootstrap`) fail out of the box on that family of distros.

**This is an addition to the two earlier wrapper fixes, not a replacement:**

- **#150** (`fix/rootless-docker-uid-remap`) introduced the `-u 0:0` remap for rootless Docker; **#162** later lifted it into `ROOTLESS_DOCKER` and preserved it on the macOS arm.
- **#218** (issue **#212**, reported on openSUSE Tumbleweed) added `--security-opt label=disable` for SELinux-enforcing hosts.

Both were correct for the engine they were reported against, and both read the same Docker-only `docker info` field — so neither ever fired on podman. This PR leaves their logic untouched and only teaches the probe a second dialect, plus corrects one subcommand classification (see notes below).

The wrapper has documented `AI_MEMORY_DOCKER=podman` since it was introduced; the security probes added later assumed Docker.

No issue was filed for this one — it was found while running the wrapper on Fedora.

## Test plan

- [x] Manual test: on Fedora with rootless podman and SELinux enforcing (`getenforce` → `Enforcing`; `podman info` → `Host.Security.Rootless=true`, `Host.Security.SELinuxEnabled=true`; `/usr/bin/docker` provided by `podman-docker`). Confirmed `docker info --format '{{.SecurityOptions}}'` exits 125 with `can't evaluate field SecurityOptions in type system.infoReport`. Then ran the wrapper against the live engine using a shim that forwards `info` to podman and captures the `docker run` argv instead of starting a container:
      - pre-branch wrapper: all 18 subcommands probed get `-u 1000:1000` and no `--security-opt` — no gate fires at all under podman
      - this branch: all 8 host-file subcommands (`install-mcp`, `install-hooks`, `setup-agent`, `install-instructions`, `install-skills`, `uninstall`, `backup`, `bootstrap`) get `-u 0:0` and `--security-opt label=disable`
      - this branch: every thin client probed (`status`, `search`, `read-page`, `write-page`, `lint`, `embed`, `reindex`, `init`, `completions`) keeps `-u 1000:1000` and no label, so the exception stays scoped

      `cargo test -p ai-memory-cli --test packaging` → 34 passed, 0 failed.

## Commit attribution

## CHANGELOG (merge gate)

- [x] I added a `CHANGELOG.md` `[Unreleased]` entry — **required** for any user-facing change: new flag / env var / endpoint / MCP tool / marker key, changed behaviour, or an observable bug fix. (Exempt only for internal refactors, dead-code removal, and test-only churn.)
- [x] Any changed **default** (flag behaviour, config default, env var, response shape) is called out explicitly in "What changed" above.

Reviewers treat a missing entry as blocking — adding it up front is what keeps your PR merging on the first pass.

## Notes for reviewers

- The podman path is deliberately a **fallback**, not a branch on engine name. The wrapper can never know which engine it is talking to: `podman-docker` makes `$DOCKER` literally `docker`. So it keys off the Docker probe returning nothing, which is true for podman however it is reached — through the shim or via `AI_MEMORY_DOCKER=podman`.
- The `bootstrap` half is a direct correction to a classification made in #150, which put it on the safe side of the line: *"Thin-client commands (status, bootstrap, search, ...) are untouched since they only touch the /data named volume, which isn't host-visible."* That holds for `status` and `search`, but `bootstrap` reads the repo bind-mounted at `/work`. The classification, not the mechanism, was the bug — which is why it survived both earlier fixes.
- The substantive test change is `FakeInfo` in `packaging.rs`: the fake `docker` can now impersonate either engine, reproducing podman's exact template error and exit 125. Every previous fixture spoke Docker, which is why a green suite
  — including `selinux_label_exception_requires_enforcement_and_daemon_support`
  — could not catch this. No test had ever asked what happens when the probe answers *nothing*.
- The fallback costs two extra `docker info` calls (three total on podman). One combined template, `'{{.Host.Security.Rootless}} {{.Host.Security.SELinuxEnabled}}'`, returns `true true` on a live podman and would cut that to two. Happy to fold it in if preferred — left out to keep this diff scoped to the bug.
- Possible follow-up, not in this PR: `restore --from <tarball>` reads a host path but is in neither list. Probing it on the same podman host shows it still gets `-u 1000:1000` and no label — exactly the shape `bootstrap` had before this PR. I stopped at the flags and did not drive it to an actual failure, so treat it as a strong suspicion rather than a reproduced bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- The `bin/ai-memory` wrapper now detects rootless mode and SELinux under
+  podman, so host-file commands stop failing with `Permission denied (os error
+  13)` on podman-based distros. Both the `-u 0:0` remap and `--security-opt
+  label=disable` were decided from `docker info --format
+  '{{.SecurityOptions}}'`, a Docker-only field that podman cannot evaluate;
+  the swallowed error left both gates off exactly where they were needed.
+  Podman's `.Host.Security.*` keys are consulted when that probe comes back
+  empty. `bootstrap` is now covered as well: it only reads host files, but an
+  unmapped UID blocks reads just as hard, and it degraded silently to "no
+  `.git` found" before dying (#385).
+
 ## [1.25.0] - 2026-08-07
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Podman's `.Host.Security.*` keys are consulted when that probe comes back
   empty. `bootstrap` is now covered as well: it only reads host files, but an
   unmapped UID blocks reads just as hard, and it degraded silently to "no
-  `.git` found" before dying (#388).
+  `.git` found" before dying. Restore archives, explicit config paths, and
+  commands using a host-backed `AI_MEMORY_DATA_DIR` now receive the same
+  host-file treatment (#388).
 
 ## [1.25.0] - 2026-08-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Podman's `.Host.Security.*` keys are consulted when that probe comes back
   empty. `bootstrap` is now covered as well: it only reads host files, but an
   unmapped UID blocks reads just as hard, and it degraded silently to "no
-  `.git` found" before dying (#385).
+  `.git` found" before dying (#388).
 
 ## [1.25.0] - 2026-08-07
 
--- a/README.md
+++ b/README.md
@@ -578,7 +578,9 @@ one matching entry.
   or relabel `$HOME`; do not add `:z`/`:Z` to the whole home bind. Rootless
   engines also get `-u 0:0` for those commands. Docker and podman report
   rootless mode and SELinux support under different `info` keys; both are
-  read. See [`docs/install.md`](docs/install.md#selinux-enforcing-hosts).
+  read. The same treatment applies whenever `AI_MEMORY_DATA_DIR` selects a
+  host directory or an explicit `--config` reads a host file. See
+  [`docs/install.md`](docs/install.md#selinux-enforcing-hosts).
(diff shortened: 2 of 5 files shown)


akitaonrails:
Maintainer follow-up added before merge:

- applied the rootless UID and SELinux treatment to `restore` host archive reads
- covered explicit global `--config` paths
- covered every command when `AI_MEMORY_DATA_DIR` is a host bind
- aligned the macOS rootless branch with the same host-touch decision
- added packaging regressions and corrected README/install/changelog claims

Focused packaging tests pass locally. All gating hosted checks pass on `b7c17c1f5c6d1e667b4ad54ebc783a8a562af356`; only the explicitly non-gating Windows suite is still running.

Issue #394, opened 2026-08-15
Mid-session events from non-git directories create phantom projects; add session-inheritance fallback (+ opt-in sticky routing)


aguirreSL:
**Problem this would solve**

With hooks installed using `--project-strategy repo-root`, per-event project resolution works well inside a repository. But when a mid-session command's cwd is **outside any git repo and has no `.ai-memory.toml` marker**, resolution falls through to `basename(cwd)`, even though the event carries a `session_id` whose session already has a resolved project.

The most common trigger is the agent harness scratch directory. Claude Code gives each session a scratch dir at `/private/tmp/claude-<uid>/<sanitized-project-path>/<session-id>/scratchpad`. Any Bash command that `cd`s there produces observations routed to a phantom project literally named `scratchpad` (or `data`, `tmp`, ... for its subdirectories).

Observed on 1.24.0 and still present in 1.26.1: 463 observations from 11 sessions across 6 real projects accumulated in a `scratchpad` project over three weeks, plus smaller phantom buckets (`data`, `tmp`, `clone_teste`). Recovering them required direct SQL with the server stopped (`UPDATE observations SET project_id = (SELECT project_id FROM sessions WHERE id = session_id) WHERE project_id = <bucket>`), since `rename-project` refuses an existing target and `reorg` operates on whole sessions.

A sibling of the same problem: a mid-session `cd` into a *different git repository* silently routes those observations to the other repo's project, splitting one session's raw record across two projects, while the session row and its consolidated page stay in one (`sessions.project_id` is single-valued, and consolidation already reads by `session_id` and writes one page in the session's project).

**Proposed solution**

Extend the fallback chain for **mid-session** events (session-start routing unchanged):

1. `.ai-memory.toml` marker (unchanged)
2. git repo root under `repo-root` strategy (unchanged)
3. **new:** if neither resolves and the event's `session_id` matches an existing session, inherit that session's project
4. `basename(cwd)` (unchanged; still applies to session-start events, so opening a session in a plain non-git folder still names the project after the folder)

This keeps every current behavior for real projects (git or not) and only changes the case that today produces phantom buckets.

Optionally, a config knob for the cross-repo case:

```toml
[routing]
mid_session = "follow-cwd"   # today's behavior, default
# mid_session = "sticky"     # all mid-session events inherit the session's project
```

[...]

Issue #103, opened 2026-06-16
Unscoped queries resolve to a catch-all project: the cwd->project resolver stores the bare cwd as repo_path


rthiago:
> **Correction (edited).** My original report blamed the HTTP transport and the single global "active project" pointer (`mode=Single`). After digging into the code that turned out to be a *symptom*, not the cause. The real bug is in the cwd->project resolver and is transport-independent. I've rewritten the issue accordingly; the original symptom evidence (logs, `global=true` comparison) still holds and is summarized under "Original symptoms" below.

**Version:** `ai-memory 1.1.0` (resolver code is unchanged on `main` as of this edit).

## Summary

Opening an agent session in a directory that is an **ancestor of your projects** (most commonly your home dir, e.g. `cd ~ && claude`) permanently creates a project row whose `repo_path` is that ancestor directory. From then on, the longest-prefix cwd->project match resolves **every** project nested beneath it to that one catch-all project. Unscoped `memory_query` / `memory_read_page` therefore return the catch-all project's data (a jumble of every session's observations) instead of the calling session's project. That is exactly the "wrong project / empty hits for projectA, but `global=true` returns projectA correctly" behavior I originally reported.

## Root cause

1. `derive_project_from_cwd` in `crates/ai-memory-hooks/src/router.rs` (v1.1.0: ~L458-480) computes a `repo_path` for the session's cwd, and when there is no git root it falls back to the **bare cwd**:
   ```rust
   let repo_path = root.map_or_else(|| cwd.to_string(), |p| p.to_string_lossy().into_owned());
   (name, Some(repo_path))   // always Some
   ```
   `root` is `Some` only under the `MainRepoRoot` strategy *inside* a git repo. The default strategy is `Basename`, where `root` is always `None`, so `repo_path` is **always** the bare cwd, including for non-repo directories like `$HOME`.

2. `get_or_create_project` in `crates/ai-memory-store/src/ops.rs` writes `repo_path` only on INSERT and never updates it, so that first bare-cwd value is sticky for the life of the row.

[...]

PR #398, opened 2026-08-16
feat(routing): opt-in sticky mid-session attribution (#394)

akitaonrails:
Stacked on #397 — review that first; the base will retarget to `main` once it merges.

## What changed

Adds `[routing] mid_session`, the knob #394 left open after #395 shipped the out-of-tree session-inheritance fallback.

- **`follow-cwd`** (default) is unchanged: every mid-session event re-resolves from its own cwd.
- **`sticky`** keeps the session's project wherever the agent wanders, closing the cross-repo `cd` case that split one session's raw record across two projects while its session row and compiled page stayed in the first.

## The wire-level part

The knob has to overrule a host-derived project name while still honoring a marker-declared one — and both arrive as the same `project` query parameter, which is exactly why #395 excluded it. Hooks now tag which they sent:

| `project_src` | meaning | may yield to the session? |
|---|---|---|
| `marker` | a `.ai-memory.toml` named it — deliberate rescope | no, in either mode |
| `repo-root` | the host hook derived it from the enclosing checkout | only under `sticky` |
| absent | client older than this release | no — stays authoritative |

So `sticky` degrades safely on a mixed-version fleet instead of silently capturing deliberate rescopes. Tagged by all four emitters: the POSIX hook lib, the PowerShell lib, the native `hook` command, and both OpenClaw plugin variants.

The broad-anchor guard is deliberately **not** relaxed under `sticky`: a session rooted at `/` or `$HOME` still never sticks, or one stray session would fold every project beneath it into a single bucket (#103).

## Note for reviewers

`[routing]` sits **above** `[auth]` in the config template because `ai-memory init` appends bare `[auth]` keys (`token_pepper`, …) to the end of the rendered file — a section added below silently captures them. The template now says so, and the init test asserts the structural rule rather than relying on whichever key happens to be checked.

## Test plan

- [x] `TAILWIND_SKIP=1 cargo test --workspace` — 2390 tests
- [x] `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings`
- [x] `tests/hooks/test_lib.sh` — 44 passed
- [x] Verified on a live server: config renders correctly and `token_pepper` lands under `[auth]`

Tests include two table-driven gate tests and four end-to-end router tests: sticky keeps the session project across a sibling checkout; follow-cwd still splits; a marker still rescopes under sticky; an untagged old-client override stays authoritative.

## CHANGELOG

- [x] `[Unreleased]` → `Added`, referencing #394.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added `[routing] mid_session` to choose how a mid-session event is attributed
+  once the agent's cwd has moved. `follow-cwd` (default) keeps the historical
+  per-event resolution; `sticky` keeps the session's project wherever the agent
+  wanders, closing the cross-repo `cd` case that split one session's raw record
+  across two projects. Lifecycle hooks now tag each `project` override with its
+  provenance (`project_src=marker` or `project_src=repo-root`), so `sticky`
+  overrules a host-derived repo name while a `.ai-memory.toml` marker still
+  wins in both modes. Clients older than this release send no provenance and
+  keep their overrides authoritative (#394).
 - Added `[auth].secure_cookie` for HTTPS reverse-proxy deployments to mark
   `/web` browser authentication cookies `Secure` while preserving plain-HTTP
   loopback compatibility by default (#396).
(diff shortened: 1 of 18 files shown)


akitaonrails:
Superseded by #400 — same branch and commit, retargeted to `main` after this PR's base branch was deleted on merge of #397.

PR #399, opened 2026-08-16
feat(purge): strong per-session deletion (#387)

akitaonrails:
**Draft — not for 1.27.0.** Held back pending a production-scale test; see "Before this merges" below. Stacked on #398.

## What changed

Implements #387: `ai-memory purge-session` and `POST /admin/purge-session`, so an application that promises "forget this conversation" can keep that promise. Five commits, following the staging the issue itself asked for.

**Stage A — core + provenance.** Migration V50 adds a content-free tombstone ledger and FK-free `source_session_id` columns. Everything that identifies derived content (`auto_improve_runs.session_id`, `rejections.source_run_id`/`.source_proposal_id`) is `ON DELETE SET NULL`, so deleting the session first destroys the evidence — targets are collected before anything is cut, by identity only, never by searching page text.

**Stage B — no way back in.** `410 {"code":"session_purged","terminal":true}` on ingest, a `Terminal` drainer outcome that drops the spool entry without spending a retry, a local spool sweep by exact identity, and per-entry session markers in the monthly log.

**Stage C — git strong deletion.** Deleting a file and committing leaves every prior version reachable, and `restore-page` hands one back. The repository is rebuilt without those paths and the old object database physically destroyed. History is preserved (not squashed), blobs shared with surviving pages are spared, verification runs before anything destructive, and an interrupted swap converges from a durable journal at startup.

**Stage D — files, not just tables.** FTS5 deletes are *logical*: they write a delete marker and leave the original terms in the index segments, so `MATCH` returned nothing while the purged text stayed readable in `pages_fts_data`. Plus WAL checkpointing and page reclamation. Restore reapplies the tombstone ledger and fails closed; reindex drops tombstoned pages.

## What the verification found

Three bugs that unit tests structurally could not reach:

1. **FTS5 logical deletes** — every relational assertion passed while the content sat on disk. Caught by searching *files* for a canary rather than querying tables.
2. **Historical versions of the shared monthly log** — the current file was scrubbed, earlier committed versions were not.
3. **Commit messages** — they embed the session's own prompt text and were carried into the rebuilt history verbatim.

The last two were found only by running the purge end-to-end against a live server; the fixtures used generic commit messages and had no shared files.

## Verified live

Canary absent from 0 DB layers, 0 files, and **0 git objects** (all objects, reachable or not) while the sibling session stayed intact in all five layers. `410` on replay without resurrection, `already_purged` on repeat with a stable tombstone, 401/401/400/422 across the authz matrix, no prompt content in any server log. V50 applies cleanly to a genuine v1.26.1 database (49 → 50, all rows preserved), and a legacy pre-V50 session purges via the deterministic path fallback while its provenance-less log correctly refuses to guess until `--force`.

## Before this merges

- [ ] **Production-scale test.** The purge runs *synchronously inside the admin HTTP request* and rebuilds the entire wiki repository commit-by-commit plus `VACUUM`s the database. Fixtures had 3 commits. A production instance with ~2000 sessions and ~5000 page versions is untested — if it exceeds the HTTP timeout, the git rebuild likely needs to move off the request path, which is a design change better made before the API ships.
- [ ] **Remaining E2E gaps**, covered by unit tests but not stitched into the combined flow: LLM-simulated single/multi-page consolidation, auto-improve artifacts and sidecars, spool SHA-256 comparison, tarball extraction and inspection.

## Test plan

- [x] `TAILWIND_SKIP=1 cargo test --workspace` — 2428 tests
- [x] `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings`
- [x] `tests/hooks/test_lib.sh` — 44 passed; companion importer — 11 passed
- [x] Live end-to-end against a running server

## CHANGELOG

- [x] `[Unreleased]` entries under `Added`, referencing #387.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added `ai-memory purge-session` and `POST /admin/purge-session` for strong
+  per-session deletion, so an application that promises "forget this
+  conversation" can keep that promise. Removes the session, its observations
+  and both FTS indexes, handoffs it produced or accepted, queued consolidation
+  and auto-improvement claims, derived auto-improvement runs/proposals/events/
+  rejections, every version of its derived pages, and the wiki files — leaving
+  only a content-free tombstone. Root-only, `--confirm` required, full UUID
+  only, idempotent, and `409` while the session is still in flight unless
+  `--force`. Page provenance is now structural and sticky, so an LLM rewrite or
+  hand edit cannot launder a derived page out of the purge's reach. Managed
+  workstreams are explicitly out of scope: their `native_session_id` has no
+  foreign key to `sessions.id` and is never matched by coincidence (#387).
 - Added `[routing] mid_session` to choose how a mid-session event is attributed
   once the agent's cwd has moved. `follow-cwd` (default) keeps the historical
   per-event resolution; `sticky` keeps the session's project wherever the agent
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   hand edit cannot launder a derived page out of the purge's reach. Managed
   workstreams are explicitly out of scope: their `native_session_id` has no
   foreign key to `sessions.id` and is never matched by coincidence (#387).
+- Hook ingest for a purged session is now refused as terminal with
+  `410 {"code":"session_purged","terminal":true}`, so an event spooled by a
+  client that was offline during the purge cannot resurrect it. The native
+  drainer treats that response as terminal — the entry is dropped without
+  spending a retry attempt — and `purge-session` additionally sweeps the local
+  `hook-spool` by exact session identity. Spools on other hosts cannot be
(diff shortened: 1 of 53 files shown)


akitaonrails:
Closing as superseded by the re-scope in #387.

This draft grew to +5048 lines across 53 files — a tombstone ledger, a git-history rewriter with interrupted-run recovery, and a log scrubber — for a report that asked for one thing: delete a session by UUID so `sessions`, `observations` and `handoffs` stop returning it.

Two measurements shrank it further. The relational side already scrubs (matched-pair test: text greppable in the raw SQLite file before a delete, gone after), so the rebuild/VACUUM machinery here solves a problem that no longer exists. And the git-history rewrite answers a question the reporter never asked — forensic unrecoverability rather than API-level deletion.

It is also 163 commits behind `main`, and the surfaces it touches have moved underneath it (#463 descriptor queries, #467 `PagePath` validation, the writer path). Rebasing would import the cruft along with the useful core.

The part worth keeping is recorded in #387 so it is not lost: targets must be collected **before** anything is cut, because `auto_improve_runs.session_id`, `auto_improve_rejections.source_run_id` and `pages.supersedes` are all `ON DELETE SET NULL` — delete the session first and you destroy the evidence needed to find what it produced.

Not deleting the branch, so the git-purge work is recoverable if forensic deletion is ever scoped as its own issue.

PR #408, opened 2026-08-17
feat(sanitize): complete GitHub/AWS/Stripe token families; add Meta, Google refresh, Telegram, GoHighLevel

dk96-creator:
## What changed

`BUILTIN_PATTERN_STRS` gains seven credential shapes. Three of them **complete
vendor families the sanitizer already covered only in part**:

| Vendor | Covered before | Added |
|---|---|---|
| GitHub | `ghp_`, `github_pat_` | `gho_`, `ghu_`, `ghs_`, `ghr_` |
| AWS | `AKIA…` (long-lived) | `ASIA…` (STS temporary) |
| Stripe | `sk_live_` | `rk_live_` (restricted) |

and four are providers with no prior coverage at all:

- **Google OAuth refresh tokens** (`1//…`) — longer-lived than the `AIza…`
  keys already handled: they mint fresh access tokens until explicitly
  revoked, so a leaked one outlives the session it came from.
- **Meta / Facebook Graph access tokens** (`EAA…`) — ad-account, page and
  business-management scope.
- **Telegram bot tokens** (`<bot-id>:AA…`) — full control of the bot,
  including every message it can see.
- **GoHighLevel Private Integration Tokens** (`pit-…`) — long-lived
  server-to-server credentials that do not expire until manually revoked.

**Behaviour change:** text matching these shapes is now replaced with
`[REDACTED]` where it previously persisted verbatim. An operator who wants
one kept visible (a public bot id, say) can add it to `[sanitize].allowlist`.
No config, flag, endpoint or response shape changes.

The module doc comment's "what we redact" list is updated to match, and
`CHANGELOG.md` has an `[Unreleased]` entry.

## Why

`gho_` is the specific gap that prompted this. It is the prefix `gh auth
login` writes to disk, so it is the GitHub token most likely to appear in a
captured shell excerpt — and it was the one prefix of the family that fell
through. Once that surfaced, the same shape showed up twice more: `ASIA…`
next to `AKIA…`, `rk_live_` next to `sk_live_`. Each is the same vendor and
the same blast radius as a prefix already on the list, so the partial
coverage reads as an oversight rather than a decision.

The three new providers were chosen on longevity rather than popularity: a
Google refresh token and a Meta long-lived token both survive the session
that leaked them, which is exactly the property that makes a durable
markdown wiki the wrong place for them.

This matters more here than the `extra_patterns` escape hatch suggests.
Per `SECURITY.md`, operator `extra_patterns` run **server-side only**, so a
secret matched only by an operator rule can still sit in the excerpt on the
client spool and the wire. Only `BUILTIN_PATTERN_STRS` scrubs client-side —
which is why these belong in the builtin list rather than in everyone's
config.

## Test plan

- [ ] `cargo deny check` passes

⚠️ **I could not run the four gates — there is no Rust toolchain on the
machine I authored this on.** I am not claiming they pass; please treat CI
as the first real run. To compensate I verified what was verifiable without
a compiler, and I would rather state that plainly than tick boxes I did not
execute:

- All 22 builtin patterns (16 existing + 6 new) compile under an independent
  regex engine, and every new pattern matches its fixture.
- **Negative control:** the `EAA…` pattern does *not* match
  `b3BlbnNzaC1rZXktdjEAAAAA` — the base64 OpenSSH fixture already used by
  `scrubs_pem_private_key_block`, which contains the substring `EAAAAA`. The
  `{20,}` tail is what prevents `EAA…` from eating arbitrary base64. This is
  asserted in a new test (`meta_pattern_does_not_eat_short_base64_runs`) so
  the constraint cannot silently regress.
- **Regression check:** the fixtures from the nine existing sanitizer tests
  still redact, and two control strings that should stay clean
  (`just a normal sentence…`, `version 1.28.0 released on 2026-08-17`) do —
  the second specifically guards the new `1//` pattern against version
  strings.
- Line widths in the changed Rust are ≤100 cols with no trailing
  whitespace or tabs, so `cargo fmt` should be a no-op, but that is an
  inspection and not a run.

Nine tests added, matching the existing `scrubs_*` style and the `FAKE…`
synthetic-fixture convention used by the `ghp_FAKE…` fixture in
`ai-memory-wiki`. Every fixture is a shape, never a live value — the
GoHighLevel one uses `DEADBEEF`/`CAFEBABE` in uppercase hex, which also
exercises the case-insensitivity.

Two of the nine are negative controls rather than positive assertions
(`meta_pattern_does_not_eat_short_base64_runs`,
`ghl_pattern_does_not_eat_hyphenated_english`). Both encode a false-positive
boundary that is otherwise invisible and easy to regress.

Incidentally: `sk_live_` had no test before this — `scrubs_stripe_restricted_key`
is the first coverage for that line, via its `rk_live_` sibling.

## Commit attribution

## CHANGELOG (merge gate)

- [x] Added an `[Unreleased]` entry — this is a user-facing behaviour change
      (text that previously persisted is now redacted).
- [x] The behaviour change is called out explicitly in "What changed" above.

## Notes for reviewers

Three judgement calls worth your scrutiny:

0. **GoHighLevel's token tail is not vendor-documented, and I want that on
   the record rather than buried.** The `pit-` prefix *is* documented — their
   MCP guide shows `Authorization: Bearer pit-your-token` — but HighLevel
   publishes no character set or length for what follows, on that page or
   any other I could find. Every PIT observed in practice carries a UUID, so
   the pattern anchors on prefix + UUID shape and accepts either hex case
   (the vendor specifies none). I deliberately rejected the permissive
   alternative `pit-[A-Za-z0-9\-]{20,}`, because `pit-` is an English
   fragment and that form redacts `pit-stop-strategy-analysis` — pinned by
   the `ghl_pattern_does_not_eat_hyphenated_english` negative control. The
   accepted cost: if HighLevel ever issues a non-UUID tail, this misses it.
   If you would rather not carry a pattern whose tail rests on observation
   rather than documentation, dropping that one line leaves the other six
   independently sound.

1. **False-positive appetite on `EAA…`.** The file's own comment says *"False
   positives are acceptable — better to redact a stray hash than to leak a
   credential,"* and I followed it: `EAA[A-Za-z0-9]{20,}` will occasionally
   redact a base64 run that merely starts with those three characters. Real
   Meta tokens run to ~200 chars, so tightening to `{30,}` or `{40,}` would
   cut false positives with little practical loss of coverage. I went with
   `{20,}` to match the stated doctrine, but I would defer to you here.

2. **`1//` is only three characters of prefix.** It is anchored by the
   `{20,}` base64url tail, and I checked it against version strings
   (`1.28.0`) and URL paths. Still, it is the loosest of the six and the one
   I would most expect to surprise someone.

I deliberately left out one vendor I could not verify to primary
documentation. I would rather ship six patterns I have confirmed than seven
with a guess in a security-critical const array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- The built-in sanitizer now redacts six further credential shapes, completing
+- The built-in sanitizer now redacts seven further credential shapes, completing
   three vendor families it already covered only in part. **GitHub:** every
   token prefix — `gho_` (OAuth, the form `gh auth login` writes to disk),
   `ghu_`, `ghs_`, `ghr_` — where previously only `ghp_` and `github_pat_`
@@ -17,8 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   providers with no prior coverage: **Google OAuth refresh tokens** (`1//…`,
   longer-lived than the `AIza…` keys already handled — they mint access
   tokens until revoked), **Meta / Facebook Graph access tokens** (`EAA…`,
-  ad-account, page and business-management scope), and **Telegram bot
-  tokens** (`<bot-id>:AA…`). These run in `BUILTIN_PATTERN_STRS`, so unlike
+  ad-account, page and business-management scope), **Telegram bot
+  tokens** (`<bot-id>:AA…`), and **GoHighLevel Private Integration Tokens**
+  (`pit-…`, which do not expire until manually revoked). These run in
+  `BUILTIN_PATTERN_STRS`, so unlike
   operator `[sanitize].extra_patterns` they also scrub client-side, before an
   excerpt reaches the local spool or the wire.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- The built-in sanitizer now redacts six further credential shapes, completing
+  three vendor families it already covered only in part. **GitHub:** every
+  token prefix — `gho_` (OAuth, the form `gh auth login` writes to disk),
+  `ghu_`, `ghs_`, `ghr_` — where previously only `ghp_` and `github_pat_`
+  were caught. **AWS:** `ASIA…` STS temporary key ids alongside `AKIA…`.
+  **Stripe:** `rk_live_` restricted keys alongside `sk_live_`. Plus three
+  providers with no prior coverage: **Google OAuth refresh tokens** (`1//…`,
+  longer-lived than the `AIza…` keys already handled — they mint access
+  tokens until revoked), **Meta / Facebook Graph access tokens** (`EAA…`,
+  ad-account, page and business-management scope), and **Telegram bot
+  tokens** (`<bot-id>:AA…`). These run in `BUILTIN_PATTERN_STRS`, so unlike
+  operator `[sanitize].extra_patterns` they also scrub client-side, before an
+  excerpt reaches the local spool or the wire.
+
+  Behaviour change: text matching these shapes is now replaced with
+  `[REDACTED]` where it previously persisted verbatim. An operator who
+  *wants* one of them kept visible (for example a public bot id) can add it
+  to `[sanitize].allowlist`.
+
 ## [1.28.0] - 2026-08-17
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   longer-lived than the `AIza…` keys already handled — they mint access
   tokens until revoked), **Meta / Facebook Graph access tokens** (`EAA…`,
   ad-account, page and business-management scope), **Telegram bot
-  tokens** (`<bot-id>:AA…`), and **GoHighLevel Private Integration Tokens**
+  tokens** (`<bot-id>:…`, covering both the `AA…` form every issued token uses
+  and the shape Telegram's own docs publish), and **GoHighLevel Private
+  Integration Tokens**
   (`pit-…`, which do not expire until manually revoked). These run in
   `BUILTIN_PATTERN_STRS`, so unlike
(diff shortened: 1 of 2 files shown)


tahazarif10:
One thing I'd loosen here is the Telegram pattern

Right now it requires an 8–10 digit bot id and `:AA...` but Telegram's own Bot API docs still show `123456:ABC-DEF...` as the token shape

So this would miss their documented example. I'd probably add that exact example as a regression test and avoid relying on the `AA` prefix

https://core.telegram.org/bots/api#authorizing-your-bot

dk96-creator:
Good catch — you're right, and I verified it against the docs before changing anything: the published example is `123456:ABC-DEF1234ghIkl-zyx57W2v1u123ew11`, which my pattern missed on both counts (6-digit id *and* no `AA`).

I went with two branches rather than dropping the `AA` reliance outright:

```
\b\d{6,10}:(?:AA[A-Za-z0-9_\-]{30,}|[A-Za-z0-9_\-]{34,35})\b
```

The reason for keeping an anchor on the second branch: a bare `\d+:[A-Za-z0-9_-]+` also matches `2026:08`, `16:9`, `8080:my-service-name-here`, SRT cues like `00:00:00,000 --> 00:00:04,120`, and `<short-sha>:<hex>` pairs — five false positives across the fixtures I tried. Length-anchoring that branch keeps the documented shape and drops all five. The `AA` branch stays open-ended so a future length change can't silently retire it.

Added your example as a regression test (`scrubs_telegram_documented_example_shape`), plus a negative control pinning the false-positive cases so the anchor can't be removed later without a test going red. Thanks for the link.

akitaonrails:
Merged via #409 — thank you, and nice first contribution. Your three commits landed verbatim, which is why this PR shows as merged.

Two maintainer commits went on top:

- `d869efd` anchors the AWS rule to the published format, `(?:AKIA|ASIA)[0-9A-Z]{16}`, rather than an open `{12,}` tail. `ASIA` is also an English word, so the open tail redacted ordinary uppercase text — `ASIAPACIFICREGION` and `ASIAEAST1CLUSTER` were both destroyed. Worth knowing for a privacy strip: over-redaction is the safe default for *credential* shapes, but not when the prefix is also a word. I added a negative test asserting ordinary text survives verbatim.
- `4ee9bcb` exempts your new fixtures in `.gitleaks.toml`. Fixtures for a sanitizer are key-shaped by construction, so the secret scanner flagged the Stripe and Telegram ones and failed CI.

I left the Meta `EAA{20,}` and the Telegram second branch as you wrote them — over-redaction is the right default there, and Meta publishes no minimum length.

PR #410, opened 2026-08-18
docs: document OrcaRouter as an openai-compat provider

XiaoHuo888-hue:
## What changed

Documents [OrcaRouter](https://www.orcarouter.ai) as a drop-in
`openai-compat` provider. OrcaRouter is an OpenAI-compatible router
(200+ models — OpenAI, Anthropic, Google, DeepSeek, ...) behind a single
API key, and it accepts the same `provider/model` id format as OpenRouter,
so ai-memory needs **no new provider type** — it works through the existing
`openai-compat` provider.

- **README.md**: added OrcaRouter to the `openai-compat` provider table row
  and to the "Ollama/OpenRouter/Atlas Cloud" example list.
- **docs/install.md**: added a named OrcaRouter example mirroring the
  existing OpenRouter/Atlas Cloud examples (`AI_MEMORY_LLM_BASE_URL=
  https://api.orcarouter.ai/v1` + `LLM_API_KEY=sk-orca-...`), and an
  embeddings row mirroring the OpenRouter one.
- **CHANGELOG.md**: `[Unreleased]` → `Changed` entry.

I'm an engineer on the OrcaRouter team.

## Why

ai-memory already routes `openai-compat` LLM and embedding calls through a
configurable base URL, and the install guide names OpenRouter and Atlas
Cloud as hosted compatibility endpoints. OrcaRouter fits the same slot with
no code change; this PR makes it discoverable the same way, with the exact
env vars a deployment needs.

## Test plan

This is a docs-only change — no Rust code or config behaviour was touched,
so the `cargo fmt` / `cargo clippy` / `cargo test` workspace gates are not
affected by this diff. What I did run:

- [x] `git diff --check` passes (no whitespace errors)
- [x] **Live API (real OrcaRouter key):** the exact `openai-compat` wire
      path (`POST {base}/chat/completions` with a Bearer token and
      `max_tokens`) against `https://api.orcarouter.ai/v1` returns HTTP 200
      with a valid completion for `openai/gpt-4o` and for
      `deepseek/deepseek-v4-flash`.
- [x] **Live API:** `POST {base}/embeddings` with
      `openai/text-embedding-3-small` returns HTTP 200 with an embedding
      vector, so the documented embeddings row works.
- [x] Manual test: `grep`-verified the anchor `#llm-provider-tiers` that
      README links to still resolves.

## Commit attribution

## CHANGELOG (merge gate)

      (No defaults changed — docs only.)

## Notes for reviewers

- No secrets are committed: the example uses the placeholder key
  `sk-orca-...` (matching how the OpenRouter example uses `sk-or-v1-...`).
- The CHANGELOG entry's `(#NNN)` trailing reference is finalized to
  `(#410)` in the follow-up commit `18d551f`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Documented OrcaRouter through the existing `openai-compat` provider instead
   of adding a redundant provider type, including the endpoint, model, and API
-  key mapping needed for deployment. (#NNN)
+  key mapping needed for deployment. (#410)
 
 ## [1.28.0] - 2026-08-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   *wants* one of them kept visible (for example a public bot id) can add it
   to `[sanitize].allowlist`.
 
+### Changed
+- Documented OrcaRouter through the existing `openai-compat` provider instead
+  of adding a redundant provider type, including the endpoint, model, and API
+  key mapping needed for deployment. (#NNN)
+
 ## [1.28.0] - 2026-08-17
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


akitaonrails:
Merged via #418 — thank you. Both your commits landed on `main` verbatim.

One maintainer commit on top relocated the CHANGELOG entry: your PR predates v1.28.1, so git auto-merged the entry *inside* the now-frozen `[1.28.1]` section without raising a conflict, which would have claimed the docs change shipped in a release that did not contain it. Content unchanged, just the section it sits under.

Since you disclosed working on OrcaRouter, I verified the claims against the live service rather than the PR body — standard here for vendor-authored docs, nothing adversarial:

- `api.orcarouter.ai` resolves; `GET /v1/models` → 200
- 191 models, 189 in `provider/model` form, matching the OpenRouter-compatible id format you documented
- every id cited in the docs exists: `openai/gpt-4o`, `anthropic/claude-sonnet-4.6`, `deepseek/deepseek-v4-flash`, and `openai/text-embedding-3-small` for the embeddings row

All accurate. Routing it through the existing `openai-compat` provider rather than adding a provider type was the right call — that is exactly how this surface is meant to extend. Ships in the next release.

PR #411, opened 2026-08-18
fix(install-hooks): honor PI_CODING_AGENT_DIR for Pi/OMP extension paths

### LINKED ISSUE #406: install-hooks: verify whether pi/omp extension paths should honor PI_CODING_AGENT_DIR
Found while auditing config-path resolution after #75376ac (`CODEX_HOME`).

## The question

`install_hooks.rs` hardcodes these:

- `pi_extension_path()` → `~/.pi/agent/extensions/ai-memory.ts`
- `omp_extension_path()` → `~/.omp/agent/extensions/ai-memory.ts`

Meanwhile `ai-memory-workstream`'s harness adapter *does* honor env overrides for the same two agents when locating transcripts:

```rust
ManagedHarness::Pi  => value("PI_CODING_AGENT_SESSION_DIR")
                        .or_else(|| value("PI_CODING_AGENT_DIR").map(|d| d.join("sessions"))),
ManagedHarness::Omp => value("PI_CODING_AGENT_DIR").map(|d| d.join("sessions")),
```

That is the same shape as the `CODEX_HOME` bug: one half of an install honors a home variable, the other half hardcodes `$HOME`, so with the variable set the hooks land where the agent never reads them and capture silently does nothing.

## Why this is not simply fixed the same way

The internal evidence is contradictory, so I did not want to guess:

- The harness derives `$PI_CODING_AGENT_DIR/sessions`, while the extension path is `~/.pi/**agent**/extensions` — different nesting under the supposed same root. If `PI_CODING_AGENT_DIR` relocates `~/.pi`, the extension path should become `$PI_CODING_AGENT_DIR/agent/extensions`, but that does not follow from the sessions path alone.
- `PI_CODING_AGENT_DIR` governing OMP's `~/.omp` is odd on its face and may be an unrelated reuse.

Guessing wrong reproduces the exact bug being fixed — writing extensions somewhere the agent will not load them — so this needs Pi's and OMP's own documentation rather than inference from this codebase.

## What to check

1. Does `PI_CODING_AGENT_DIR` relocate Pi's whole config home, or only its session storage?
2. Is there a separate variable for the extensions directory?
3. Does OMP honor `PI_CODING_AGENT_DIR`, or does it have its own?

If the answer is "it relocates the home", the fix mirrors `codex_hooks_path_in`: resolve via `path_util::agent_config_home`, add a two-branch test, and `uninstall` picks it up for free since it delegates to the same resolver.

## Not affected

`opencode_plugin_path` and `zero_hooks_path` hardcode `~/.config` **deliberately** — both carry a comment stating the policy, and `install-hooks --config-file` covers non-default XDG setups (verified working). Those are not part of this.

rekcilyssup:
## What changed

`install-hooks --agent pi`, `install-hooks --agent omp`, and `uninstall` now honor `PI_CODING_AGENT_DIR` when resolving the generated extension path, instead of always writing to `~/.pi/agent/extensions/ai-memory.ts` and `~/.omp/agent/extensions/ai-memory.ts`.

- When `PI_CODING_AGENT_DIR` is set (non-empty, non-whitespace): extension goes to `$PI_CODING_AGENT_DIR/extensions/ai-memory.ts`.
- When unset or blank: behavior is unchanged (default `~/.pi/agent/...` / `~/.omp/agent/...`).
- `uninstall` picks this up for free because it delegates to the same resolvers.

## Why

Fixes #406. Both Pi and OMP use `PI_CODING_AGENT_DIR` to relocate their **whole** agent config home (`~/.pi/agent` / `~/.omp/agent` — not just session storage; per [Pi's environment-variable docs](https://pi.dev/docs/latest/environment-variables) and [OMP's docs](https://github.com/can1357/oh-my-pi/blob/main/docs/extension-loading.md), extensions live under that home in `extensions/`). With the variable set, the old code wrote the extension where the agent never loads it: the install reported success and capture silently did nothing.

`ai-memory-workstream`'s harness adapter already honored the variable when resolving Pi/OMP transcripts (`ManagedHarness::Pi` / `ManagedHarness::Omp`), so this closes the gap where one half of the same install disagreed with the other — the same shape as the `CODEX_HOME` fix in #379/1.28.0.

Note: the issue's working hypothesis ("if `PI_CODING_AGENT_DIR` relocates `~/.pi`, the extension path becomes `$PI_CODING_AGENT_DIR/agent/extensions`") is incorrect per upstream docs — the variable relocates the agent dir itself, so the extension resolves to `$PI_CODING_AGENT_DIR/extensions`. This mirrors the harness's existing `$PI_CODING_AGENT_DIR/sessions` shape exactly.

## Test plan

- [x] New tests `pi_extension_path_honours_pi_coding_agent_dir` and `omp_extension_path_honours_pi_coding_agent_dir` cover the env-set branch, the unset branch, and the blank-as-unset branch (env value injected as a parameter — the workspace forbids `unsafe`, so `std::env::set_var` is not an option).

## Commit attribution

## CHANGELOG (merge gate)

- [x] Added `CHANGELOG.md` entry under `## [Unreleased]` → `### Fixed`.

## Notes for reviewers

- The `--config-file` override path is untouched and still wins when provided.
- Named OMP profiles (`omp --profile X`) intentionally ignore `PI_CODING_AGENT_DIR` upstream; this fix honors the variable exactly where OMP itself honors it (default profile). `PI_CONFIG_DIR` handling for OMP's config root was deliberately left out of scope, matching the harness adapter's current behavior.
- README support-matrix lines still describe the default path; the variable note lives in `docs/install.md` and `docs/mcp-install.md` where the env behavior matters.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- `install-hooks --agent pi`, `--agent omp`, and `uninstall` now honor
+  `PI_CODING_AGENT_DIR`, instead of always writing to `~/.pi/agent/extensions/`
+  and `~/.omp/agent/extensions/`. Both agents relocate their whole agent
+  config home (`~/.pi/agent` / `~/.omp/agent`) through that variable, so on an
+  install with it set the extensions landed where the agent never loads them:
+  the install reported success and capture silently did nothing. ai-memory
+  already honored the variable when resolving Pi/OMP transcripts, so the two
+  halves of one install disagreed about where that home was. Installs without
+  `PI_CODING_AGENT_DIR` set are unaffected.
+
 ### Added
 - The built-in sanitizer now redacts seven further credential shapes, completing
   three vendor families it already covered only in part. **GitHub:** every
--- a/crates/ai-memory-cli/src/commands/install_hooks.rs
+++ b/crates/ai-memory-cli/src/commands/install_hooks.rs
@@ -162,8 +162,28 @@ pub(crate) fn opencode_plugin_path() -> anyhow::Result<std::path::PathBuf> {
         .join("ai-memory.ts"))
 }
 
+/// `$PI_CODING_AGENT_DIR/extensions/ai-memory.ts` when the var is set, else
 /// `~/.omp/agent/extensions/ai-memory.ts` — OMP lifecycle extension.
+///
+/// `PI_CODING_AGENT_DIR` relocates OMP's whole agent config home
+/// (`~/.omp/agent`), so extensions written to the default path are never
+/// loaded by an OMP configured that way: capture silently does nothing, with
+/// a successful-looking install. ai-memory already honors the variable when
+/// it resolves OMP transcripts (`ManagedHarness::Omp` in
+/// `ai-memory-workstream`), so ignoring it here left one half of the same
+/// install pointing somewhere the other half did not.
 pub(crate) fn omp_extension_path() -> anyhow::Result<std::path::PathBuf> {
+    omp_extension_path_in(std::env::var_os("PI_CODING_AGENT_DIR"))
+}
+
+/// The env value comes in as a parameter so tests can exercise both branches
+/// without mutating process env (mirrors [`codex_hooks_path_in`]).
+fn omp_extension_path_in(
+    env_override: Option<std::ffi::OsString>,
+) -> anyhow::Result<std::path::PathBuf> {
+    if let Some(dir) = crate::commands::path_util::agent_config_home(env_override) {
+        return Ok(dir.join("extensions").join("ai-memory.ts"));
+    }
     Ok(home_dir()
         .context("could not locate $HOME for ~/.omp/agent/extensions")?
         .join(".omp")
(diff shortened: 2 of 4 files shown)


akitaonrails:
Merged via #419, closing #406 — thank you. Your commit `05059af` is on `main` verbatim.

Two things worth reporting back.

**I verified the premise independently.** #406 was open specifically to *check whether* Pi/OMP extension paths should follow this variable, so I went to OMP's `docs/extension-loading.md` rather than taking the PR body's word:

> "the active agent directory's `extensions/` (default `~/.omp/agent/extensions`)" … "`PI_CODING_AGENT_DIR` can override the agent directory"

Confirmed. Reusing `path_util::agent_config_home` and the `*_in(env_override)` pattern from the `CODEX_HOME` fix was the right seam — that parameter injection is what makes both branches testable at all, given `unsafe_code = "forbid"` rules out `set_var`.

**One gap I fixed on top.** The env-aware resolution landed on the `--apply` path only. The instructions printed *without* `--apply` still hardcoded `~/.pi/agent/extensions/ai-memory.ts`, so someone with the variable set who installs by hand was told to copy the extension exactly where the agent never loads it — the same silent no-capture, surviving in the half of the command that output exists to serve. Both hints now route through your resolvers, with a test covering set and unset.

**A limitation neither of us handles**, noted for anyone who hits it: Pi and OMP read the *same* variable and both load from `$DIR/extensions/`, while ai-memory writes the same filename with different content for each (Pi's bridges MCP tools; OMP's does not). Running both with the variable set is last-install-wins. That is inherent to upstream sharing one variable, and still strictly better than ignoring it. OMP also documents per-profile homes (`~/.omp/profiles/<name>/agent/extensions`) that neither this change nor the base handles. I will open a follow-up issue.

Ships in the next release.

Issue #412, opened 2026-08-18
memory_read_page root-level anyOf breaks generic MCP clients routed through Moonshot upstreams (OpenCode + Console Go)


igormaneschy:
**Environment**
- ai-memory 1.28.0 (macOS arm64 binary), MCP over HTTP (`http://127.0.0.1:49374/mcp`)
- Client: **OpenCode** (generic MCP client, sends no custom query params on `tools/list`)
- Model provider: "Console Go" gateway forwarding tool definitions verbatim to the Moonshot API

**Reproduction**
1. Connect the ai-memory MCP server to OpenCode with any Moonshot-routed model.
2. Make any request that sends the tool list upstream.
3. HTTP 400 from the provider:

```
Invalid request: tools.function.parameters is not a valid moonshot flavored json schema,
details: <At path 'root': when using anyOf, type should be defined in anyOf items instead of the parent schema>
```

**Root cause**
`ReadPageArgs` encodes the "exactly one of `path`/`query`" contract (#155 — added to stop OpenCode's `path=null` retry loop) via a root-level `anyOf` in the tool input schema. Moonshot's validator rejects `type` at the parent level alongside `anyOf`.

v1.28.0 ships `restricted_schema_tool_list` (strips root-level `anyOf`/`oneOf`/`allOf`) — but it activates **only** when the client calls `tools/list?flavor=moonshot` or `?flavor=bedrock` (Kimi Code legacy / Kiro install conventions, from #195/#351). OpenCode, Cursor, and other generic MCP clients never send that query parameter, so they receive the unpatched schema and every Moonshot-routed request fails hard with no fallback.

**Suggested fixes (any one)**
1. **Detect by client identity too**: read `clientInfo.name` from the `initialize` handshake (`opencode`, `cursor`, `claude-desktop`, …) and serve the restricted schema to a known list of clients that forward schemas verbatim to strict validators.
2. **Make it configurable**: `[mcp] strip_root_combinators = true` (or env `AI_MEMORY_STRICT_SCHEMA`) so operators behind any strict upstream can opt in without per-client detection.
3. Keep the runtime "exactly one of" validation as the enforcement backstop (it already exists), so any schema patch never regresses #155.

**Workaround**
Use a non-Moonshot model for sessions that call ai-memory tools, or avoid `memory_read_page` (`memory_query` has no root combinator and works).

PR #413, opened 2026-08-18
feat(mcp): add strip_root_combinators toggle for schema-strict upstreams

changshenhan:
@-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- New `strip_root_combinators` config flag (env `AI_MEMORY_STRIP_ROOT_COMBINATORS`,
+  or `strip_root_combinators = true` in config.toml) strips root-level
+  `anyOf`/`oneOf`/`allOf` from MCP tool input schemas on every `tools/list`.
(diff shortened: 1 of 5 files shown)


akitaonrails:
Merged via #420, closing #412 — thank you. Your commit `68d650d` is on `main` verbatim.

Worth saying plainly: #416 arrived ~100 minutes after yours and solved the same issue at the same seam — same `||` in `list_tools`, same bool on `AiMemoryServer`, same builder method, same config field, same call site. Two people independently finding the same right answer, which is a good sign the seam is correct.

I took yours on two counts:

- **You tested both directions.** Your suite asserts the toggle strips *and* that with it off the root `anyOf` survives, which is what pins the #155 early-refusal behavior. The other PR asserted only that stripping works, so nothing held the default in place. In this repo that control-test habit has caught real regressions that review missed — it is the difference that decided this.
- **Naming.** `strict_schema` would have sat in the same `Config` struct as the existing `llm_compat_strict` and read as related when it is not. `strip_root_combinators` says what it does.

One maintainer commit on top ported #416's `config.default.toml` block, adapted to your naming and credited to rizperdana — a toggle that lives only in source and an env var is one an operator can only find if they already know it exists. The key is placed above the first `[section]` header, since a bare top-level key emitted after one gets parsed into that section (the trap that broke `ai-memory init` when `[routing]` was added).

Ships in the next release.

PR #414, opened 2026-08-18
feat(mcp): add strip_root_combinators toggle for schema-strict upstreams

changshenhan:
Adds an opt-in `strip_root_combinators` toggle (issue #412).

## Problem

Generic MCP clients such as OpenCode and Cursor never send the `?flavor=` marker, yet forward tool schemas verbatim to strict upstreams (Moonshot, Bedrock) that reject root-level `anyOf`/`oneOf`/`allOf` with a 400. Operators routing through such an upstream had no mode-independent way to serve the restricted schema dialect.

## What

- New `strip_root_combinators` config flag — `AI_MEMORY_STRIP_ROOT_COMBINATORS=true` or `strip_root_combinators = true` in `config.toml`.
- `AiMemoryServer::with_strip_root_combinators(enabled)` builder mirroring the existing toggle builders.
- On `tools/list`, the restricted dialect is served unconditionally when the toggle is on; the `?flavor=` marker remains the per-client override.

Runtime "exactly one of" validation is unchanged and remains the enforcement backstop.

## Tests

New `stateless_config_strip_strips_root_any_of_without_flavor` exercises the toggle over the same stateless HTTP transport wiring — no flavor parameter, restricted schema still served.

Full `cargo test -p ai-memory-mcp` passes (216 unit + 131 integration tests, 0 failures), including the existing flavor-marker tests to confirm the per-client path is untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- New `strip_root_combinators` config flag (env `AI_MEMORY_STRIP_ROOT_COMBINATORS`,
+  or `strip_root_combinators = true` in config.toml) strips root-level
+  `anyOf`/`oneOf`/`allOf` from MCP tool input schemas on every `tools/list`.
+  Generic MCP clients such as OpenCode and Cursor never send the `?flavor=`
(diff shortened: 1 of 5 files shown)


changshenhan:
Duplicate of #413 — I renamed the local branch (fix/ → feat/) and this re-push created a second PR for the same commit. #413 is canonical.

PR #416, opened 2026-08-18
feat(mcp): configurable strict (combinator-stripped) tool schemas

### LINKED ISSUE #412: memory_read_page root-level anyOf breaks generic MCP clients routed through Moonshot upstreams (OpenCode + Console Go)

rizperdana:
Fixes #412

## Problem

`restricted_schema_tool_list()` (strips root-level `anyOf`/`oneOf`/`allOf` from tool input schemas) activates **only** for requests carrying `?flavor=moonshot` or `?flavor=bedrock`. Generic MCP clients — OpenCode, Cursor, and others — never send that marker, yet they forward tool schemas verbatim to strict upstream validators (Moonshot's "moonshot flavored json schema" dialect). Result: any Moonshot-routed session calling `memory_read_page` fails hard with HTTP 400 at `tools/list` and no fallback.

## Change

New opt-in config `AI_MEMORY_STRICT_SCHEMA=true` (config.toml: `strict_schema = true`, default `false`) forces the restricted (combinator-stripped) tool list for **every** client, regardless of flavor marker. Operators behind any strict upstream can enable it without per-client detection heuristics.

- `config.rs`: `Config::strict_schema` (env `AI_MEMORY_STRICT_SCHEMA`)
- `serve.rs`: wired into `AiMemoryServer` via `.with_strict_schema()`
- `server.rs`: `AiMemoryServer::strict_schema` field + builder; `list_tools` serves the restricted list when `self.strict_schema ||` flavor marker present
- `config.default.toml`: documented commented option

Root combinators only are stripped; nested combinators and runtime "exactly one of" validation are unchanged (the #155 contract stays).

## Verification

- New integration test `stateless_strict_schema_strips_root_any_of_without_flavor` — `tools/list` with no `?flavor=` returns `memory_read_page` without root combinators, `properties` intact.
- Existing flavor tests (`stateless_moonshot_flavor_strips_root_any_of`, `stateless_bedrock_flavor_strips_root_any_of`, control `stateless_tools_list_without_flavor_keeps_root_any_of`) all still pass — strict mode is opt-in, default behavior unchanged.
- `cargo check -p ai-memory-cli` + `cargo fmt --check` clean.

### DIFF
--- a/crates/ai-memory-cli/src/commands/serve.rs
+++ b/crates/ai-memory-cli/src/commands/serve.rs
@@ -485,7 +485,8 @@ pub async fn run(config: &Config, args: ServeArgs) -> Result<()> {
         .with_active_project(active_project.clone())
         .with_sanitizer(sanitizer.clone())
         .with_trusted_proxy_identity(trusted_proxy_identity_enabled(&config.auth))
-        .with_per_user_slots(config.slots.per_user);
+        .with_per_user_slots(config.slots.per_user)
+        .with_strict_schema(config.strict_schema);
     if let Some(e) = embedder.clone() {
         server = server.with_embedder(e);
     }
--- a/crates/ai-memory-cli/src/config.rs
+++ b/crates/ai-memory-cli/src/config.rs
@@ -140,6 +140,12 @@ pub struct Config {
     /// already supplies its own schema. Set
     /// `AI_MEMORY_LLM_COMPAT_STRICT=false` for an incompatible endpoint.
     pub llm_compat_strict: bool,
+    /// Serve restricted (combinator-stripped) tool schemas on every
+    /// `tools/list`, even when the client sends no `?flavor=` marker, for
+    /// strict upstream validators that forward tool schemas verbatim
+    /// (Moonshot/Bedrock dialect) from generic MCP clients (issue #412).
+    /// Set with `AI_MEMORY_STRICT_SCHEMA=true`.
+    pub strict_schema: bool,
     /// Opt-in: run LLM consolidation on SessionEnd (in addition to the
     /// always-written heuristic session page), when an LLM provider is
     /// configured. Off by default. Provider work is durably queued after the
@@ -543,6 +549,7 @@ impl Default for Config {
             llm_compat_strict: true,
             consolidate_on_session_end: false,
             capture_assistant: false,
+            strict_schema: false,
             reranker: None,
             embedding_provider: None,
             embedding_model: None,
--- a/crates/ai-memory-cli/templates/config.default.toml
+++ b/crates/ai-memory-cli/templates/config.default.toml
@@ -35,6 +35,13 @@ log_level = "info"
 # --capture-assistant`. Env equivalent: `AI_MEMORY_CAPTURE_ASSISTANT=true`.
 # capture_assistant = false
 
+# Serve restricted tool schemas on every tools/list (strips root-level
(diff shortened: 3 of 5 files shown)


akitaonrails:
Closing in favour of #413 (merged as #420) — but part of your work shipped, and you are credited on the commit.

To be straight about what happened: you and @changshenhan solved #412 independently, about 100 minutes apart, and landed on the *same* seam — one `||` in `list_tools`, a bool on `AiMemoryServer`, a builder method, a config field, one call site in `serve.rs`. The core diffs are functionally indistinguishable. This was not a duplicate of something already obvious; two people reading the same code found the same correct answer.

#413 went in on two differences:

- **Test coverage.** It asserts both directions — the toggle strips, *and* with it off the root `anyOf` survives. That second one pins the #155 early-refusal behavior. Yours asserts the stripping works, so nothing held the default in place. Control tests have caught regressions in this repo that review missed, so that weighed heavily.
- **Naming.** `strict_schema` would sit in the same `Config` struct as the existing `llm_compat_strict` and read as related to it. `strip_root_combinators` names the mechanism.

**What shipped from your PR:** the `config.default.toml` documentation block, which #413 did not have. I ported it in the maintainer commit on #420 with `Co-Authored-By: rizperdana`. That was the right instinct and the concrete thing yours had that the other did not — a toggle documented only in Rust source and an env var is one an operator can only find by already knowing it exists.

Thanks for the work, and please do send more. Nothing here was wrong; it came down to the negative test.

Issue #421, opened 2026-08-18
install-hooks: Pi and OMP share PI_CODING_AGENT_DIR, and neither handles OMP profiles


akitaonrails:
Two limitations found while auditing #411 (merged as #419). Neither blocks that fix — both are strictly better than the pre-#411 behavior, where the variable was ignored entirely — but they are real and worth tracking.

## 1. Pi and OMP collide on one filename

Pi and OMP read the **same** `PI_CODING_AGENT_DIR`, and both load extensions from `$PI_CODING_AGENT_DIR/extensions/`. ai-memory writes the same filename, `ai-memory.ts`, for both — with **different content**: `build_pi_extension` bridges MCP tools, `build_omp_extension` does not.

So with the variable set and both agents installed:

```
ai-memory install-hooks --agent pi  --apply   # writes $DIR/extensions/ai-memory.ts (Pi content)
ai-memory install-hooks --agent omp --apply   # overwrites it with OMP content
```

Last install wins, silently, and one of the two agents ends up running an extension built for its sibling. Without the variable set the two land in `~/.pi/agent/` and `~/.omp/agent/` and never collide.

This is inherited from upstream sharing one variable across both forks, so there may be no fully correct answer. Options worth weighing:

- detect an existing `ai-memory.ts` whose content targets the other agent and warn rather than overwrite;
- write agent-distinct filenames if both agents will load any `*.ts` in that directory (needs checking against how each fork enumerates extensions);
- document the collision and leave it.

## 2. OMP profiles are not handled

OMP's `docs/extension-loading.md` describes per-profile agent homes:

> `~/.omp/profiles/<name>/agent/extensions` when using `omp --profile <name>`

with `PI_CODING_AGENT_DIR` still able to override at that level. Neither #411 nor the code before it resolves a profile, so `install-hooks --agent omp` on a profile-using install writes to the non-profile home and the extension is never loaded — the same silent no-capture shape as #406, just triggered by `--profile` instead of the env var.

Reproducing this needs an OMP install with a named profile; I have not confirmed it against a live one, only against the vendor docs.

## Context

- Env-var handling landed in #419 (from #411), closing #406
- Same failure family as #379 (`CODEX_HOME`) and #406: install reports success, capture silently does nothing, because two halves of one install disagree about where the agent home is

PR #422, opened 2026-08-18
fix(install-hooks): resolve Pi and OMP extension collision and suppor…

kdahal7:
…t OMP profiles (#421)

## Test plan

- [ ] Manual test: 

## Commit attribution

## CHANGELOG (merge gate)

- [ ] I added a `CHANGELOG.md` `[Unreleased]` entry — **required** for any

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `config.default.toml` so it is discoverable without reading the source.
 
 ### Fixed
+- `install-hooks` now uses agent-distinct filenames (`ai-memory-pi.ts` for Pi and `ai-memory-omp.ts` for OMP) to eliminate silent extension overwrites when `PI_CODING_AGENT_DIR` is set for both agents, automatically cleaning up legacy `ai-memory.ts` files. Added support for OMP profiles via `--profile` flag and `OMP_PROFILE` environment variable. (#421)
 - `install-hooks --agent pi`, `--agent omp`, and `uninstall` now honor
   `PI_CODING_AGENT_DIR`, instead of always writing to `~/.pi/agent/extensions/`
   and `~/.omp/agent/extensions/`. Both agents relocate their whole agent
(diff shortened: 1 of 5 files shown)


akitaonrails:
Merged via #433, closing #421 — thank you. Your commit is on `main` verbatim, and the profile support was the piece I did not have a plan for.

One maintainer commit on top. Three things needed fixing, and I want to be specific because two of them are easy to hit again.

**It did not compile.** Two `E0716` borrow errors — `std::env::var_os(..).as_ref().and_then(..)` borrows from a temporary that drops at the end of the statement. Worth knowing that **no CI had run on this PR**: fork PRs here need maintainer approval before workflows execute, so a green-looking PR page is not evidence. Bind the `OsString` first and it is fine.

**The legacy cleanup could delete a user's file.** Both apply paths removed any `ai-memory.ts` beside the new one with a bare `remove_file`. `uninstall` gates every delete behind `generated_file_is_ours`, and `removal.rs` already asserts that a *user-owned* file with that exact name survives — so install would have destroyed precisely what uninstall refuses to touch, silently, on a path the operator never asked to modify. Removal now requires this tool's generated marker **and** matches the installing agent, so a Pi install cannot delete OMP's file in a shared home. Your uninstall-side change was already correct — it routed through `push_generated_delete`.

**Distinct filenames do not actually fix #421.** This is the one worth sitting with. Per OMP's docs it loads *every* direct `*.ts` in the extensions directory. With `PI_CODING_AGENT_DIR` set, both files land in one directory, so each agent loads both — and since the Pi extension is the OMP one with `const AGENT` swapped and neither guards on its host, every lifecycle event would be captured **twice under two agent identities**. The rename trades a silent overwrite for silent duplication, which is worse for the store.

That collision is upstream — both agents honour one variable — so the installer cannot fix it alone. It now warns when it detects the shared directory and names the two ways out. The renaming is still right; it just is not sufficient on its own.

I also simplified the profile precedence: the derivation re-rooted only when the override path's last component happened to be named `agent`, so `/custom/agent` and `/custom/pi-agent` produced different shapes from identical inputs. Upstream does not document composing the variable with a profile, so each keeps its documented meaning.

Ships in the next release.

kdahal7:
Thank you for merging and for the detailed write-up on the follow-up commit! Glad the profile support helped!

PR #423, opened 2026-08-18
fix: update default Gemini model to 3.5 Flash

vitorvilas:
## Summary

- Change the default Gemini provider model from `gemini-2.5-flash` to `gemini-3.5-flash`.
- Keep the thinking-budget workaround for both the legacy model and the new default.
- Update documentation, smoke-test defaults, and changelog.

Google documents that `gemini-2.5-flash` will be discontinued on **October 20, 2026**:
https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/model-versions?hl=pt-br

## Validation

- `git diff --check` ✅
- `bash -n tests/e2e/handoff_smoke.sh` ✅
- Live Gemini API call from a Docker installation returned `OK` using `gemini-3.5-flash`.

## Notes

- No changes to the Google API integration itself — only the default model string, docs, and tests.
- The container image and deployment flow remain unchanged.


akitaonrails:
Merged via #430 — thank you, and good catch on getting ahead of the 2.5 retirement. Your commits are on `main` verbatim.

One maintainer commit on top, for a defect worth knowing about since it is easy to hit again.

`default_thinking_config_for` matches on **substrings**, so `contains("gemini-3.5-flash")` also caught `gemini-3.5-flash-lite`. I checked each model against the live API rather than reasoning about it:

| model | `thinkingBudget: 0` |
|---|---|
| `gemini-3.5-flash` | 200 |
| `gemini-2.5-flash-lite` | 200 |
| `gemini-3.5-flash-lite` | **400 invalid argument** |
| `gemini-3.5-flash-lite`, field omitted | 200 |

So 3.5-lite behaves differently from its 2.5 predecessor: it rejects the field outright. Anyone setting `AI_MEMORY_LLM_MODEL=gemini-3.5-flash-lite` would have failed on every call, and the e2e `MODEL_B` change would have failed at the session-2 call, since `gemini_call` hardcodes `thinkingBudget=0`.

The fix excludes `-lite` for 3.5 only. 2.5 deliberately keeps sending the field — narrowing it there would re-enable its default thinking and let hidden thought tokens truncate strict JSON, which is the reason that function exists. `MODEL_B` stays `gemini-2.5-flash-lite`.

I also adjusted the retirement date: the **Gemini API** deprecation table lists October 16 2026 for `gemini-2.5-flash-lite`, while October 20 is the Vertex AI / Agent Platform date. ai-memory talks to the Gemini API, so the earlier one applies. Your direction was right either way.

Ships in the next release.

vitorvilas:
Thanks, Akita! Good catch on the 3.5 Flash Lite behavior. I hadn’t noticed the substring matching side effect there. Also noted on the Gemini API retirement date. Glad the PR helped!

PR #424, opened 2026-08-19
feat: add flake.nix for NixOS / Nix support (#405)

BobDylans:
## What changed

Added a `flake.nix` so NixOS and Nix users can build and run ai-memory without the Rust toolchain or Docker:

- `nix build` → native release binary (`result/bin/ai-memory`)
- `nix run . -- --version` → smoke-test without installing
- `nix develop` → dev shell with Rust 1.95 (pinned via `rust-toolchain.toml`)

The build is fully self-contained — no `nativeBuildInputs` needed:
- SQLite is bundled via `rusqlite`'s `bundled` feature
- libgit2 is vendored via `git2`'s `vendored-libgit2` feature
- TLS uses rustls with `webpki-roots` — no OpenSSL, no system-library hunting, no `cacert` wrapping
- `TAILWIND_SKIP=1` makes the web crate use the vendored `static/tailwind.css` instead of downloading the Tailwind CLI (which a sandboxed Nix build cannot do)

`doCheck = false` skips the packaging test suite (`tests/packaging.rs`). Those 24 tests exercise `bin/ai-memory`, a Docker-wrapper shell script that needs `docker`/`podman` on PATH — they are host-environment tests, not build tests. The rest of the workspace test suite (unit + integration) does not need Docker and can be run via `nix develop -c cargo test --workspace`.

`postInstall` copies `hooks/` to `$out/share/ai-memory/` (mirroring the AUR PKGBUILD) and the config template to `$out/etc/ai-memory/`, so `ai-memory serve` and `install-hooks` find their runtime assets.

## Why

Closes #405. @wtf9880 opened the issue and wrote a first `flake.nix` based on `ai-jail`'s, but it failed in `checkPhase` because the packaging tests need Docker and the flake had unnecessary dependencies (`openssl`, `sqlite`, `git`, `pkg-config`, `cacert`). This PR fixes both: no extra deps, and `doCheck = false` for the Docker-dependent tests.

## Design decisions

- **No `nativeBuildInputs`**: the build is self-contained. Adding `openssl`/`sqlite`/`git`/`pkg-config`/`cacert` (as the first attempt did) is unnecessary and misleading — `Cargo.lock` has zero `openssl` occurrences and 19 `rustls` occurrences.
- **`doCheck = false`** rather than `checkPhase` patching: the packaging tests are fundamentally host-environment tests, not Nix's responsibility. A Nix user with Docker can still run them via `nix develop -c cargo test -p ai-memory-cli --test packaging`.
[...]

PR #425, opened 2026-08-19
improve(handoff): multi-signal open_questions heuristic

BobDylans:
## What changed

  The `open_questions` field in automatic SessionEnd handoffs now uses multi-signal heuristics (`derive_open_questions`) instead of blindly copying the last user prompt.

  ### Before
```

open_questions: ["Continue from: 好的"]  ← useless

```

  ### After
```

// Scenario: last prompt is a question
open_questions: ["Unresolved question: what is the return type?"]

// Scenario: edit without subsequent Stop (crash / abnormal exit)
open_questions: [
  "Changes were made but the session ended before verification",
  "Run tests or lint to confirm the changes are correct"
]

// Scenario: mid-task exit (Stop without SessionEnd)
open_questions: ["Continue from (mid-task exit): fix the bug in main.rs"]

// Scenario: normal end with substantive last prompt
open_questions: ["Continue from: fix the bug in main.rs"]

// Scenario: last prompt is acknowledgment ("ok", "thanks", "好的")
open_questions: []  ← filtered, no useless entry

```

  ## Why

  The previous `build_auto_handoff` always did `"Continue from: <last prompt>"`. This produced useless entries when the last prompt was an acknowledgment or when the session ended abnormally after an edit. The next agent
receiving the handoff would see `Continue from: 好的` — no information value.

  ## Heuristics (first match wins)

  1. Last prompt ends with `?` or `？` → tagged as `Unresolved question: …`
  2. Last tool was edit/write/patch with no subsequent Stop → `Changes were made but the session ended before verification` + `Run tests or lint to confirm the changes are correct`
  3. Has Stop but no SessionEnd → `Continue from (mid-task exit): …` (flag abnormal exit)
  4. Normal end → `Continue from: …` (unchanged from old behaviour)
  5. Fallback → empty Vec (same as old behaviour when no substantive prompts)

  Acknowledgments ("ok", "thanks", "好的", "谢谢", …) are filtered by `is_acknowledgment` so they never produce a `Continue from: 好的` line.

  ## Test plan

  - [x] `cargo test -p ai-memory-hooks -- open_questions` — 7 tests pass:
    - `open_questions_extracts_question_mark_prompt`
    - `open_questions_detects_unverified_edit`
    - `open_questions_filters_acknowledgment`
    - `open_questions_uses_substantive_last_prompt`
    - `open_questions_empty_when_no_prompts`
    - `open_questions_chinese_question_mark_detected`
    - `open_questions_mid_task_exit_has_signal`
    - `is_acknowledgment_catches_common_phrases`
  - [x] `cargo test -p ai-memory-hooks` — all 251 tests pass (no regressions)

  ## CHANGELOG (merge gate)

  - [x] Added `### Improved` entry under `[Unreleased]`.
  - [x] No default changed — the normal-end path (heuristic 4) produces identical output to the old behaviour.

  ## Notes for reviewers

  - The change is purely in `build_auto_handoff` logic; no schema, API, or storage changes.
  - `is_acknowledgment` is intentionally conservative: only short well-known phrases are caught. Anything longer than 20 chars or containing substantive content is kept.
  - `cap_handoff_text` is a standalone copy of the inner `cap` closure so `derive_open_questions` can use it without access to the closure. The two are kept in sync by the test suite.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,7 @@ evals/runs/
 
 # ai-memory atomic-write backups (apply_atomic in apply_shared.rs)
 *.bak-*
+
+# Nix flake build output (nix build / nix run).
+result
+result-*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   unchanged ([#412]). The key is documented in the generated
   `config.default.toml` so it is discoverable without reading the source.
 
+### Improved
+- The `open_questions` field in automatic SessionEnd handoffs now uses
+  multi-signal heuristics instead of blindly copying the last user prompt.
+  Trailing acknowledgments ("ok", "thanks", "好的") are filtered; a
+  question-mark-terminated final prompt is tagged as an unresolved question;
+  an edit/write tool call without a subsequent Stop produces an explicit
+  "changes not verified" advisory; and a mid-task exit (Stop without
+  SessionEnd) is flagged so the receiver knows the previous session did not
+  finish cleanly. (#NNN)
+
 ### Fixed
 - `install-hooks --agent pi`, `--agent omp`, and `uninstall` now honor
   `PI_CODING_AGENT_DIR`, instead of always writing to `~/.pi/agent/extensions/`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added a `flake.nix` so NixOS and Nix users can build and run ai-memory
+  without the Rust toolchain or Docker: `nix build`, `nix run . --
+  --version`, or `nix develop` for a dev shell. The build is self-contained
+  (SQLite bundled, libgit2 vendored, rustls with webpki-roots — no OpenSSL,
+  no system-library hunting). The flake skips the packaging test suite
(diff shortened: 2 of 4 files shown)


akitaonrails:
Merged via #432 — thank you. Your commits are on `main` verbatim.

The underlying complaint was right and worth fixing: `open_questions: ["Continue from: 好的"]` is exactly the kind of handoff line that wastes a receiver's time, and the acknowledgment filter is nicely conservative about it.

One maintainer commit on top, for something worth knowing about.

**Heuristic 2 could never fire.** It matched `tool.title` against `"edit"` / `"write"` / `"patch"`, but a PostToolUse observation's title is `canonical_tool_name(tool_family)` — only ever `"file"`, `"search-list"`, `"non-file"`, or `"unknown"`. The reserved protocol deliberately carries no raw tool names, so that branch was unreachable in production.

The test passed because its fixture built `mk_obs(PostToolUse, "edit", …)` by hand — a value the ingest path never emits. That is the trap worth taking away from this: a fixture that constructs the input directly can validate a function perfectly while the feature is dead, and the greener the suite looks the less likely anyone re-checks. Tests now build fixtures through `canonical_tool_name`, and there is an explicit guard asserting raw tool names do **not** drive the branch.

Rewording was forced by the same closed schema: `ToolFamily::File` covers reads and writes alike and `ToolOutcome` is only success/error/unknown, so nothing available can tell whether the file was modified. The advisory now says the session ended without a normal Stop while working with files and asks the receiver to check the tree — true with the signals that exist, rather than asserting changes were made.

I also dropped the `flake.nix` and `.gitignore` this branch carried; they were byte-identical to #424 and shipped there as #431.

Ships in the next release.

Issue #427, opened 2026-08-19
`latest` tag published as amd64-only manifest, breaking arm64 hosts (exec format error)


lucasliet:
## Bug report

After the most recent image update, the `ai-memory` container fails to start on arm64 (aarch64) hosts with:

```
exec /usr/local/bin/ai-memory: exec format error
```

## Root cause

The `latest` tag was pushed as a **single-architecture (amd64) manifest**, not a multi-arch manifest list, so Docker on an arm64 host pulls the amd64 layer and fails to exec the binary.

Confirmed via Docker Hub API (`/v2/repositories/akitaonrails/ai-memory/tags`):

| tag | architectures | last_updated |
|---|---|---|
| `latest` | `['amd64']` | 2026-08-18T17:46:54Z |
| `1.28.1` | `['arm64', 'amd64']` | 2026-08-18T16:15:27Z |
| `1.28.0` and every earlier version tag | `['arm64', 'amd64']` | — |

`docker manifest inspect akitaonrails/ai-memory:latest` returns a plain `application/vnd.docker.distribution.manifest.v2+json` (single image config + layers), not a manifest list — confirming `latest` has no arm64 variant, while `1.28.1` (pushed ~90 minutes earlier in the same release) is properly multi-arch.

This looks like the release pipeline's `latest` tag push ran only on the amd64 build job instead of being created as a manifest-list combining both `amd64` and `arm64` builds (as every versioned tag correctly does).

## Environment

- Host: arm64 (aarch64) Linux, Docker `linux/aarch64`
- `docker pull akitaonrails/ai-memory:latest` → `amd64` image pulled
- Container logs: `exec /usr/local/bin/ai-memory: exec format error`
- Workaround: pinning to `akitaonrails/ai-memory:1.28.1` (multi-arch) resolves the issue on arm64.

## Suggested fix

Ensure the release pipeline publishes `latest` as a manifest list (`docker manifest create`/`docker buildx --platform linux/amd64,linux/arm64 --push`) combining both architecture builds, same as the versioned tags.

akitaonrails:
Confirmed and fixed — thank you for catching this quickly, and sorry for the breakage. **This was my error, not a CI problem.**

[...]

PR #435, opened 2026-08-19
feat(llm): configurable per-request provider timeout (AI_MEMORY_LLM_TIMEOUT_SECS)

lucasliet:
## What changed

- Every chat provider (openai, openai-compat, anthropic, anthropic-oauth, gemini, copilot, openai-oauth, opencode) applied a hardcoded 300s `reqwest` timeout at client construction. Now the timeout is a provider field applied per request (`RequestBuilder::timeout`), overridable end-to-end:
  - new env var `AI_MEMORY_LLM_TIMEOUT_SECS` (or `llm_timeout_secs = 900` in config.toml), read once by `Config::load`
  - flows through `ProviderConfig::request_timeout_secs` and the factory into every provider builder
- The Copilot GitHub token exchange and the shared OAuth refresh grant (`refresh_grant`) are bounded by the same ceiling instead of inheriting an unbounded client.
- **Default unchanged**: 300s (`ai_memory_llm::DEFAULT_REQUEST_TIMEOUT_SECS`), still tolerating a local engine cold-loading a large model.

## Why

Slow hosted gateways — observed with free aggregator tiers streaming long structured completions — exceed five minutes per request. Once generation crossed the ceiling, every request failed with `http: error sending request`, so LLM consolidation exhausted its five retries and left heuristic pages behind, with no operator-level remedy. Raising the bound at startup fixes it without code changes.

## Test plan

- [x] Manual test: ran a deployment against the OpenCode Zen gateway where consolidation of large sessions hit the 300s ceiling and failed with `http: error sending request` in exactly 300.000s; with the timeout raised, the same sessions consolidate successfully.

## Commit attribution

## CHANGELOG (merge gate)

## Notes for reviewers

- Timeout moved from client-level to per-request so the shared `reqwest::Client` (connection pool) stays untouched; per-request `RequestBuilder::timeout` is the idiomatic reqwest mechanism.
- `refresh_grant` is called by both the provider and the CLI's `auth_bearer` with their own clients, so it applies the default ceiling internally rather than changing its public signature.
- Validation: `llm_timeout_secs` fails closed at startup on zero (a sub-second value would sever every request before it is sent).
- The `[#NNN]` reference in the CHANGELOG entry will be updated with this PR's number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- New `llm_timeout_secs` config key (env `AI_MEMORY_LLM_TIMEOUT_SECS`, or
+  `llm_timeout_secs = 900` in config.toml) overrides the per-request timeout
+  every chat provider applies to its HTTP calls — previously hardcoded at
+  300s per provider. Slow hosted gateways (observed with free aggregator
+  tiers whose long completions exceed five minutes) failed every request with
+  `http: error sending request` once generation crossed the ceiling, so LLM
+  consolidation exhausted its retries and left heuristic pages behind; now
+  operators raise the bound at startup instead. The Copilot token exchange
+  is bounded by the same value, and the 300s default is unchanged ([#435]).
 - New `strip_root_combinators` config flag (env `AI_MEMORY_STRIP_ROOT_COMBINATORS`,
   or `strip_root_combinators = true` in config.toml) strips root-level
   `anyOf`/`oneOf`/`allOf` from MCP tool input schemas on every `tools/list`.
--- a/README.md
+++ b/README.md
@@ -966,12 +966,13 @@ automatic PreCompact/PostCompaction checkpoint fall back to the deterministic
 rule-based page; admission, storage, and scope errors still fail closed. The
 validated minimums are 6,000 input and 1,000 output tokens.
 
-Every chat provider bounds each HTTP request at 300 seconds
+Every chat provider bounds each completion request at 300 seconds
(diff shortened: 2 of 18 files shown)


akitaonrails:
Merged via #437 — thank you, and this one needed **no maintainer fixes at all**, which was not true of anything else in today's batch.

Two things I want to call out specifically, because they are the parts people usually get wrong.

**You caught the hazard your own change created.** Moving from a client-level timeout to per-request means any call site that forgets `.timeout()` goes from a 300s ceiling to hanging forever. That was the first thing I went looking for, and `stored_token.rs` already had it bounded with a comment naming the exact reason:

> Bounded here because callers hand us their shared client, whose global timeout was removed in favour of per-request timeouts

Every reqwest call site in the touched providers sets it, including both of Copilot's — the chat call and the token exchange. All eight factory arms pass the value through, and `opencode`/`openai_compat` delegate to their inner provider with a test asserting the delegation lands.

**The default is genuinely unchanged.** 300s matches what every chat provider hardcoded before (anthropic ×2, copilot, openai, gemini, openai_oauth), so nobody's behaviour shifts until they opt in. Embeddings keeping their 120s is right too — different concern, and the `llm_timeout_secs` name says so.

I verified it end-to-end rather than just reading it. Against the live Gemini API: `llm_timeout_secs = 1` gives `operation timed out`, `= 120` on the same request succeeds, and `= 0` is rejected at config load with a message that names the unit. So the value is genuinely applied per request.

The only thing I noted and deliberately did not change: there is no upper bound. It is operator-set rather than attacker-controlled, and capping it would defeat the point of surviving a slow gateway.

Also — thank you for #427. That report was correct, the cause was my own deploy script flattening the multi-arch manifest, and it is fixed with a guard so it cannot recur.

Ships in the next release.

Issue #428, opened 2026-08-19
Expose hook ingestion and backpressure telemetry in status


Parikalp-Bhardwaj:
## Problem

`ai-memory` intentionally uses bounded hook ingestion and backpressure instead of allowing work to accumulate indefinitely.

Native hooks can use a local spool, hook ingestion can become saturated, and requests may be retried when the server cannot accept more work immediately.

This is a useful reliability model, but it is currently difficult for an operator to understand the health of the capture pipeline.

For example, when using `ai-memory` for a long-running project, it can be difficult to tell whether:

* agent hooks are reaching the server normally,
* events are waiting in the local spool,
* retries are occurring frequently,
* the ingestion queue is close to capacity,
* the SQLite writer is falling behind,
* or the system is healthy and processing events normally.

A memory system can continue running while capture is delayed or unhealthy, making these problems harder to notice until much later when expected context is missing.

## Motivation

It would be useful to have a small operational view of the memory ingestion pipeline.

Conceptually, the pipeline looks something like:

```text
Agent
  │
  │ lifecycle hook
  ▼
Native hook
  │
  ├── local spool / retry
  │
  ▼
HTTP hook ingestion
  │
  ▼
bounded ingestion queue
  │
  ▼
sanitization / observation processing
  │
  ▼
writer queue
  │
  ▼
SQLite writer
  │
  ▼
durable observation
```

Today, if memory appears to be missing, an operator has to determine where in this pipeline the problem occurred.

Some basic telemetry could make that much easier.

For example:

* Are hooks currently reaching the server?
* How many hook events have been accepted?
* Has ingestion recently been saturated?
* Are requests being retried?
* Are native hook events accumulating in the local spool?
* How old is the oldest pending spooled event?
* Is the writer queue consistently near capacity?
* Is the SQLite writer keeping up with ingestion?
* When was the last hook successfully persisted?

This would be especially helpful for:

* long-running `ai-memory` installations,
* remote or homelab deployments,
* multiple agents sharing the same memory server,
* debugging missing observations,
* diagnosing temporary load spikes,
* understanding whether memory capture is delayed rather than completely broken.

## Proposed direction

Could `ai-memory status` expose a small set of ingestion and backpressure health indicators?

For example, conceptually:

```text
Hook ingestion
  status: healthy
  accepted_total: 18432
  saturated_total: 3
  last_received: 2s ago
  last_persisted: 2s ago

Local spool
  pending: 0
  oldest_pending_age: 0s
[...]

PR #438, opened 2026-08-19
feat(status): report local hook-spool health

rekcilyssup:
## What

Adds a local hook-spool view to `ai-memory status` so an operator can tell whether lifecycle-hook capture is healthy without digging through logs — the client-side slice of #428 that the maintainer flagged as "nearly free" and immediately useful.

## Changes

- `crates/ai-memory-cli/src/commands/hook_spool.rs`: new `SpoolHealth { pending, oldest_age_ms, retries_total }` snapshot + `spool_health()`. Pending count and oldest-event age are read from the timestamp-embedded spool file names (`{created_ms:013}-{pid}-{seq}.json`) with **zero file reads**; retries sum the per-entry `attempts` field with one bounded read per queued event (bounded by the spool's existing 10k hard cap). The snapshot is deliberately content-free — counts, ages, attempt sums only, never payload excerpts, URLs, or token material (the spool holds private capture until it drains).
- `crates/ai-memory-cli/src/commands/status.rs`: renders a `spool:` block in the human output (`pending` / `oldest` / `retries`, oldest as compact `Xs`/`Ym Zs`/`Hh Mm`/`Dd Hh`) and a `spool` object in `--json` (`pending`, `oldest_age_ms`, `retries_total`).
- 7 new unit tests: empty/missing dir, age + retries accuracy, unparseable-entry tolerance, filename parsing, age-line rendering.
- `CHANGELOG.md` entry under `[Unreleased]` → Added, referencing #428.

## Why this slice

- **No hot-path work**: `status` is operator-invoked; nothing runs on the per-tool-call hook path. No new server state, no counter map, no lock contention — the server-side counters (accepted/saturated/writer-queue) intentionally stay out, per the maintainer's note that they deserve their own design pass.
- **Directly answers "is capture delayed?"**: a non-empty spool or an aging oldest event means hook events are queued locally instead of reaching the server; "oldest event age" distinguishes high throughput (many queued, seconds old) from a stuck pipeline (few queued, minutes old).
- **Client-side by construction**: the spool lives in the local data-dir, so the section reflects local health even when the server is remote.

## Notes

- JSON shape: `"spool": { "pending": 0, "oldest_age_ms": null, "retries_total": 0 }` — `oldest_age_ms` is `null` when the spool is empty.
- Manual verification performed against the running gates below.

## Verification

- `git diff --check`
- `TAILWIND_SKIP=1 cargo test --workspace` (all pass)

Closes nothing — part of #428; the server-side counters remain open for the separate design pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `ai-memory status` now reports local hook-spool health: pending event
+  count, age of the oldest queued event, and total failed-delivery attempts.
+  This makes "memory capture is delayed, not broken" visible to an operator
+  without digging through logs: a non-empty spool or an aging oldest event
+  means hook events are queued locally instead of reaching the server. The
+  spool is client-side, so the section reflects the local data-dir even when
+  the server is remote. The JSON form gains a matching `spool` object
+  (`pending`, `oldest_age_ms`, `retries_total`). (#428)
 - New `strip_root_combinators` config flag (env `AI_MEMORY_STRIP_ROOT_COMBINATORS`,
   or `strip_root_combinators = true` in config.toml) strips root-level
   `anyOf`/`oneOf`/`allOf` from MCP tool input schemas on every `tools/list`.
--- a/crates/ai-memory-cli/src/commands/hook_spool.rs
+++ b/crates/ai-memory-cli/src/commands/hook_spool.rs
@@ -103,6 +103,68 @@ pub fn spool_len(spool: &Path) -> usize {
     list_entries(spool).map_or(0, |f| f.len())
 }
 
+/// Snapshot of local hook-spool health for operator status reporting.
+///
+/// Deliberately content-free: counts, ages, and attempt sums only — never
+/// payload excerpts, URLs, or token material (the spool holds private capture
+/// until it drains).
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize)]
+pub struct SpoolHealth {
+    /// Events queued locally awaiting delivery.
+    pub pending: usize,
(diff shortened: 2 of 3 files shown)


akitaonrails:
Merged via #441 — thank you. Your commit is on `main` verbatim.

This was a well-scoped implementation of the spool half of #428, and three things I specifically went looking for were already right:

- **Content-free by construction.** `SpoolHealth` can only hold `pending`, `oldest_age_ms`, `retries_total` — there is no field a payload could occupy, and you documented that intent rather than leaving it implicit. That was my main concern when I suggested splitting this off.
- **Off the hot path**, with the age read from filenames so reporting never pays a read per queued event just to compute it.
- **Reused `spool_dir`/`list_entries`** instead of re-deriving spool paths.

Two maintainer commits on top.

**Foreign files skewed the numbers.** `created_ms_from_name` returned `0` for any unparseable name and that went straight into `now - created`. A `.json` this queue did not write counted as pending and, if it sorted first, reported the whole Unix epoch. macOS writes AppleDouble sidecars (`._<original>`) on filesystems without extended attributes — they end in `.json` and sort before any digit. Measured before the fix: one real 1-second-old entry plus one sidecar gave `pending=2, oldest_age_ms=1787181461533` — about 56 years. Entries are now filtered by the writer's own naming convention and the parser returns `Option`. Your test asserting `created_ms_from_name("garbage.json") == 0` is what pointed me at it; the value was fine in isolation, it was the downstream subtraction that hurt.

**It was silent exactly when it mattered.** `run` fetched `/admin/status` before computing spool health, so an unreachable server returned early and the section never printed — and the spool exists to buffer while the server is down. The one question it answers could not be asked in the one situation that prompts it. Now read first and printed on the error path, on stderr so `--json` consumers still get a single object on stdout.

Your three original tests are unchanged and still pass, including the one asserting a name-valid but content-invalid file still counts as pending — which is the correct call.

Ships in the next release.

PR #440, opened 2026-08-19
ci: reject duplicate CHANGELOG section headings

akitaonrails:
Caught while cutting v1.29.0 — the release section came out with two `### Fixed` headings, so I stopped and fixed the cause instead of the symptom.

## What happened

I introduced a duplicate `### Fixed` under `[Unreleased]` when merging #429, fixed it in #436, then **reintroduced it in #439** and only noticed because `bin/release` produced a visibly wrong 1.29.0 section. Twice in one day, same mistake, because a repeated heading is valid Markdown and invisible in review.

It matters because `bin/release` copies `[Unreleased]` into the new version verbatim: the release notes ship with the entries split across two identical headings.

## The check

`scripts/check-changelog-sections.sh`, wired into CI as a standalone job. It fails rather than auto-merging — the second author should see the collision, not have it silently resolved.

Verified both directions: passes on the corrected file, fails on a synthetic duplicate.

## It found four pre-existing cases

Not a today problem — the same defect sits in **1.19.0, 1.13.0, 1.4.0, and 1.1.1**, going back months. Those are merged in this commit so the check can be enforced repo-wide rather than needing an exemption list.

Mechanically verified: **478 entries before and after**, every issue reference byte-identical, no entry text altered — only which heading they sit under, and heading order (Keep a Changelog: Added, Changed, Improved, Fixed, Security).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,6 +145,17 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - run: scripts/check-native-packaging.sh
 
+  # `bin/release` copies `## [Unreleased]` into the new version section
+  # verbatim, so a repeated `### ` heading there ships release notes with the
+  # entries split across two identical headings. Valid Markdown, invisible in
+  # review, and it has happened repeatedly.
+  changelog:
+    name: changelog sections
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - run: scripts/check-changelog-sections.sh
+
   # End-to-end docker smoke: packages the release-build binary and confirms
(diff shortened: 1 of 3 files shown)


akitaonrails:
Note on the diff size, since 120/130 looks large for folding five headings.

The fold is deliberately **minimal**: duplicates are merged into their first occurrence and section order is left exactly as it was. My first attempt also re-sorted every section into Keep a Changelog order, which produced a 196/180 diff touching all 92 released sections — noise that would have made this unreviewable. I reverted to the smallest change that fixes the defect.

Integrity verified by comparing every non-heading line before and after:

| check | result |
|---|---|
| content lines added or altered | **0** |
| lines removed | 5 — the blank separators of the 5 removed headings |
| `### ` headings | 166 → 161 (exactly the 5 duplicates) |
| `## [version]` headings | 92 → 92 |

The remaining diff size is inherent to folding: when a duplicate `### Fixed` sits five sections below the first one, merging relocates that whole block upward. No entry text changed and nothing moved between versions.

PR #443, opened 2026-08-20
feat(cli): show local hook-spool backlog in status

Parikalp-Bhardwaj:
## Summary

Adds local hook-spool health information to `ai-memory status`.

The status command can now answer two useful questions:

* How many hook events are waiting in the local spool?
* How old is the oldest pending event?

This makes it easier to tell when capture is falling behind locally, even when the server itself still appears healthy.

## Output

Text output now includes a `hook spool` section:

```text
hook spool:
  pending:        12
  oldest pending: 8s
```

For an empty spool:

```text
hook spool:
  pending:        0
  oldest pending: none
```

JSON output also includes:

```json
{
  "hook_spool": {
    "pending": 12,
    "oldest_age_secs": 8
  }
}
```

`oldest_age_secs` is `null` when the spool is empty.

## Why

Before this change, `ai-memory status` mainly reported server-side state.

That meant the server could look healthy while hook events were accumulating in the local spool because of a slow or temporarily unreachable server.

There was no simple operator-visible signal showing that local capture was delayed.

This PR implements the spool-side part of the observability work discussed in the issue. Server-side ingestion and writer metrics are intentionally left for a separate design pass.

## Implementation

The implementation stays within the existing spool architecture:

* Reuses `hook_spool::spool_len()` for the pending event count.
* Adds `oldest_spool_entry_time()` to determine the oldest queued entry.
* Uses the timestamp already encoded in spool filenames rather than reading or deserializing event payloads.
* Computes and formats the age in the CLI as compact values such as `8s`, `2m`, `1h`, or `4d`.

The oldest entry is selected using the minimum timestamp, so filesystem enumeration order does not matter.

## Privacy

The new status information does not inspect or expose captured content.

It only uses:

* event count
* timestamp metadata
* calculated age

It does not read or print prompts, tool payloads, observations, spool contents, or spool filenames.

## Edge cases

The implementation handles:

* missing or empty spool directories as `pending: 0`
* no oldest entry as `oldest pending: none`
* multiple entries regardless of filesystem ordering
* malformed `.json` filenames without panicking
* timestamps in the future by saturating the reported age to `0s`

Malformed `.json` entries remain included in `spool_len()`, preserving its existing counting behavior, but are ignored when determining the oldest timestamp.

## Compatibility

This is observability-only.

It does not change:

* hook behavior
* spool format
* retry behavior
* server behavior
* ingestion backpressure
* SQLite writer behavior

## Tests

Added coverage for:

### `hook_spool.rs`

* empty/missing spool
* single entry
* oldest entry across multiple files
* malformed filename handling

### `status.rs`

* duration formatting boundaries
* empty spool output
* pending count and oldest age
* future timestamp handling

Tests use the existing production spool APIs with deterministic timestamps, so no sleeps are required.

## Files changed

* `crates/ai-memory-cli/src/commands/hook_spool.rs`
* `crates/ai-memory-cli/src/commands/status.rs`

## Out of scope

As discussed in the issue, this PR intentionally does not add server-side telemetry such as:

* accepted/rejected hook counters
* saturation counters
* writer queue depth
* retry metrics
* Prometheus/OpenTelemetry support
* new health endpoints

Those require a separate design pass.

Related #428

### DIFF
--- a/crates/ai-memory-cli/src/commands/hook_spool.rs
+++ b/crates/ai-memory-cli/src/commands/hook_spool.rs
@@ -103,6 +103,37 @@ pub fn spool_len(spool: &Path) -> usize {
     list_entries(spool).map_or(0, |f| f.len())
 }
 
+/// Enqueue time of the oldest queued spool entry, or `None` when the spool is
+/// missing/empty. Derived solely from the Unix-ms prefix baked into each file
+/// name at [`enqueue`] time — the same zero-padded ordering key [`drain`] sorts
+/// on — so it never opens a spool file. That keeps status telemetry off the
+/// captured payload entirely (a spool body must never surface in status output)
+/// and makes the scan pure metadata: one `read_dir`, no deserialization.
+///
+/// A `.json` file whose name has no parseable millisecond prefix (i.e. one
+/// [`enqueue`] never wrote) has no usable age, so it is skipped here even though
+/// [`spool_len`] still counts it — an unrelated file can make `pending` and the
+/// oldest age disagree, matching `spool_len`'s existing count-every-`*.json`
+/// policy rather than inventing a stricter one.
+#[must_use]
+pub fn oldest_spool_entry_time(spool: &Path) -> Option<SystemTime> {
+    let oldest_ms = list_entries(spool)?
+        .iter()
+        .filter_map(|path| entry_created_ms_from_name(path))
+        .min()?;
+    Some(UNIX_EPOCH + Duration::from_millis(oldest_ms))
+}
+
+/// Parse the leading Unix-ms `created_ms` field from a spool file name
+/// (`<ms>-<pid>-<seq>.json`). Returns `None` for any name that field is absent
+/// or non-numeric in, so a stray file can never poison the oldest-age scan.
+fn entry_created_ms_from_name(path: &Path) -> Option<u64> {
+    path.file_name()
+        .and_then(|name| name.to_str())
+        .and_then(|name| name.split('-').next())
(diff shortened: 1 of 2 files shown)


akitaonrails:
Closing as superseded, not as wrong — and the timing here is unlucky rather than the work.

You filed #428, and #438 from @rekcilyssup implemented the same thing and landed first (merged this morning as #441). Both were good; only one can be the code.

**Worth saying plainly: your parser was the better of the two.** `entry_created_ms_from_name` returns `Option<u64>` and skips names it cannot parse, and your `oldest_spool_entry_time_skips_unparseable_name_without_panicking` test asserts a stray `.json` cannot corrupt the oldest age. The version that merged returned `0` for an unparseable name and fed that straight into `now - created` — so a foreign file reported roughly 56 years of phantom backlog. I had to fix exactly that after merging, and your design would not have had the bug.

One difference remains in the merged version's favour: it also excludes non-conforming files from the `pending` count, where yours counts them (your test asserts `spool_len == 2`). Reasonable either way — a stray file arguably *is* occupying the spool directory — but excluding it keeps every number in the section describing the same set of things.

The shipped feature also gained something neither PR had: it now reports when the server is **unreachable**, which is when a backlog most needs seeing. The spool exists to buffer while the server is down, and computing it after the `/admin/status` fetch meant the one question it answers could not be asked in the one situation that prompts it.

#428 stays open for the server-side half — accepted/rejected counts, ingestion saturation, writer-queue depth, last-persisted timestamp. That part needs a design pass more than an implementation, and given you both filed it and built the client half correctly, a sketch from you on the issue would be welcome.

PR #444, opened 2026-08-20
fix(packaging): bump stale AUR PKGBUILD pkgver to 1.28.0

alvinhui:
## What
The AUR `PKGBUILD` (`packaging/aur/PKGBUILD`) still declared `pkgver=0.3.2`, roughly a hundred releases behind the `1.28.0` that `Cargo.toml`, `CHANGELOG.md`, and the latest git tag actually ship.

## Why it matters
Anyone building the AUR package as currently published would fetch the `v0.3.2` source tarball instead of anything close to current `main`, silently missing every feature and fix shipped since — including behavior, security, and dependency changes.

## Change
- Bump `pkgver` 0.3.2 → 1.28.0 in `packaging/aur/PKGBUILD` (`pkgrel` stays at `1` since this is a version bump, not a rebuild of the same upstream release).
- Add a `### Fixed` entry under `[Unreleased]` in `CHANGELOG.md`, per CONTRIBUTING.md's merge-gate requirement for user-facing/observable fixes.

## Checks
This only touches packaging metadata and the changelog, no Rust source, so `cargo fmt`, `cargo clippy`, and `cargo test` are unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   *wants* one of them kept visible (for example a public bot id) can add it
   to `[sanitize].allowlist`.
 
+### Fixed
+- The AUR `PKGBUILD` (`packaging/aur/PKGBUILD`) still declared `pkgver=0.3.2`,
+  roughly a hundred releases behind the `1.28.0` that `Cargo.toml` and the
+  latest tag actually ship. Anyone building the AUR package as published
+  would fetch the `v0.3.2` source tarball instead of current `main`,
+  silently missing every feature and fix since. Bumped `pkgver` to `1.28.0`
+  to match the current release; `pkgrel` stays at `1` since this is a fresh
+  version bump, not a rebuild of the same upstream version.
+
 ## [1.28.0] - 2026-08-17
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


akitaonrails:
Closing this one, but the observation behind it was fair and it prompted a real fix — see #453.

The in-repo `pkgver` is a placeholder that never reaches a user. `release.yml` copies both PKGBUILDs at publish time and rewrites it from the release tag:

```
sed -i -e "s/^pkgver=.*/pkgver=${version}/" \\
       -e "s/^pkgrel=.*/pkgrel=1/" \\
       -e "s/^sha256sums=('.*')/sha256sums=('${source_sha}')/" \\
       "$source_pkgbuild"
```

The published packages confirm it — checked just now against the AUR RPC:

```
ai-memory        1.29.0-1
ai-memory-bin    1.29.0-1
```

So Arch users have been getting the current version all along. Merging the bump would have written `1.28.0`, which was already two releases behind by the time the PR was opened, and it would go stale again at the next release.

**That said, you were right that the file is confusing.** A PKGBUILD reading `0.3.2` in a repo shipping 1.29.0 looks like neglect, and nothing in the file said otherwise. #453 adds a comment to both PKGBUILDs explaining the placeholder and pointing at the AUR RPC as the way to check what is actually published, so nobody else spends time on this.

Thanks for looking at the packaging — it is the part of the repo fewest people read.

PR #445, opened 2026-08-20
improve(sanitize): head-tail truncation for observation bodies

BobDylans:
## What changed

  Observation bodies that exceed the 16 KiB durable ceiling were truncated **head-only** (`truncate_utf8_bytes`), losing the tail entirely. A 50 KB tool output (e.g. a large `read` call) lost its outcome and any trailing error
context — the LLM consolidator saw an incomplete picture.

  This PR adds `truncate_utf8_bytes_head_tail` and switches `Sanitized::new` to use it. The new function preserves the first and last ~8 KiB with a `[truncated N bytes]` marker between them. The total budget is unchanged
(`OBSERVATION_BODY_MAX_BYTES = 16 KiB`).

  ### Before
```

[head 16384 bytes]…
                     ↑ tail lost entirely

```

  ### After
```

[head ~8124 bytes]
...[truncated 33528 bytes]...
[tail ~8124 bytes]

```

  ## Why

  The LLM consolidator reads observation bodies to produce `concepts/`, `decisions/`, and `gotchas/` pages. If a tool output is 50 KB and only the first 16 KB is stored, the consolidator misses the outcome (success/error) and any
trailing context. Head-tail truncation keeps both ends visible without increasing the storage budget.

  ## Design decisions

  - **New function instead of modifying `truncate_utf8_bytes`**: the existing function is used by `payload.rs`, `assistant_capture.rs`, and others where head-only is the right behaviour. Changing it would affect all those call
sites.
  - **Fallback for small budgets**: when `max < 64` the head-tail split would produce tiny fragments with a marker longer than the content. Falls back to `truncate_utf8_bytes`.
  - **Marker format**: `\n...[truncated N bytes]...\n` — 48 bytes reserved, visible to both humans and LLMs, no ambiguity with the `…` ellipsis used by head-only truncation.
  - **UTF-8 safe**: walks `char_indices()` forward for the head and `.rev()` for the tail, never splitting a code point.

  ## Test plan

  - [x] `cargo test -p ai-memory-core -- sanitize` — all sanitize tests pass
  - [x] `cargo test -p ai-memory-core -- head_tail` — 4 new tests pass:
    - `head_tail_truncation_preserves_head_and_tail`
    - `head_tail_truncation_short_input_unchanged`
    - `head_tail_truncation_small_budget_falls_back_to_head_only`
    - `head_tail_truncation_no_split_on_utf8_boundary`
  - [x] Updated existing test `observation_boundary_caps_after_sanitizing_without_splitting_utf8` — now asserts `TAIL_SENTINEL` survives (previously asserted it was lost).

  ## CHANGELOG (merge gate)

  - [x] Added `### Improved` entry under `[Unreleased]`.
  - [x] No default changed — the budget is still 16 KiB, only the truncation strategy changed.

  ## Notes for reviewers

  - The `TAIL_SENTINEL` test was updated because the old assertion (`!contains("TAIL_SENTINEL")` + `ends_with(…)`) validated head-only behaviour. The new assertion validates head-tail: `contains("TAIL_SENTINEL")` +
`contains("[truncated")`.
  - Other call sites of `truncate_utf8_bytes` (`payload.rs`, `assistant_capture.rs`) are **unchanged** — they cap smaller excerpts where head-only is correct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -158,6 +158,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   file-tool activity without a subsequent Stop produces an advisory to check
   the working tree; and a mid-task exit (Stop without SessionEnd) is flagged
   so the receiver knows the previous session did not finish cleanly. (#425)
+- Observation bodies that exceed the 16 KiB durable ceiling are now
+  truncated with a **head-tail** strategy instead of head-only: the first
+  and last ~8 KiB are preserved with a `[truncated N bytes]` marker between
+  them. A 50 KB tool output previously lost its tail entirely, so the LLM
+  consolidator saw an incomplete picture; the new truncation keeps the
+  outcome and any trailing error context visible without increasing the
+  storage budget.
 
 ## [1.28.1] - 2026-08-18
 
--- a/crates/ai-memory-core/src/lib.rs
+++ b/crates/ai-memory-core/src/lib.rs
@@ -59,6 +59,7 @@ pub use page::{
 pub use routing_snippet::{MARKER_END, MARKER_START, SNIPPET_BODY, find_marker_line, full_block};
 pub use sanitize::{
     OBSERVATION_BODY_MAX_BYTES, SanitizeConfig, Sanitized, Sanitizer, truncate_utf8_bytes,
+    truncate_utf8_bytes_head_tail,
 };
 pub use slots::{
     SLOT_PREFIX, SlotPlacement, SlotVisibility, is_slot_named, is_slot_path, slot_owner,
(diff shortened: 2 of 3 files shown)


akitaonrails:
Merged, thank you. The head/tail split is a real improvement — losing the tail of a long tool output is exactly the case where a summariser ends up with half the picture.

I property-tested the safety invariant before merging, since `OBSERVATION_BODY_MAX_BYTES` is a hard bound and the marker eats into it: budgets 64 through 65536, inputs at 1x/2x/10x/100x the budget, and fillers `a` / `é` / `→` / `😀` to exercise 1-, 2-, 3- and 4-byte code points. Output never exceeded `max` (worst-case headroom 21 bytes) and never split a code point. The 48-byte marker reservation holds even when `omitted` runs to 19 digits.

The `max < 64` fallback to head-only is the right call — below that the marker would be longer than the content it separates.

Issue #446, opened 2026-08-20
Opt-in capture scope: allowlist mode + a switch to skip UserPromptSubmit


F-Lex96:
First: thanks for the honest docs. `docs/research-karpathy-llm-wiki.md` separating what Karpathy actually wrote from community paraphrase, and the marker-file page spelling out exactly what capture retains and what it does not, made this evaluation possible without reading much source. That is rare.
I evaluated v1.29.0 on macOS (native aarch64 binary, Zero-LLM, loopback server) for a setup that runs 3-5 parallel coding-agent sessions across different repositories on one machine. Some of those repositories contain material covered by professional secrecy obligations (in Germany, § 203 StGB for lawyers, tax advisers and insolvency administrators). Breaching that is a criminal matter, not a policy question, so the bar is "structurally cannot be captured", not "is unlikely to be captured".
I had to stop the trial and uninstall. Two properties, both documented and both working as designed, combine into something I cannot configure my way out of.
It writes to `~/.claude/settings.json`, so it takes effect for **every** running and future session on the machine, regardless of repository. There is no way to enable capture for one repository and leave the rest untouched.
Measured: roughly two minutes after `--apply`, observations from six other projects had landed in the shared store under workspace `default`, because those repos had no marker files. That is exactly what the docs say will happen. It is just very fast, and there is no dry-run staging step between "installed" and "capturing everything".
`docs/marker-file.md` is explicit: `UserPromptSubmit` stores its prompt text, and
> Capture exclusions are evaluated only where paths have a proven schema, so they do not claim to filter those other bodies.
That is a reasonable design decision and clearly documented. But for my use case the sensitive content is overwhelmingly **in the prompt**, not in file paths. "Check this claim filing against …" carries the confidential material in the sentence itself.
To be fair to the design, the rest of the capture contract is genuinely narrow, and I verified it: `PreToolUse` retained ~53 bytes per event with no commands, arguments or paths; `PostToolUse` is capped at 2000 bytes; assistant capture is double opt-in and off. The capture surface is tight everywhere except this one field.
Two things, either of which would let me run this:
[...]

PR #448, opened 2026-08-21
feat(install-hooks): add Claude prompt capture opt-out

lihuiyang1024:
## What changed

- Added `install-hooks --agent claude-code --no-capture-prompts` to omit ai-memory's `UserPromptSubmit` hook before prompt text can enter the local spool or wire.
- Added `--capture-prompts` as an explicit way to re-enable prompt capture.
- Preserved the opt-out across bare `--apply` runs and upgrade refreshes by recovering state only from ai-memory-owned hooks.
- Preserved third-party hooks registered under `UserPromptSubmit`.
- Limited both flags to Claude Code because other agents may use their prompt hook to inject handoff context.
- Added CLI parsing, ownership/state recovery, apply/re-apply/re-enable regression coverage, documentation, and a CHANGELOG entry.

The default is unchanged: fresh Claude Code installs still capture user prompts.

## Why

This implements the smaller install-time prompt-capture opt-out requested in #446. Removing the hook client-side provides the structural privacy property described there: sensitive prompt text is never spooled, transported, or stored. Keeping the first version Claude Code-only avoids breaking handoff delivery in clients such as Kimi Code.

## Test plan

- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes — local Windows host lacks the MSVC `link.exe`; CI should run this in the maintained build environment
- [ ] `cargo test --workspace` passes — a targeted CLI test build was attempted but stopped before project-code compilation because `link.exe` is unavailable
- [x] `cargo metadata --no-deps --format-version 1` passes
- [ ] Manual test: local binary build is blocked by the same missing MSVC linker; the new apply/re-apply/re-enable behavior is covered by focused filesystem tests

## Commit attribution

- [x] I verified the name and email on every commit in this PR and corrected the commit to the authenticated GitHub account's noreply address.

## CHANGELOG (merge gate)

- [x] The default is unchanged and is called out explicitly above.

## Notes for reviewers

The benefit of this shape is that the privacy boundary sits before local spooling and transport, and upgrade refreshes cannot silently undo the opt-out. The tradeoff is intentionally visible: without prompt observations, session summaries and recall quality lose direct user-intent evidence. The PR does not attempt the broader allowlist-mode half of #446.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `install-hooks --agent claude-code --no-capture-prompts` can now omit
+  ai-memory's `UserPromptSubmit` hook, preventing prompt text from entering the
+  local spool or wire while leaving session, tool, compaction, and handoff
+  capture active. Re-applying or upgrading preserves the opt-out, third-party
+  hooks under the same event survive, and `--capture-prompts` explicitly
+  enables prompt capture again. Other agents reject both flags because some
+  depend on their prompt hook to inject handoff context. (#446)
 - Windows tests moved to their own workflow
   (`.github/workflows/windows.yml`), running on every push to `main`,
   nightly, on demand, and on any PR labelled `windows`. They were the
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -1629,6 +1629,16 @@ pub struct InstallHooksArgs {
     /// without this flag removes it (idempotent). Default off.
     #[arg(long)]
     pub capture_assistant: bool,
+    /// Do not install Claude Code's `UserPromptSubmit` capture hook. Prompt
+    /// text is then excluded before it can enter the local spool or wire.
+    /// A bare `--apply` re-run preserves an existing opt-out; use
+    /// `--capture-prompts` to enable prompt capture again explicitly.
+    #[arg(long, conflicts_with = "capture_prompts")]
+    pub no_capture_prompts: bool,
+    /// Explicitly enable Claude Code prompt capture after an earlier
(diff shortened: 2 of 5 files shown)


akitaonrails:
Merged via #452 — thank you, and this one matters more than a typical flag.

#446 is from a user who evaluated v1.29.0 under a **criminal-liability** constraint (§ 203 StGB — German professional secrecy for lawyers, tax advisers and insolvency administrators), found prompt capture unavoidable, and uninstalled. Their bar was explicitly *"structurally cannot be captured"*, not *"is unlikely to be captured"*.

**Your implementation meets that bar and a server-side filter would not have.** Removing the `UserPromptSubmit` entry from `settings.json` outright means Claude Code never fires the event, so the text never reaches the spool or the wire. That is the difference between a policy and a guarantee.

**The idempotency handling is the part I want to single out.** The same reporter wrote, independently and before seeing your PR:

> A protection that disappears on upgrade without saying so is worse than no protection.

Your `install_claude_prompt_capture` already solved exactly that — a bare `--apply` re-run recovers the opt-out from the existing install, and recovers it **only from ai-memory-owned hook entries**, so a user's own hooks are never misread as consent and a non-ai-memory install falls back to the default. Your comment names the failure mode outright: *"so `upgrade` cannot silently restore a privacy opt-out."* Getting that right unprompted is the reason this could ship as-is.

One maintainer commit: `render_claude_code` reached eight parameters and tripped clippy's `too_many_arguments` under `-D warnings`. I grouped them into a `ClaudeCaptureScope { assistant, prompts }` rather than silencing the lint — the call site read `true, false`, and those two bools are precisely the privacy-relevant part of the install.

Worth knowing: **no CI ran on your PR** — fork PRs here wait on workflow approval, so the page showed no checks rather than passing ones. That clippy failure surfaced only in the local gate.

I have pointed the reporter of #446 at this so they can test it against their actual threat model.

Ships in the next release.

PR #450, opened 2026-08-21
feat(mcp): serve Gemini/Vertex-safe tool schemas behind an opt-in

lucasliet:
## What changed

Pointing a pass-through MCP client at a Gemini/Vertex model killed the whole
session at `tools/list`:

```
Unable to submit request because `ai-memory_memory_auto_improve` functionDeclaration
`parameters.max_proposals` schema specified other fields alongside any_of.
When using any_of, it must be the only field set.
```

- **New per-request marker `?flavor=gemini`** (alias `?flavor=vertex`) serves
  tool input schemas in the subset Google's `Schema` accepts.
- **New config key `gemini_safe_schemas`** (env `AI_MEMORY_GEMINI_SAFE_SCHEMAS`,
  or `gemini_safe_schemas = true` in `config.toml`) does the same server-wide,
  for clients that cannot carry a query marker. Documented in the generated
  `config.default.toml`.
- In that dialect, `"type": ["integer", "null"]` becomes `"type": "integer"` plus
  `"nullable": true`, and an `Option`-shaped combinator carrying siblings
  (`{"description": …, "anyOf": [{"$ref": …}, {"type": "null"}]}`) merges into its
  lone real branch, parent keys winning so the doc comment survives. It also
  strips root combinators, so it is a strict superset of
  `strip_root_combinators`.
- `docs/mcp-install.md` gains a **Schema dialects for strict upstreams** section
  collecting all three markers, their config keys, and who needs which.

**No changed defaults.** Without the marker or the key, every client receives
byte-identical schemas — `strip_root_combinators`, `?flavor=moonshot`, and
`?flavor=bedrock` behave exactly as before, unions included, and there is a test
pinning that. Runtime argument validation is unchanged in every dialect; the
handlers' "exactly one of `path`/`query`" check and `serde` deserialization remain
the enforcement backstop, as with #412.

## Why

`rmcp` builds tool input schemas with `SchemaSettings::draft2020_12()` and
post-processes nothing beyond the root `title`/`description`, so schemars 1.2's
`allow_null` renders every optional tool argument as a nullable union:

```json
"max_proposals": {
  "description": "Override the maximum validated proposal count for this run.",
  "type": ["integer", "null"], "format": "uint", "minimum": 0
}
```

Google's `Schema` (Vertex/Gemini `functionDeclaration.parameters`) is an
OpenAPI-3 subset: one `type` per field, and `any_of` exclusive with every sibling
key. A converter that forwards our schema verbatim therefore rewrites the union
into `any_of` and copies `description` next to it, and Vertex refuses the whole
request. The field named in the error is arbitrary — Vertex stops at the first
offender, and **every** `Option` argument across all 18 tools has this shape.

Gemini CLI never hits this because it performs the same collapse client-side
before building the functionDeclaration
(`packages/core/src/tools/mcp-client.ts` rewrites a two-element `type` array into
`type` + `nullable`). That is also the evidence for how narrow this fix should
be: Gemini CLI ships `$schema`, `$ref`, `$defs`, `title`, `const`, `default:
null`, and `format: "uint"` to Vertex untouched and works, so none of those are
part of the problem and none of them are rewritten here. Clients with no such
pass — OpenCode on a Vertex model is where this was reported — do hit it.

This is the same shape of problem as #155 (`?flavor=moonshot`) and #412
(`?flavor=bedrock` plus the markerless `strip_root_combinators` opt-in, added
precisely because OpenCode and Cursor never send a marker), so it rides the same
rails rather than inventing new ones. The `bool` toggle becomes an ordered
`SchemaDialect`, letting the operator's configured floor and a request's marker
combine with a plain `max`.

## Test plan

- [x] `TAILWIND_SKIP=1 cargo test --workspace` passes — the only failures are two
      pre-existing `logging::tests` cases that assert a `0o555` directory is
      unwritable and so cannot fail for uid 0; confirmed identical on a pristine
      checkout of `main` in the same container.
- [x] `scripts/check-changelog-sections.sh` passes
- [x] Manual test against a live `ai-memory serve --transport http`, comparing
      `tools/list` across dialects:

      unflavored      {"type":["integer","null"],"format":"uint","minimum":0,…}
      ?flavor=moonshot {"type":["integer","null"],"format":"uint","minimum":0,…}
      ?flavor=gemini  {"type":"integer","format":"uint","minimum":0,…,"nullable":true}
      ?flavor=vertex  {"type":"integer","format":"uint","minimum":0,…,"nullable":true}

      and with `AI_MEMORY_GEMINI_SAFE_SCHEMAS=true` and **no** marker, the same
      collapsed shape, with `jq` reporting `0` union types and `0`
      combinators-with-siblings across every tool's argument schemas.
      `memory_read_page`'s root `anyOf` is present unflavored and absent under
      `?flavor=gemini`.

New tests: unit coverage for each rewrite and for the shapes deliberately left
alone; a dialect-isolation test pinning that the Moonshot/Bedrock path still
leaves unions intact; three stateless-HTTP tests (`?flavor=gemini`, the config
key with no marker, and Moonshot keeping its unions); and a guard that walks
every registered tool under the new dialect and rejects any array `type` or
combinator-with-siblings, so a future optional argument cannot silently
reintroduce the bug.

## Commit attribution

## CHANGELOG (merge gate)

- [x] `CHANGELOG.md` `[Unreleased]` → `Added` entry describing the new key, the
      new marker, the failure it fixes, and that it implies
      `strip_root_combinators`.
- [x] No default changed — called out under "What changed" above.

## Notes for reviewers

Three decisions worth scrutiny:

1. **`nullable: true` rather than dropping the null.** It keeps the schema
   semantically lossless, it is a real field on Google's `Schema`, and it matches
   both Gemini CLI's client-side transform and this repo's own outbound
   `ai_memory_llm::gemini::normalize_nullable_types`. Happy to drop it if you'd
   rather the dialect not emit a non-JSON-Schema keyword.
2. **Known limit, deliberate.** A `$defs` entry can still carry a combinator with
   a sibling `description` — `FeedbackKind` renders as a `oneOf` of `const`
   branches. Collapsing it to `{"type": "string", "enum": [...]}` would be valid
   and arguably nicer, but it drops the per-value docs, and Gemini CLI forwards
   that exact shape to Vertex today without trouble, so it is out of scope here.
   The all-tools guard documents the exclusion at the point it skips `$defs`.
   Say the word and I'll add the `enum` fold.
3. **`install-mcp` appends nothing.** OpenCode is provider-agnostic, so writing
   `?flavor=gemini` into its config would be wrong for most OpenCode users; the
   server-side key is the right lever there. Gemini CLI and Antigravity CLI need
   no marker at all. That also means `uninstall.rs::mcp_url_candidates` needs no
   new candidate, and `install_hooks.rs::hook_server_url_from_mcp_url` already
   strips any query before deriving hook URLs.

No version bump, per the maintenance rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- New `gemini_safe_schemas` config flag (env `AI_MEMORY_GEMINI_SAFE_SCHEMAS`, or
+  `gemini_safe_schemas = true` in config.toml) and a matching `?flavor=gemini`
+  MCP URL marker (alias `?flavor=vertex`) serve tool input schemas in the subset
+  Google's `Schema` accepts. `schemars` renders every optional tool argument as a
+  nullable union (`"type": ["integer", "null"]`), but Vertex/Gemini
+  `functionDeclaration.parameters` takes a single `type` and treats `any_of` as
+  exclusive with every sibling key — so a client that forwards our schema
+  verbatim produced `any_of` with `description` beside it and Vertex failed the
+  whole session at `tools/list` with "specified other fields alongside any_of".
+  The dialect collapses those unions to a single `type` plus `nullable: true`,
+  the same normalization Gemini CLI performs client-side, which is why Gemini CLI
+  and Antigravity CLI never hit this and need no marker; pass-through clients such
+  as OpenCode on a Vertex model did. It implies `strip_root_combinators`, and
+  runtime argument validation is unchanged in every dialect. The key is
+  documented in the generated `config.default.toml`, and `docs/mcp-install.md`
+  now collects all three schema dialects in one table. (#450)
 - Windows tests moved to their own workflow
   (`.github/workflows/windows.yml`), running on every push to `main`,
   nightly, on demand, and on any PR labelled `windows`. They were the
--- a/crates/ai-memory-cli/src/commands/serve.rs
+++ b/crates/ai-memory-cli/src/commands/serve.rs
@@ -533,7 +533,8 @@ pub async fn run(config: &Config, args: ServeArgs) -> Result<()> {
         .with_sanitizer(sanitizer.clone())
         .with_trusted_proxy_identity(trusted_proxy_identity_enabled(&config.auth))
         .with_per_user_slots(config.slots.per_user)
-        .with_strip_root_combinators(config.strip_root_combinators);
+        .with_strip_root_combinators(config.strip_root_combinators)
+        .with_gemini_safe_schemas(config.gemini_safe_schemas);
     if let Some(e) = embedder.clone() {
         server = server.with_embedder(e);
     }
--- a/crates/ai-memory-cli/src/config.rs
+++ b/crates/ai-memory-cli/src/config.rs
@@ -177,6 +177,20 @@ pub struct Config {
     /// `AI_MEMORY_STRIP_ROOT_COMBINATORS=true` or `strip_root_combinators = true`
     /// in config.toml.
     pub strip_root_combinators: bool,
+    /// Serve MCP tool input schemas in the subset Google's `Schema`
+    /// (Vertex/Gemini `functionDeclaration.parameters`) accepts: the nullable
+    /// unions `schemars` emits for every optional argument collapse to a single
+    /// `type` plus `nullable: true`. Vertex rejects the union outright once a
+    /// client forwards it verbatim — "specified other fields alongside any_of"
+    /// — and fails the whole session at `tools/list`. Gemini CLI and
+    /// Antigravity CLI normalize schemas client-side and need nothing; this is
+    /// for pass-through clients such as OpenCode on a Gemini/Vertex model.
+    /// Implies `strip_root_combinators`. Runtime validation is unchanged. Set
+    /// with `AI_MEMORY_GEMINI_SAFE_SCHEMAS=true` or
+    /// `gemini_safe_schemas = true` in config.toml; the per-request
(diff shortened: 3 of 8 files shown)


akitaonrails:
Merged, thank you — and this is the second one from you that needed no maintainer fixes.

I verified the central claim rather than taking it on trust, and the result is worth recording because it is subtler than the PR body suggests.

**On the endpoint I can reach, the broken schema is accepted.** Sending `{"description": ..., "anyOf": [...]}` in a `functionDeclaration` to `generativelanguage.googleapis.com` with an API key returns 200. So this is not reproducible against the Gemini API directly — the failure is specific to Vertex's stricter validation, which needs GCP credentials I do not have.

**But the claim holds, and it is well attested.** The exact error is reported independently across four projects, including OpenCode — the very pass-through client you name:

- [langchain-google #1216](https://github.com/langchain-ai/langchain-google/issues/1216)
- [litellm #11164](https://github.com/BerriAI/litellm/issues/11164)
- [opencode #14509](https://github.com/anomalyco/opencode/issues/14509)
- [opencode-gemini-auth #57](https://github.com/jenslys/opencode-gemini-auth/issues/57)

That makes the opt-in default exactly right. A client that normalizes schemas itself (Gemini CLI, Antigravity) needs nothing, the Gemini API tolerates the union anyway, and only a pass-through client on Vertex is affected — so this should not be on by default, and it is not.

**I also confirmed the collapsed form is accepted**, so enabling it costs nothing on the Gemini API path: `{"type": "string", "nullable": true}` returns 200 for the same call.

Two design choices I want to note approvingly: implying `strip_root_combinators` rather than reimplementing it means #420 and this compose instead of competing, and mirroring `ai_memory_llm`'s existing `normalize_nullable_types` keeps one convention for the same transform on both the inbound and outbound sides.

Gate: fmt, clippy under `-D warnings`, `cargo test --workspace` → 2509 passed, 0 failed.

PR #455, opened 2026-08-21
feat(llm): allow overriding Gemini provider base URL

adhamlucas:
## What changed

* **Before:** `AI_MEMORY_LLM_PROVIDER=gemini` ignored the `LLM_BASE_URL` environment variable (or `llm_base_url` config parameter) and always strictly connected to `generativelanguage.googleapis.com`.
* **After:** `GeminiProvider::new` is dynamically extended via `.with_base_url()` if a custom base URL is supplied in the config.

## Why

To support self-hosted, local Gemini-compatible models. Users hosting models like Gemma 3 offline often have endpoints that strictly speak the Google Generative Language API dialect (e.g. `/v1beta/models/...:generateContent`). Because they do not speak OpenAI's `/v1/chat/completions` dialect, these users cannot use `openai-compat` as a workaround and must use the native `gemini` provider pointing to localhost.

This is especially critical for local setups where Gemma runs with a command like `gemini gemma start`, since `gemini-cli` is still actively maintained and deployed under enterprise plans.

## Test plan

- [x] Manual test: Set `AI_MEMORY_LLM_PROVIDER=gemini`, `AI_MEMORY_LLM_MODEL=gemma3-1b-gpu-custom`, and `LLM_BASE_URL=http://localhost:9379`. Verified via server logs and local endpoint that the request is successfully routed to the local server instead of Google's public endpoint.

## Commit attribution

## CHANGELOG (merge gate)

## Notes for reviewers

I've added two unit tests in `crates/ai-memory-cli/src/config.rs` inside the `mod tests` block:
1. `llm_provider_config_passes_base_url_to_gemini` to ensure that the custom base URL is correctly parsed and passed to the Gemini provider configuration.
2. `llm_provider_config_gemini_uses_default_base_url_when_none_provided` to ensure that omitting `LLM_BASE_URL` leaves the base URL as `None` (falling back to the standard Google API endpoint).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- The `gemini` LLM provider now respects the `LLM_BASE_URL` configuration value.
+  Previously, the base URL was hardcoded to `generativelanguage.googleapis.com`.
+  This allows using self-hosted, local Gemini-compatible models (like Gemma 3 hosted via Litert) that expect the native `/v1beta/models/...:generateContent` payload shape.
+
 ## [1.30.0] - 2026-08-21
 
 ### Fixed
--- a/crates/ai-memory-cli/src/config.rs
+++ b/crates/ai-memory-cli/src/config.rs
@@ -2038,3 +2038,35 @@ mod tests {
         assert_eq!(provider.provider, ProviderChoice::AnthropicOAuth);
     }
 }
+
+#[test]
+fn llm_provider_config_passes_base_url_to_gemini() {
+    let cfg = Config {
+        llm_provider: Some("gemini".into()),
+        llm_base_url: Some("http://localhost:9379".into()),
+        runtime_env: RuntimeEnv {
+            gemini_api_key: Some(SecretString::from("dummy")),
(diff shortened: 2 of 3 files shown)


akitaonrails:
Merged, thank you. This wires up a capability that already existed but was unreachable for Gemini — `GeminiProvider::with_base_url` was defined and never called from the factory.

I verified the override actually takes effect rather than just being stored. Pointing it at a dead local port:

```
AI_MEMORY_LLM_BASE_URL=http://127.0.0.1:9379 ai-memory llm-test --provider gemini ...
→ error sending request for url (http://127.0.0.1:9379/v1beta/models/gemini-3.5-flash:generateContent)
```

The request goes to the custom host with the `/v1beta/models/<model>:generateContent` path preserved, not to `generativelanguage.googleapis.com`. That is what makes this usable against a local proxy or a compatible gateway.

Two details I checked because they were easy to get wrong:

- **It composes with the per-request timeout** from #435/#437 — `with_timeout_secs` is still applied after `with_base_url`, so an override does not silently drop the timeout.
- **It mirrors the `AnthropicOAuth` arm's pattern** rather than inventing a new shape for the same idea, which keeps the factory readable.

Both of your tests are the right pair: one asserting the URL propagates, one asserting the default survives when none is given. Gate: fmt, clippy under `-D warnings`, 2515 passed, 0 failed.

PR #457, opened 2026-08-22
refactor(handoff): group Handoff's 18 fields into 4 sub-structs

Murillofilho86:
## Summary
- `docs/v0.3-roadmap.md` flagged `Handoff`'s flat 18-field shape and deferred grouping until a 17th field showed up ("not urgent... revisit if/when adding a 17th field"). It's now at 18.
- `Handoff` now holds four grouped sub-structs: `HandoffScope` (id, workspace_id, project_id), `HandoffOrigin` (from_session_id, from_agent, to_agent, cwd, owner_user), `HandoffContent` (summary, open_questions, next_steps, files_touched), `HandoffLifecycle` (state, created_at, accepted_by, accepted_at, accepted_by_session, accepted_by_user). Named `HandoffLifecycle` rather than anything with "Acceptance" to avoid colliding with the existing `HandoffAcceptance` input struct.
- **Wire format is unchanged.** `Handoff` serializes directly into the `memory_handoff_accept` MCP response with no DTO — nesting fields could have broken that JSON shape, which under this project's semver policy would require a major bump. Each sub-struct field is `#[serde(flatten)]`ed so the JSON stays exactly as flat as before (only key order shifts, which JSON doesn't treat as significant).
- Added `memory_handoff_accept_response_json_stays_flat` in `ai-memory-mcp/src/server.rs` to lock in the flat wire shape as a regression test.
- `NewHandoff` (insert-side struct) and `HandoffAcceptance` (accept-request struct) are untouched — only `Handoff`'s read-model shape and its direct construction/consumption sites changed, across `ai-memory-store`, `ai-memory-hooks`, `ai-memory-mcp`, and `ai-memory-web`.

## Test plan
- [x] `cargo test --workspace` (all passing, including the new wire-shape regression test)
- [x] `cargo deny check`
- [x] No CHANGELOG entry — internal Rust-only refactor, MCP JSON response unchanged, no new flag/endpoint/tool/behavior.

### DIFF
--- a/crates/ai-memory-core/src/handoff.rs
+++ b/crates/ai-memory-core/src/handoff.rs
@@ -110,15 +110,20 @@ pub struct HandoffAcceptance {
     pub receiving_cwd: Option<String>,
 }
 
-/// Materialised view of a handoff row.
+/// Row identity and tenancy for a handoff.
 #[derive(Debug, Clone, Serialize)]
-pub struct Handoff {
+pub struct HandoffScope {
     /// Stable identifier.
     pub id: HandoffId,
     /// Owning workspace.
     pub workspace_id: WorkspaceId,
     /// Owning project.
     pub project_id: ProjectId,
+}
+
+/// Where a handoff came from and who it's for.
+#[derive(Debug, Clone, Serialize)]
+pub struct HandoffOrigin {
     /// Session that produced this handoff, if any.
     pub from_session_id: Option<SessionId>,
     /// Agent CLI that produced this handoff.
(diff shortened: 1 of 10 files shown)


akitaonrails:
Merged, thank you. Grouping 18 flat fields into four named sub-structs makes the type describe its own shape, and `#[serde(flatten)]` was the right mechanism.

I verified the wire compatibility rather than trusting the claim, because that is the whole risk in this change. Extracting the serde field names from both sides:

```
main:  18 fields
pr457: 18 fields
diff:  (none) — identical set
```

So the serialized JSON is byte-compatible and no sub-struct name leaks into the wire form. Three further things I checked because `serde(flatten)` can bite in ways that compile fine:

- **No `deny_unknown_fields`** anywhere on these types — flatten and that attribute interact badly.
- **`Handoff` derives no `JsonSchema`** and feeds no MCP tool schema, so this cannot disturb the schema work that shipped in #420/#450.
- **No test assertion changed** — your diff only updates construction sites. That was the strongest signal: the tests that verify behaviour still pass unmodified, which is exactly what a behaviour-preserving refactor should look like.

Gate: fmt, clippy under `-D warnings`, 2516 passed, 0 failed.

One note unrelated to your work: no CI ran on this PR at all — fork PRs here wait on workflow approval, so the page showed no checks rather than green ones. The verification above is local.

PR #459, opened 2026-08-22
fix(hooks): drop dead bin/regen-hooks, enumerate bundles in parity test

omartelo:
## What

Delete `bin/regen-hooks`, and make `bundled_posix_and_powershell_hooks_stay_in_parity`
enumerate the bundles under `hooks/` instead of listing them.

`bin/regen-hooks` calls itself "the single source of truth" for `hooks/*/*.sh`
in its own header. It has not been that for months, and running it today is
destructive rather than a no-op.

## Why

**The generator's premise no longer holds.** It templates a flat
`3 agents x 7 events` matrix from two hardcoded maps. What it claims to own is
now 11 agent bundles / 75 POSIX scripts, and they are not interchangeable:

| bundle | events |
| --- | --- |
| `antigravity-cli`, `command-code` | 4 |
| `kiro-cli` | 5 |
| `codex`, `cursor`, `gemini-cli`, `opencode` | 7 |
| `claude-code`, `grok`, `kimi-code` | 9 (`subagent-start` / `subagent-stop`) |
| `devin` | 7, but `post-compaction` instead of `pre-compact` |

The bodies diverged too. Every script now sources `hooks/_lib.sh` (or
`hooks/lib/ai-memory-hook.ps1`), each `.sh` has a `.ps1` peer, and several carry
per-agent behaviour: Claude Code's `session-start.sh` wraps the handoff in
`hookSpecificOutput.additionalContext`, Antigravity gates its startup handoff.

**Running it reverts three bundles.** Executed in a scratch worktree, it
rewrites 21 files across `claude-code`, `codex` and `opencode` back to a
pre-`_lib.sh` shape:

```
28 files changed, 372 insertions(+), 339 deletions(-)
```

Each regenerated script loses:

- the `. "$_lib_dir/_lib.sh"` source and every helper behind it
  (`ai_memory_extract_cwd`, `ai_memory_marker_qs`, `ai_memory_url_encode`)
- the `.ai-memory.toml` marker walk-up
- `cwd` / `workspace` / `project` / `project_strategy` in the query string
- `session_id` on the handoff URL
- Claude Code's `hookSpecificOutput` wrapper, replaced by bare text, which
  reintroduces the "Hook output does not start with {" spam PR #78 fixed
- host-side repo-root resolution from PR #111

It emits no `.ps1` at all, so the regenerated `.sh` files would also drift from
their untouched PowerShell peers and break the parity test.

**Nothing calls it, and nothing documents it.** Not CI, not `bin/release`, not
`scripts/`. `AGENTS.md` inventories `bin/` as "host wrapper scripts
(`ai-memory`, `deploy`, `release`)" and does not list it. The CHANGELOG has no
entry for it in either direction. It was added once and never modified since,
while `hooks/` kept moving.

The contract that actually holds is in
`crates/ai-memory-cli/src/commands/render_shared.rs`. Each event profile
(`CLAUDE_CODE_EVENTS`, `KIMI_CODE_EVENTS`, `KIRO_CLI_V2_EVENTS`,
`KIRO_CLI_V3_EVENTS`, `DEVIN_EVENTS`) already documents the real workflow: add
the entry, add the matching `.sh` and `.ps1`, and the parity test fails if a
bundle drifts. None of them mention running a generator.

Rebuilding the generator to match would mean encoding the twelve event profiles in that file,
Devin's renamed event, the `opencode` directory to `agent=open-code` mapping,
the PowerShell peers, and the per-agent hook bodies. That is `render_shared.rs`
rewritten in bash, which is a second source of truth and the cause of this bug
rather than a fix for it.

**The parity test had the same defect in miniature.** It iterated a hardcoded
list of 9 agent directories, so `command-code` and `kiro-cli` have shipped
uncovered since they landed. Deleting the generator and pointing at that test as
the guard would have traded one silent drift for another, so this PR fixes both.

## How

1. `git rm bin/regen-hooks`.

2. `bundled_posix_and_powershell_hooks_stay_in_parity` now reads `hooks/`,
   keeps every directory except `lib/` (the shared PowerShell helper, not a
   bundle), sorts for deterministic failure messages, and asserts the result is
   non-empty so a wrong path fails loudly instead of passing vacuously. Every
   bundle is covered on arrival, which is the guard the generator was supposed
   to be.

3. `docs/v0.3-roadmap.md` was the only doc still asking for the generator. Its
   entry now records that it was tried and dropped, why the premise broke, and
   what covers the drift instead.

`AGENTS.md` needs no change: its `bin/` inventory is already correct once the
file is gone.

## Test plan

- `TAILWIND_SKIP=1 cargo test -p ai-memory-cli` -> 805 passed, 0 failed,
  2 ignored. The parity test now covers 11 bundles instead of 9.
- Negative check on the new guard: dropped a `hooks/fake-agent/stop.sh` with no
  `.ps1` peer and confirmed it fails, then removed it.

  ```
  assertion `left == right` failed: fake-agent: every .sh hook must have a .ps1 peer
    left: ["stop"]
   right: []
  ```

- `TAILWIND_SKIP=1 cargo clippy -p ai-memory-cli --all-targets -- -D warnings` -> clean
- `cargo fmt --all -- --check` and `git diff --check` -> clean
- `sh tests/hooks/test_lib.sh` -> 44 passed, 0 failed (unchanged, no hook script
  was touched)
- `hooks/` is byte-identical to `main`. The generator run used to measure the
  damage was reverted with `git restore hooks/` and the tree verified clean.

## Notes

Net 127 lines removed. No behaviour change for installed hooks: this deletes a
tool that was never part of the install path and closes a drift class in the
test that guards it.


akitaonrails:
Merged, thank you — and the coverage claim was exactly right, which I verified rather than assumed.

The hardcoded list covered nine bundles; `hooks/` contains eleven:

```
covered:   antigravity-cli, claude-code, codex, cursor, devin,
           gemini-cli, grok, kimi-code, opencode
UNCOVERED: command-code, kiro-cli
```

So both had been shipping with no parity check at all. To confirm the enumeration actually closes that rather than merely looking like it does, I deleted `hooks/kiro-cli/post-tool-use.ps1` and re-ran the test — it now fails with `kiro-cli: every .sh hook must have a .ps1 peer`, naming the previously-invisible bundle. The old list could not have caught that; the string `kiro-cli` did not appear in it.

Your comment gets the failure mode right too: a hardcoded list does not fail when someone adds an agent, it just quietly stops covering it, which is the worst shape for a guard to have.

`bin/regen-hooks` is genuinely dead — the only references were `docs/v0.3-roadmap.md` and its own header comment, and you updated the former in the same change. Nothing in CI, packaging, or another script invoked it.

Gate: fmt, `cargo test --workspace` → 2513 passed, 0 failed.

Issue #107, opened 2026-06-17
macOS: Docker quick-start is non-functional out of the box on Apple Silicon (4 issues)


matheus-rodrigues00:
## Summary

Running ai-memory "as a final user" on Apple Silicon, the **Docker-wrapper quick-start does not produce a working setup for a native agent**. The native binary path works (with two small caveats). Details, evidence, and suggested fixes below. A companion docs PR adds `docs/macos.md` documenting the working path + these limitations.

**Environment:** macOS 26.5.1, Apple Silicon (arm64), Docker Desktop 28.5.2, image `akitaonrails/ai-memory:latest` (reports v1.1.0), release tarball `ai-memory-macos-aarch64.tar.gz` (v1.1.0).

---

## 1. `:latest` is a single-arch `linux/amd64` image (Apple Silicon runs emulated)

**Symptom.** `docker run akitaonrails/ai-memory:latest` on arm64 prints:
`WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8)`. The container runs under emulation.

**Evidence.**
- `docker buildx imagetools inspect akitaonrails/ai-memory:latest` → `MediaType: application/vnd.docker.distribution.manifest.v2+json` (a single manifest, **not** a multi-arch manifest list).
- `docker exec ai-memory uname -m` → `x86_64`.

**Contradicts the docs.** `README.md` (~L208-210), `docs/install.md` (~L30-31), and the support-matrix macOS row state the image is multi-arch / "runs natively on Apple Silicon via the `linux/arm64` manifest."

**Note.** `release.yml` *does* build arm64 for tags (`platforms: linux/arm64`; `docker buildx imagetools create`; per-arch verify), so this looks like a **`:latest`-tag publish gap**, not missing arm64 support.

**Suggested fix.** Publish `:latest` as a multi-arch manifest list including the arm64 image; or correct the docs to say `:latest` is amd64-only and point arm64 users at a multi-arch version tag.

---

## 2. macOS Docker-wrapper thin-client → `403 forbidden host`

**Symptom.** After the README quick-start `docker run`, `ai-memory status` → `Error: server returned 403 Forbidden: forbidden host`.

[...]

PR #464, opened 2026-08-23
fix(wrapper): route Windows helper commands at the host server

AhhYeaah:
## What changed

The native Windows Docker wrapper (`bin/ai-memory.ps1`) defaulted every
helper-container command to `http://127.0.0.1:49374` when
`AI_MEMORY_SERVER_URL` was unset.

- **Before:** with a healthy `docker run -p 127.0.0.1:49374:49374` server,
  `ai-memory status` failed with `Connection refused (os error 111)`, while
  `curl.exe http://127.0.0.1:49374/admin/status` from PowerShell worked. The
  wrapper runs the CLI inside a short-lived helper container, so `127.0.0.1`
  meant *that helper*, not the Windows host.
- **After:** thin-client commands get
  `AI_MEMORY_SERVER_URL=http://host.docker.internal:49374` injected and reach
  the host-published server.

**No changed default for host-side output.** `install-mcp`, `install-hooks`
and `setup-agent` still render `http://127.0.0.1:49374` into host agent
config, because `host.docker.internal` does not resolve on the Windows host —
baking it in would silently break MCP and every capture hook. An explicit
`AI_MEMORY_SERVER_URL` (homelab / remote server) still wins, so those setups
are unaffected.

## Why

This is the same constraint issue #107 fixed for macOS: Docker Desktop gives
Linux containers no host networking on Windows either. The POSIX wrapper's
`Darwin` arm already keys the decision off `RENDERS_HOST_CONFIG`:

```bash
if [ -z "${AI_MEMORY_SERVER_URL:-}" ] && [ "${RENDERS_HOST_CONFIG}" -eq 0 ]; then
  ENV_ARGS+=(-e "AI_MEMORY_SERVER_URL=http://host.docker.internal:49374")
fi
```

The PowerShell wrapper never received it. This PR ports that existing
semantics rather than rewriting the address globally, including the argv scan
that resolves the real subcommand past global options (`--config` and
`--data-dir` consume a value, so their argument is stepped over).

The image-side half already works: `docker/Dockerfile` allowlists
`host.docker.internal` in `AI_MEMORY_ALLOWED_HOSTS` for exactly this reason.

## Test plan

- [x] `cargo test --workspace` passes (`packaging`: 14/14, including the new test)
- [x] Manual test: on Windows 11 + Docker Desktop against a real
      `akitaonrails/ai-memory:latest` container, with `AI_MEMORY_SERVER_URL`
      unset. `ai-memory status` now reports
      `server: http://host.docker.internal:49374` and returns live data
      (1 page, 2 sessions, 100 observations) where it previously failed with
      `Connection refused`. `ai-memory install-mcp --client cursor` still
      renders `"url": "http://127.0.0.1:49374/mcp"` for the host.

New regression test `windows_wrapper_routes_urls_by_real_subcommand` mirrors
the existing `macos_wrapper_routes_urls_by_real_subcommand`, using the same
fake-`docker.cmd` pattern as the current `#[cfg(windows)]` wrapper test. It
covers the three host-config subcommands, a thin client, `search install-hooks`
(routing must follow the real subcommand), global options preceding the
subcommand, and one case the macOS test does not have: an explicit
`AI_MEMORY_SERVER_URL` must win over the alias and still be forwarded by name.

## Commit attribution

## CHANGELOG (merge gate)

- [x] Added a `CHANGELOG.md` `[Unreleased]` / `### Fixed` entry.
- [x] The host-side rendering behaviour is called out explicitly in
      "What changed" above.

## Notes for reviewers

**This PR needs the `windows` label.** Per `AGENTS.md`, PRs touching
platform-sensitive code should opt into `windows.yml` before merge, and the
new test is `#[cfg(windows)]` — it does not compile in `ci.yml` at all. I
cannot add labels as an outside contributor, so could a maintainer apply it?

Two design points worth scrutiny:

1. I duplicated the subcommand scan in PowerShell instead of factoring it out,
   since the two wrappers share no code. The `Get-WrapperSubcommand` arm list
   must stay in sync with the Bash `RENDERS_HOST_CONFIG` case.
2. I kept the emptiness test as `IsNullOrEmpty` to match both the Bash `-z`
   and the neighbouring env-allowlist gate, rather than `IsNullOrWhiteSpace`,
   so a whitespace-only value cannot produce two conflicting `-e` flags.

`bin/ai-memory.ps1` was strictly ASCII with no BOM, and I kept it that way —
Windows PowerShell 5.1 decodes a BOM-less file as ANSI, so I avoided the
non-ASCII characters used freely in the Bash wrapper's comments.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- The Windows Docker wrapper's thin-client commands now reach a
+  loopback-published server instead of failing with `Connection refused (os
+  error 111)`. `ai-memory status`, `search`, `bootstrap` and every other
+  thin-client command runs inside a short-lived helper container, where the
+  CLI's default `http://127.0.0.1:49374` resolves to that helper rather than
+  to the Windows host, so a healthy `docker run -p 127.0.0.1:49374:49374`
+  server was unreachable from the wrapper while `curl` from PowerShell worked.
+  The wrapper now injects `AI_MEMORY_SERVER_URL=http://host.docker.internal:49374`
+  for those commands, matching what the POSIX wrapper has done on macOS since
+  (#107) — Docker Desktop gives Linux containers no host networking on either
+  platform. `install-mcp`, `install-hooks` and `setup-agent` still render
+  `http://127.0.0.1:49374` into host-side agent config, because
+  `host.docker.internal` does not resolve on the Windows host, and an explicit
+  `AI_MEMORY_SERVER_URL` (homelab or remote server) is still honoured. (#464)
+
 ## [1.31.0] - 2026-08-22
 
 ### Fixed
--- a/bin/ai-memory.ps1
+++ b/bin/ai-memory.ps1
@@ -26,6 +26,26 @@ function Get-EnvOrDefault {
     return $value
 }
 
+# Resolve the real subcommand, skipping the global options that may precede
+# it. `--config` and `--data-dir` take a separate value, so their argument has
+# to be stepped over as well or it would be mistaken for the subcommand.
+function Get-WrapperSubcommand {
+    param([string[]]$WrapperArgs)
+
+    $Index = 0
+    while ($Index -lt $WrapperArgs.Count) {
+        $Arg = $WrapperArgs[$Index]
+        if ($Arg -eq "--config" -or $Arg -eq "--data-dir") {
+            $Index += 2
+        } elseif ($Arg -like "--*") {
+            $Index += 1
+        } else {
+            return $Arg
+        }
+    }
+    return ""
+}
+
 $Image = Get-EnvOrDefault "AI_MEMORY_IMAGE" "akitaonrails/ai-memory:latest"
 $Docker = Get-EnvOrDefault "AI_MEMORY_DOCKER" "docker"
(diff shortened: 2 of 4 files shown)


akitaonrails:
Merged via #466 — thank you. Your commit is on `main` verbatim, and the wrapper change needed no adjustment.

Diagnosing the `127.0.0.1` problem correctly is the hard part here: it looks like a server or firewall issue from the outside, and the actual cause — the helper container resolving loopback to *itself* — is only obvious once you know the wrapper shells out to Docker. You also got the half that is easy to miss, which is that `install-mcp`/`install-hooks`/`setup-agent` must **not** receive `host.docker.internal`, because it does not resolve on the Windows host and would silently break MCP config and every capture hook.

I checked both wrappers' exempt-command lists against each other and they are identical, so the two platforms cannot drift on which commands are exempt. `Get-WrapperSubcommand` skipping flags is a good detail too — `--config X status` resolves to `status` rather than reading `--config` as the command.

One thing that happened on the integration branch, unrelated to your change: the Windows job failed on an `autoscope_stress` throughput floor. That is a pre-existing timing sensitivity on that runner, not something you introduced, and it is now fixed plus tracked in #468 — the same test does ~1790 ops/session on a dev host and 1 on `windows-latest`, which is worth its own look.

Worth knowing: **no CI ran on your PR** — fork PRs here wait on workflow approval, so the page showed no checks rather than green ones. I labelled the integration PR `windows` so your change was exercised on a real Windows runner before merging, which is what caught the unrelated flake.

Issue #407, opened 2026-08-17
upgrade does not safely recreate manually created Docker containers


vitorvilas:
#### **Version**

`ai-memory 1.28.0`

#### **Environment**

* **OS:** Linux
* **Docker:** Yes (`akitaonrails/ai-memory:latest`)
* **Agent:** `ai-memory CLI`
* **Transport:** HTTP

---

### **What happened?**

When running `ai-memory upgrade` from an unrelated project directory, two separate issues occurred:

1. **Failure to recreate container:** The image was pulled successfully, but the CLI could not recreate the running Docker container because no `docker-compose.yml` file was found in the current directory or fallback paths. The output instructed stopping and removing the container manually (`docker stop ai-memory && docker rm ai-memory`). Doing this leaves the service down and requires the user to manually reconstruct their initial `docker run` command.
2. **Breaking change on auth / host binding:** After recreating the container with the original parameters, the service entered a restart loop with the following error:
```text
Error: refusing unauthenticated plain HTTP on non-loopback address
0.0.0.0:49374: anyone on the network could access ai-memory.
Configure AI_MEMORY_AUTH_TOKEN or bind to a loopback address.

```

Even though the host binding was explicitly set to `127.0.0.1:49374:49374`, the container internally bound to `0.0.0.0`, triggering a strict security check introduced in the new image version.

---

### **What I expected**

1. **Graceful Container Upgrade:** `ai-memory upgrade` should be able to safely update/recreate standalone Docker containers (created via `docker run`) or print the exact reconstructed `docker run` command used by the existing container without instructing the user to delete it first.
2. **Clear Migration / Security Guidance:** Breaking security changes (such as rejecting unauthenticated requests when bound to `0.0.0.0`) should come with a clear migration warning or prompt during the upgrade process, mentioning `AI_MEMORY_AUTH_TOKEN` or `--allow-insecure-no-auth`.
3. **Data Preservation:** The existing named volume and container configuration should remain preserved and fully functional post-upgrade.

---

### **Steps to reproduce**

1. Run `ai-memory` manually in Docker using a named volume (e.g., `ai-memory-data`) and port mapping `127.0.0.1:49374:49374`.
2. From a directory without a `docker-compose.yml` file (e.g., `/home/example/projects/sample-app`), run:
```bash
ai-memory upgrade

```

3. Follow the output instructions to manually stop and remove the container.
4. Re-run the original `docker run` command with the updated image.
[...]

PR #470, opened 2026-08-23
fix(upgrade): verify compose container ownership

### LINKED ISSUE #469: upgrade trusts an unrelated Compose file and skips standalone container recovery
**Version**
Output of `ai-memory --version`:

```console
ai-memory 1.31.0
```

**What happened**

`ai-memory upgrade` found a running container named `ai-memory` and a
`docker-compose.yml` at a conventional search path:

```text
~/deploy/ai-memory/docker-compose.yml
```

The running container had originally been created as a standalone `docker run`
deployment. It had no `com.docker.compose.*` labels and used the persistent
named volume `ai-memory-data:/data`.

The upgrade wrapper selected the Compose file based only on its location and
ran `docker compose up -d`. Because that Compose project did not own the
existing container, Docker attempted to create another container named
`ai-memory` and failed with a name conflict.

The image pull and hook refresh had already completed, leaving the upgrade
partially applied.

This is a follow-up edge case to #407, which was fixed by #415. That fix added
safe `docker run` reconstruction when no Compose file is found. However, if a
Compose file exists at one of the wrapper's conventional search paths but does
not own the running container, the wrapper assumes it is relevant and skips the
standalone recovery path.

**What I expected**

Before invoking Docker Compose, the upgrade wrapper should verify that the
discovered Compose project actually owns the running `ai-memory` container.

If the container belongs to that project, the wrapper should continue with
`docker compose up -d`. If the container is standalone or belongs to another
Compose project, the wrapper should use the standalone-container recovery path
introduced by #415, which reconstructs the original `docker run` configuration
from `docker inspect`.

Finding a Compose file is not sufficient evidence that its project owns the
running container. The wrapper should not remove or replace the existing
container automatically when ownership is ambiguous, and the persistent
`/data` mount must be preserved.

**Steps to reproduce**
1. Start ai-memory as a standalone container:

   ```console
   docker run -d \
     --name ai-memory \
     --restart unless-stopped \
     -p 127.0.0.1:49374:49374 \
     -v ai-memory-data:/data \
     akitaonrails/ai-memory:latest
   ```

2. Create or leave an unrelated Compose file at one of the paths searched by
   the wrapper, for example `~/deploy/ai-memory/docker-compose.yml`. Have that
   Compose file declare a container named `ai-memory`, while the running
   standalone container remains unmanaged by that Compose project.

3. Run:

   ```console
   ai-memory upgrade
   ```

4. Observe that the wrapper invokes the unrelated Compose project and Docker
   reports that the container name is already in use.

The ownership mismatch can be confirmed before the upgrade:

```console
docker inspect ai-memory --format '{{json .Config.Labels}}'
{}
```

**Environment**
- OS: Linux
- Docker? yes, `akitaonrails/ai-memory:latest`
- Agent: independent of the agent; hooks were installed for Codex, Gemini CLI, and Antigravity CLI
- Transport: HTTP on `127.0.0.1:49374`

**Relevant logs**

The failure is emitted by `ai-memory upgrade`, not by `ai-memory serve`:

```console
→ restarting local ai-memory container via docker compose (/home/vitor/deploy/ai-memory)
[+] up 1/2
✘ service... Error response from daemon: Conflict. The container name "/ai-memory" is already in use by container "2e92586dcc61a3443aeedc3dcad4789cc591efdf0dc5b4b65c8952aa298781aa". You have to remove (or rename) that container to be able to reuse that name.

Error response from daemon: Conflict. The container name "/ai-memory" is already in use by container "2e92586dcc61a3443aeedc3dcad4789cc591efdf0dc5b4b65c8952aa298781aa". You have to remove (or rename) that container to be able to reuse that name.

(compose restart failed; re-run manually: cd /home/vitor/deploy/ai-memory && docker compose up -d)
```

The database was not corrupted. It remained stored in the named volume
`ai-memory-data` and was preserved during manual recovery.

vitorvilas:
## What changed

- `ai-memory upgrade` now verifies that a discovered Compose project owns the running `ai-memory` container before invoking `docker compose up -d`.
- When the container is not owned by that project, upgrade uses the existing inspected `docker run` recreation-script fallback instead of triggering a container-name conflict.
- Added regression coverage to CI and documented the ownership check and data-preserving fallback.

## Why

A standalone `docker run` installation could be mistaken for a Compose deployment solely because an unrelated `docker-compose.yml` existed in a conventional search path. Compose then failed because `/ai-memory` was already in use, and upgrade skipped the safe standalone recovery path.

Fixes #469. Related to #407 and #415.

## Test plan

- [x] Manual test: ran `tests/wrapper_upgrade.sh`; the unrelated-Compose case avoided `compose up -d` and generated a recreation script preserving the `/data` volume, while the Compose-owned case used `compose up -d`.
- [x] `scripts/check-native-packaging.sh` and `scripts/check-changelog-sections.sh` pass

## Commit attribution

## CHANGELOG (merge gate)

## Notes for reviewers

The ownership check compares the inspected target container ID with the IDs returned by `docker compose ps -q`, accepting Docker's abbreviated IDs. If Compose inspection fails or does not identify the running container, upgrade fails conservatively to the existing reviewable recreation script. It does not remove the running container or modify the `/data` volume automatically.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -134,6 +134,7 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - run: scripts/check-native-packaging.sh
+      - run: tests/wrapper_upgrade.sh
 
   # `bin/release` copies `## [Unreleased]` into the new version section
   # verbatim, so a repeated `### ` heading there ships release notes with the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- `ai-memory upgrade` now verifies that a discovered Compose project actually
+  owns the running `ai-memory` container before invoking `docker compose up`.
+  A standalone `docker run` install could previously be mistaken for a Compose
+  deployment merely because an unrelated `docker-compose.yml` existed in a
+  conventional search path; Compose then failed with a container-name conflict
+  and the safe standalone recreation-script fallback was skipped. Unowned
+  containers now use that existing fallback, preserving their inspected ports,
+  mounts, restart policy, command, and operator-set environment. (#469)
 - The Windows Docker wrapper's thin-client commands now reach a
   loopback-published server instead of failing with `Connection refused (os
   error 111)`. `ai-memory status`, `search`, `bootstrap` and every other
--- a/bin/ai-memory
+++ b/bin/ai-memory
@@ -224,6 +224,27 @@ ENVEOF
   return 0
 }
 
+# A compose file found in a conventional location is only relevant when its
+# project actually owns the running container. A stale or unrelated file may
+# declare the same container_name and `compose up` would then fail with a name
+# conflict instead of taking the safe standalone recovery path.
+compose_manages_container() {
+  local container="$1" compose_dir="$2" container_id compose_id
+  container_id="$("${DOCKER}" inspect "${container}" --format '{{.Id}}' 2>/dev/null)" || return 1
+  [ -n "${container_id}" ] || return 1
+
+  while IFS= read -r compose_id; do
+    [ -n "${compose_id}" ] || continue
+    case "${container_id}" in
+      "${compose_id}"*) return 0 ;;
+    esac
+    case "${compose_id}" in
+      "${container_id}"*) return 0 ;;
+    esac
+  done < <((cd "${compose_dir}" && "${DOCKER}" compose ps -q) 2>/dev/null)
+  return 1
+}
+
 cmd_upgrade() {
   if [ -z "${AI_MEMORY_SKIP_SELF_UPGRADE:-}" ]; then
     self_upgrade_script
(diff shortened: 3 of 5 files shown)


akitaonrails:
Merged as `a3e4114` — thank you, and thank you for following your own #407 one level deeper.

This was a genuine gap in the fix I wrote for that issue. #415 handled "no compose file found" by reconstructing the `docker run`, but I never questioned the case where a compose file *is* found: the wrapper selected it on location alone and never asked whether that project owned the running container. Your report is precise about the consequence — the pull and hook refresh complete, then `compose up -d` hits a name conflict, leaving the upgrade half-applied.

Using `docker compose ps -q` for the ownership test is better than what I would probably have reached for. Checking `com.docker.compose.*` labels would infer ownership; asking the compose project what it owns is authoritative, and the two-way prefix match handles short-vs-full container IDs cleanly.

I verified the test is not vacuous rather than trusting it. Run against `main`'s wrapper it fails with:

```
error: …/standalone/output.log does not contain expected text:
       does not manage the running ai-memory container
```

and passes on yours — so it genuinely pins the new behaviour. The negative assertions are the part I appreciated: checking that `compose up -d` is **not** attempted in the standalone case, and that no recreation script is written in the compose-owned case, catches the failure in both directions rather than only the happy one. Stubbing `docker` also means the new CI step needs no daemon, which is why it sits fine in the packaging job.

One mechanical note: GitHub refused the merge through the API because my token lacks the `workflow` scope and your PR touches `.github/workflows/ci.yml`, so I merged it locally and pushed over SSH. Your commit `9f11139` is on `main` verbatim; the PR may show as closed rather than merged as a result.

Gate before pushing: fmt, wrapper syntax, `tests/wrapper_upgrade.sh`, `check-native-packaging.sh`, changelog gate, and `cargo test --workspace` → 2535 passed, 0 failed.

vitorvilas:
Happy it worked out.

Thanks for the attention and for explaining the workflow issue. I’m still building experience with this side of the process, so that was genuinely useful.

I also really appreciate the recognition of the work behind the contribution, as well as the guidance!

Issue #474, opened 2026-08-24
install-mcp has no kimi-code client: Kimi Code CLI registers the plain /mcp URL and every request 400s (missing ?flavor=moonshot)


LeandroCorsoOrion:
## Environment

- ai-memory **1.31.0**, native Windows x86_64 binary, MCP over HTTP (`http://127.0.0.1:49374/mcp`)
- Client: **Kimi Code CLI 1.49.0** (Moonshot), registered with its own modern command:
  `kimi mcp add --transport http ai-memory http://127.0.0.1:49374/mcp`
- `install-hooks --agent kimi-code --apply` works fine — capture was never the problem.

## What happens

Every single model turn fails, not just calls to the affected tool — tool schemas ship with each request, so one rejected schema takes the whole session down. Even a prompt that uses no tools at all:

```
$ kimi --print -p "Responda apenas: 2+2=?"
Error code: 400 - {'error': {'message': "tools.function.parameters is not a valid
moonshot flavored json schema, details: <At path 'root': when using anyOf, type
should be defined in anyOf items instead of the parent schema>", 'type':
'invalid_request_error'}}
```

Isolated by removing servers one at a time: **ai-memory is the cause**, and `memory_read_page` is the single offending tool out of 18. Walking every tool's `inputSchema` recursively finds exactly one `anyOf`:

```
memory_read_page   path=root   type_on_parent=True   items=2   items_without_type=2
```

```json
{ "type": "object", "properties": { ... },
  "anyOf": [ { "properties": { "path":  { "type": "string" } }, "required": ["path"]  },
              { "properties": { "query": { "type": "string" } }, "required": ["query"] } ] }
```

Same root cause as #412 — I am **not** reporting that schema again, it is valid draft 2020-12 and #412 covers it.

## The actual gap

#412 is fixed and the fix works: `restricted_schema_tool_list` strips the root-level `anyOf`, and appending the query parameter makes everything work immediately:

```
kimi mcp add --transport http ai-memory "http://127.0.0.1:49374/mcp?flavor=moonshot"
```

After that, `memory_status` returns real data from Kimi and the session is healthy.

The problem is that **nothing tells a Kimi Code CLI user to do this**:

[...]

PR #476, opened 2026-08-24
feat(agents): add Pool (Poolside Agent CLI) hook capture and finalize-session

stefantalos:
## What

Adds **Pool (Poolside Agent CLI)** as a first-class `AgentKind`: hook-capture attribution, capture-exclusion enforcement, bundled hook scripts (`hooks/pool/`, sh + ps1), `install-hooks --agent pool`, `setup-agent` support, and `finalize-session --agent pool` — following the protocol in `docs/managed-harness-contributions.md` §1–§2.

## Native contract (verified empirically, Poolside CLI v1.0.16, macOS)

- Project hooks in `.poolside/settings.yaml` (`hooks: <Event>: [{name, matcher, command, timeout}]`), Claude-shaped events: SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, Stop. **No true session-end event** → finalize-session is the close path (same class as Codex / Antigravity CLI).
- JSON payload on stdin, snake_case (`session_id`, `cwd`, `prompt`, `tool_name`, `tool_input`); hooks fail-open; tools: shell/read/edit/write/remove.
- SessionStart stdout context injection is **not demonstrated** → `session_start_injects_handoff() = false` (fail-safe: handoff stays available via MCP `memory_handoff_accept`), noted in README.

## Migration

`V50__sessions_pool_agent_kind.sql` copies `V47__sessions_command_code_agent_kind.sql` verbatim (diff-verified identical modulo the `'pool'` token): full table rebuild preserving rows, 5 indexes, 3 scope-pairing triggers. No released migration touched. Forward-migration test included (`v49_to_v50_preserves_session_state_and_accepts_pool`).

## Verification

- `cargo build --workspace`, `cargo clippy -D warnings`, `cargo fmt --check`: clean
- Full workspace tests: **2549 passed, 0 failed** (incl. new roundtrip / capture-exclusion / install-hooks / bundle-parity tests)
- Live smoke: `serve` + pool session-start/post-tool-use hooks (202) → `finalize-session --agent pool` → "Finalized 1 pool session(s)", SQLite row `agent_kind='pool'` with `ended_at` set

## Deliberately out of scope

- Managed workstream (`ai-memory run pool`): Pool's native session-store contract (exact session identity, read-only export) is not demonstrated, so per the contributions doc the limitation is documented instead of shipping a partial adapter.
- `McpClient::Pool`: Pool has no known global MCP client config; stated plainly in README/docs.

Context: we run ai-memory across 7 harnesses (Claude Code, Codex, OpenCode, Cursor, Antigravity, Pool, Vibe) in production; Pool capture already flowed through the generic `/hook?agent=pool` ingestion — this PR closes the attribution + finalize gap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Pool (Poolside Agent CLI) is a first-class agent kind. Hook ingestion
+  recognizes `agent=pool` (alias `poolside`) and Pool's documented snake_case
+  `session_id` / `cwd` / `tool_name` / `tool_input` payload for concrete
+  session attribution and tool-family titles, and Pool's five events
+  (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `Stop` —
+  verified against Poolside CLI v1.0.16) ship as a `hooks/pool/` bundle.
+  `install-hooks --agent pool` stages the scripts and prints a ready-to-paste
+  `.poolside/settings.yaml` `hooks:` snippet — Pool's hook config is
+  project-scoped YAML at each repo root, so the installer deliberately does
+  not write project-local files. Native hook commands enforce `[capture]
+  ignore_paths` for Pool's `read`/`edit`/`write`/`remove` file tools, with
+  unknown payload shapes degrading to metadata-only capture. Pool has no true
+  session-end event, so `ai-memory finalize-session --agent pool` closes the
+  session (the same class as Codex and Antigravity CLI); a forward migration
+  (V50) rebuilds the sessions constraint and pairing triggers to accept the
+  `pool` kind without losing rows. `SessionStart` stdout injection is not
+  demonstrated, so the session-start hook never fetches the single-use
+  handoff — recover it via MCP `memory_handoff_accept`. No managed
+  workstream (`ai-memory run pool`) is claimed: Pool's native session-store
+  contract is not demonstrated, per docs/managed-harness-contributions.md.
 - `ai-memory status` now reports server-side hook-ingestion counters
   alongside the client spool section: events accepted, accepted-but-dropped
   by capture policy, shed because ingest capacity was exhausted, shed by the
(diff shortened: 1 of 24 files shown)


abhisheksharma2411 (commented):
Built and ran the full workspace on this branch: **65 suites, 2549 tests, 0 failures.** Then went after the three things most likely to be wrong in an agent-integration PR — the migration, the wire value, and the hand-written hooks — and two of my three suspicions turned out to be wrong, which is worth saying as plainly as the one that didn't.

## The migration is a faithful copy of the pattern, and I checked rather than assumed

Table-rebuild migrations are where things get silently dropped, so I diffed V50 against V47 statement by statement. Identical: same `DROP TRIGGER`/rebuild/rename order, the same five indexes, the same three triggers recreated after the rename. The only delta is the CHECK list, and it keeps every prior value — `command-code` and `hermes` included — plus `pool`.

`AgentKind::Pool.as_str()` is `"pool"` and the SQL accepts `'pool'`, so the pair that broke Zero can't be broken here. Bumping `ALL` to 20 matters more than it looks, because `begin_session_accepts_every_agent_kind` iterates it against a real database — that test is the thing standing between a future agent addition and the V26-style failure, and this PR keeps it honest.

## Two things I flagged and then had to withdraw

**`ai_memory_extract_cwd` vs `ai_memory_resolve_cwd`.** I expected the `resolve_` variant to be the safer one and pool to be using the wrong helper, which would matter for `ignore_paths`. It isn't: 11 of 12 agents use `extract_cwd`, and `resolve_cwd` exists for devin's `DEVIN_PROJECT_DIR` fallback specifically. Pool is consistent with the norm.

**No `printf '{}\n'` on stdout from `post-tool-use`.** Also fine — the convention is split, and codex, cursor, gemini-cli, kimi-code, kiro-cli and opencode all emit nothing either. Whether an agent needs it is a property of the agent, not the bundle.

Mentioning both because a reviewer's discarded hypotheses are information too, and neither is something you should have to re-defend later.

## One real, narrow finding: `yaml_single_quote` isn't byte-for-byte

The doc comment says:

> Single-quote style is the only YAML form in which the POSIX shell quoting inside the hook command survives byte-for-byte.

The `'` → `''` doubling is right, and it works. But a **newline is silently folded to a space** by any conforming YAML parser, because line breaks inside a single-quoted scalar are folded. Round-tripping the emitted scalar:

```
apostrophe in path     round-trips=true   "/Users/o'brien/bin/hook.sh --url http://x"
newline in path        round-trips=false  "/tmp/we ird/hook.sh --url http://x"
tab in path            round-trips=true   "/tmp/we\tird/hook.sh --url http://x"
```

So an `emit_root` containing a literal newline — legal on every filesystem this runs on — produces a `.poolside/settings.yaml` whose `command:` points at a *different path*, with no error at generation time and no error at parse time. The hook just never fires, and the settings file looks correct to the eye.

This is genuinely rare and I'd not hold the PR for it. But the fix is small, and it's the difference between the comment being true and being nearly true — either reject a path containing a control character at generation, or fall back to a double-quoted scalar with the usual escaping when one is present.

## Two questions, neither blocking

**`timeout: 20`** — is that from Pool's documented schema, or chosen? The rest of the PR is careful to cite (`Poolside CLI v1.0.16, hooks api 1.0`), so I assume it's grounded; worth a word in the comment if it's a judgement call, since a hook that times out is indistinguishable from one that never ran.

**`Stop` as a turn boundary.** The comment is clear that it isn't a session end and `finalize-session --agent pool` is the close path, which matches how Codex and Antigravity are handled. Does anything warn a Pool user who never runs `finalize-session`, or do those sessions just stay open? Not a defect in this PR — I ask because the `install-hooks`/`install-mcp` asymmetry in #474 is the same shape of problem, where the missing half is invisible until something downstream looks wrong.

Really good PR. Deciding that Pool "fails safe like other unproven agents" rather than assuming `SessionStart` injection works, and saying in the CLI help exactly what does and doesn't work, is the part I'd have been most tempted to skip.

PR #477, opened 2026-08-24
feat: write a page summary on both session-page write paths

### LINKED ISSUE #473: Populate frontmatter summary when compiling session pages
**Problem this would solve**

#463 made non-FTS hit descriptors prefer the page's own `frontmatter.summary`,
falling back to a heuristic that reconstructs prose from the body. On a live
1.31.1 instance, that preferred branch never fires: 0 of 25 substantive pages
sampled from one project have `summary` populated, so every descriptor comes
from the fallback.

`summary` *is* written today — but only for the pages the auto-improve
subsystem produces (`admin.rs`, `auto_improve_schedule.rs`). Compiled session
and episodic pages, which are the bulk of a real corpus and the ones the
descriptor problem was about, never get one.

So the heuristic is carrying all the weight, and it is guessing at write-time
information that the compiler already had and discarded.

**Proposed solution**

Populate `frontmatter.summary` when a session page is compiled, in
`ai-memory-consolidate`. #463's heuristic then becomes what it was meant to
be — the fallback for pages written before this, or by hand.

**Alternatives considered**

Two shapes, and the choice is a project-direction call rather than a technical
one, which is why I am asking before writing code:

1. *Deterministic, at write time.* No new dependency, no model call, no new
   failure mode. But it is largely the same heuristic moved earlier, and it
   bakes the guess into the file where a bad one is durable instead of
   recomputed.
2. *Model-generated during consolidation.* Better descriptors, and it uses
   context the compiler has and the reader does not. But it puts a model call
   on the write path, with the latency and failure handling that implies.

Leaving it as-is is also defensible if you consider the fallback good enough.

**Additional context**

Bounding what I actually measured: one instance, one project, 25 substantive
pages, through the public MCP API. It is a corpus with no hand-written
summaries, so it is the worst case for the preferred branch, not the average
one. If summaries are expected to arrive by a path I did not find, this is
moot.

Question: is (1), (2), or neither the direction you want? I am happy to
implement whichever, or to drop it.

samirhvbr:
Closes #473.

Both write paths now fill `frontmatter.summary`, so the branch #463 made
preferred actually fires.

## Zero-LLM path — `synth.rs`

`render_body` already tallied prompts, completed calls per tool, and
start/end, then discarded the structure once it had rendered markdown.
`tally_session` extracts that; the body renderer and the summary builder now
share one tally instead of the summary re-deriving or re-parsing it.
`PostToolUse` remains the only counted kind.

The summary states what the body cannot state about itself — `"2 prompts, 34
completed tool calls across Bash, Edit and Read, over 18m."` A reader deciding
whether to open the page sees how much work it holds, not only what it opened
with.

## LLM path — `ai-memory-consolidate`

Optional `summary` on `ConsolidatedPageUpdate`, `#[serde(default)]`, inserted
by `build_update` only when present and non-blank. No additional model call.

## One correction, and it makes the trap worse than described

You wrote that a summary shaped like a metadata bullet "produces an **empty
descriptor**". It does not — and the real behaviour is worse.

`page_descriptor` ends with:

```rust
if out.is_empty() {
    return truncate_chars(raw.trim(), DESCRIPTOR_MAX_CHARS);
}
```

Every line is dropped, `out` is empty, and the fallback **echoes the raw input
whole**. So the descriptor becomes the malformed summary verbatim. Since the
`COALESCE` already chose that summary over the page body, the page trades a
usable body-derived descriptor for an unusable one. An emptiness check would
not catch it — there is nothing empty to catch. `reader.rs`
`a_metadata_styled_descriptor_source_is_echoed_verbatim` pins exactly this.

Your conclusion stands unchanged: the shape constraint is real, silent, and
worth a test. Only the mechanism differs.

## Tests

- `a_written_summary_changes_what_the_descriptor_tells_the_reader` — the one
  you asked for. Same page, written twice, once with a summary. Asserts the
  descriptors differ, that the body-derived one reaches the prompts (what #463
  already delivers, so the baseline is not strawmanned), that it cannot state
  the totals, and that the summarised one can.
- `a_metadata_styled_descriptor_source_is_echoed_verbatim` — the trap.
- `the_summary_is_shaped_so_the_reader_does_not_drop_it` — asserts each
  property the reader's filter requires, so a future edit to the wording
  cannot silently violate one.
- `a_session_with_nothing_to_report_gets_no_summary` — no key rather than a
  blank one, since a blank still wins the COALESCE.
- Plus content and ordering tests for the tally itself.

## What I did not verify

You asked me to sanity-check the schema change against a small model. **I did
not** — I have no Kimi or qwen3 deployment to run it through, and I would
rather say so than imply coverage I do not have. The field description is
written tight and concrete for that reason, but the risk you flagged on `tier`
is untested here. If you have a small model handy, that is the check worth
running before release.

## Scope

`COALESCE` fallback untouched. New writes only — no backfill, no migration.

Gates: `fmt`, `clippy --workspace --all-targets -D warnings`, `cargo test
--workspace` → 2546 passed / 0 failed, `cargo deny check` → all ok.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,33 +8,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- Session pages now carry a `summary` in their frontmatter, which the
-  retrieval layer already prefers over the page body when describing a
-  non-FTS hit (#473). #463 made that `COALESCE` prefer a summary, but nothing
-  wrote one for a session page: measured against a live 1.31.1 instance, 0 of
-  25 substantive pages had the field, so every descriptor came from the body
-  fallback. Both write paths now fill it — the zero-LLM `SessionEnd` synthesis
-  and the LLM consolidator — because zero-LLM is the documented default and a
-  fix in the consolidator alone would miss the pages most installs have.
+- Gave session pages a `summary` in their frontmatter, which the retrieval
+  layer already preferred over the page body when describing a non-FTS hit.
+  #463 made that `COALESCE` prefer a summary, but no write path produced one:
+  measured against a live 1.31.1 instance, 0 of 25 substantive pages had the
+  field, so every descriptor fell back to the body. All three writers now fill
+  it — the zero-LLM `SessionEnd` synthesis, the single-page consolidator, and
+  the batch one — because zero-LLM is the documented default and `multi_page`
+  defaults to false, so covering either alone would have missed the pages most
+  installs actually hold. (#473)
 
-  On the zero-LLM path the summary is built from counts the renderer already
-  had and was discarding: prompts, completed tool calls per tool, and elapsed
-  time ("2 prompts, 34 completed tool calls across Bash, Edit and Read, over
-  18m"). That is what the body cannot state about itself — a reader deciding
-  whether to open a page can now see how much work it holds, not only what it
-  opened with. The tallying is computed once and shared with the body
-  renderer rather than parsed back out of the markdown, and it still counts
-  `PostToolUse` only, so a completed call is not double-counted.
+  On the zero-LLM path the summary came from counts the renderer already had
+  and was discarding: prompts, completed tool calls per tool, and elapsed time
+  ("2 prompts, 34 completed tool calls across Bash, Edit and Read, over 18m").
+  That is what a page cannot state about itself — a reader deciding whether to
+  open it now sees how much work it holds, not only what it opened with. The
+  page is tallied once and the count lent to both the body renderer and the
+  summary, and `PostToolUse` stays the only counted kind so a completed call is
+  not double-counted.
 
-  On the LLM path it is a new optional field on the consolidator's structured
-  output, so it costs no additional model call — the call already happens.
-  `#[serde(default)]` keeps stored outputs from older runs deserialising, and
-  a blank or whitespace-only value is dropped rather than written, since an
-  empty summary would still win the `COALESCE` and cost the page its
-  body-derived descriptor.
+  On the LLM paths it became an optional field on the consolidator's structured
+  output, costing no additional model call since the call already happens, with
+  `#[serde(default)]` so stored outputs from earlier runs still deserialise.
+  Both system prompts now request it and describe the shape it has to keep; the
+  batch prompt previously forbade the key outright.
 
-  The fallback is unchanged and still serves hand-written pages and
-  everything written before this. New writes only: no backfill, no migration.
+  A model-supplied summary is validated before it is written, because the
+  reader prefers `summary` over the body and then drops headings, metadata
+  bullets, list items and title repeats — falling back to echoing its raw
+  input when nothing survives. A structurally wrong summary is therefore not
+  ignored but reproduced verbatim as the descriptor, displacing usable body
+  text with no error anywhere. Anything the reader would discard is dropped at
+  the boundary instead, leaving the page on its body-derived descriptor.
+
+  The fallback is unchanged and still serves hand-written pages and everything
+  written before this. New writes only: no backfill, no migration.
 
 - `ai-memory status` now reports server-side hook-ingestion counters
   alongside the client spool section: events accepted, accepted-but-dropped
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Session pages now carry a `summary` in their frontmatter, which the
+  retrieval layer already prefers over the page body when describing a
+  non-FTS hit (#473). #463 made that `COALESCE` prefer a summary, but nothing
+  wrote one for a session page: measured against a live 1.31.1 instance, 0 of
+  25 substantive pages had the field, so every descriptor came from the body
+  fallback. Both write paths now fill it — the zero-LLM `SessionEnd` synthesis
+  and the LLM consolidator — because zero-LLM is the documented default and a
+  fix in the consolidator alone would miss the pages most installs have.
+
+  On the zero-LLM path the summary is built from counts the renderer already
+  had and was discarding: prompts, completed tool calls per tool, and elapsed
+  time ("2 prompts, 34 completed tool calls across Bash, Edit and Read, over
+  18m"). That is what the body cannot state about itself — a reader deciding
+  whether to open a page can now see how much work it holds, not only what it
+  opened with. The tallying is computed once and shared with the body
+  renderer rather than parsed back out of the markdown, and it still counts
+  `PostToolUse` only, so a completed call is not double-counted.
+
+  On the LLM path it is a new optional field on the consolidator's structured
+  output, so it costs no additional model call — the call already happens.
+  `#[serde(default)]` keeps stored outputs from older runs deserialising, and
+  a blank or whitespace-only value is dropped rather than written, since an
+  empty summary would still win the `COALESCE` and cost the page its
+  body-derived descriptor.
+
+  The fallback is unchanged and still serves hand-written pages and
+  everything written before this. New writes only: no backfill, no migration.
+
 - `ai-memory status` now reports server-side hook-ingestion counters
   alongside the client spool section: events accepted, accepted-but-dropped
   by capture policy, shed because ingest capacity was exhausted, shed by the
--- a/crates/ai-memory-consolidate/prompts/single_consolidate_system.md
+++ b/crates/ai-memory-consolidate/prompts/single_consolidate_system.md
@@ -100,13 +100,20 @@ schema, and nothing else:
 {
   "title": "<page title>",
   "body_markdown": "<markdown body>",
-  "tags": ["tag-1", "tag-2"]
+  "tags": ["tag-1", "tag-2"],
+  "summary": "<one line of plain prose>"
 }
 ```
 
 - Field names are EXACT and case-sensitive. Use `body_markdown`,
   not `body` or `content`. `tags` may be `[]` but the key must
   be present.
+- `summary` is ONE line of plain prose saying what the page covers, for
+  readers who see it beside the title in a listing. It must NOT be a
+  heading, a `- **key:** value` bullet, a list item, or a repeat of the
+  title — a summary shaped like any of those is discarded and the page is
+  described worse than if you had omitted it. Omit the key rather than
+  guessing.
 
 ## Output format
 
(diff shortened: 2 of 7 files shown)


samirhvbr:
Copilot was right on every substantive point, and three of them meant parts of
this PR did nothing. Fixed in f3398fd.

**The default LLM path was not covered.** `memory_consolidate` defaults to
`multi_page=false` (`server.rs:2501`) and the serve worker calls
`consolidate_session` directly, so the normal route goes through
`ConsolidatedPage` + `build_frontmatter`. I had only touched
`ConsolidatedPageUpdate` / `build_update`, which is the `multi_page=true`
opt-in. This is the same mistake @akitaonrails corrected in #473 — cover the
default, not the flagged path — and I repeated it on the other axis in the
same change. `ConsolidatedPage` now carries the field and `build_frontmatter`
surfaces it.

**The batch prompt forbade the key.** It said ``No `body`, no `content`, no
`summary` `` while the schema had just gained the field, so a model following
instructions would never emit it. Both prompts now request it and describe the
shape, with a test asserting they do — the failure mode here is silence, which
no other test would catch.

**A legal-but-unusable summary is now rejected at the boundary.** This one
sharpens something in the issue thread. The trap does not produce an empty
descriptor: `page_descriptor` ends with

```rust
if out.is_empty() {
    return truncate_chars(raw.trim(), DESCRIPTOR_MAX_CHARS);
}
```

so when every line is filtered the fallback echoes the raw input whole. A
metadata-bullet summary is therefore reproduced verbatim as the descriptor,
and since the `COALESCE` already chose it over the body, the page trades a
usable descriptor for an unusable one. For a deterministic writer a shape test
is enough; for a model-controlled value it is not, and a JSON schema cannot
express "one line of plain prose". `usable_summary` drops anything the reader
would discard, so the page keeps its body-derived descriptor instead.

**The tally was computed twice.** `render_body` and `session_summary` each
called `tally_session`, so every session-page write paid two passes and two
`BTreeMap` allocations — while the doc comment and my commit message both
claimed they shared one. @akitaonrails asked for exactly this not to happen.
The caller computes it once and lends it to both.

**The flaky test is real.** `a_session_with_nothing_to_report_gets_no_summary`
stamped both lifecycle observations with `Timestamp::now()` and asserted a
zero duration; a run straddling a second boundary would have produced
`over 1s` and failed. Fixed timestamps now.

**CHANGELOG** recast past-tense with a trailing `(#473)`, per `AGENTS.md`.

Added the regression coverage that was missing: `build_update` and
`build_frontmatter` each with a usable, a blank, a whitespace-only, a
metadata-bullet, a heading, a list-item, a multi-line and a title-identical
summary.

Gates: fmt, clippy `--workspace --all-targets -D warnings`, `cargo test
--workspace` → 2550 passed / 0 failed, `cargo deny check` ok.

Still not verified, unchanged from the description: I have no Kimi or qwen3
deployment, so the schema change is untested against a small model.

abhisheksharma2411 (commented):
Read this against `page_descriptor` rather than on its own, since `usable_summary` is explicitly a mirror of it, and built + ran the suite locally: **`ai-memory-store` 266, `ai-memory-consolidate` 258, `ai-memory-hooks` 159 — all green, 0 failed.**

The design is right, and the reason given for the guard is the part I'd have missed:

> a structurally wrong summary is not ignored — it is reproduced verbatim as the descriptor, in place of the body text that would otherwise have been used, and nothing reports an error

That's the non-obvious bit. `page_descriptor` echoing its raw input when every line is filtered turns "model wrote a bad summary" into "descriptor is worse than if there were no summary at all", silently. Enforcing at the boundary because a JSON schema can't express it is the correct call.

Also worth saying: the synth path is safe by construction, not by guard — `session_summary` builds from `tally_session` deterministically, so it can't repeat the title or emit a bullet. Sharing one tally between `render_body` and `session_summary` so the two can't disagree is a nice touch.

## One gap: `usable_summary` mirrors the exact-title rule but not the truncated-title one

`page_descriptor` has two ways a line can repeat the title:

```rust
let title_was_truncated = title_trimmed.ends_with('\u{2026}');
...
if title_was_truncated { content.starts_with(title_key) } else { content == title_key }
```

`usable_summary` only implements the second:

```rust
if text == title.trim() { return None; }
```

So when a title carries the ellipsis that `truncate_for_title` adds at 80 chars, a summary that *opens* with the title passes the guard and is then dropped by the reader — landing in exactly the fallback the guard exists to prevent. Probed directly against `page_descriptor`:

```
title    = "Bound the scheduler queue and make the backpressure test deterministic acros…"
summary  = "Bound the scheduler queue and make the backpressure test deterministic across all four workers."

descriptor            = "Bound the scheduler queue and make the backpressure test deterministic across all four workers."
echoes raw verbatim   = true
unrelated summary     = "Two prompts and 34 tool calls over 18m."   (survives normally)
```

The descriptor isn't garbage, which is why it's easy to miss — but it's a descriptor that opens with the entire title, which is the thing #463 set out to remove, and the body-derived text that *would* have carried the prompts is gone.

It's a narrow path: the consolidator's title comes from the model, so it needs a model to emit both an ellipsis-suffixed title and a summary opening with it. Narrow, but the fix is to reuse the same predicate rather than restate half of it:

```rust
let title_trimmed = title.trim();
let title_was_truncated = title_trimmed.ends_with('\u{2026}');
let title_key = title_trimmed.trim_end_matches('\u{2026}').trim();
let repeats_title = !title_key.is_empty()
    && if title_was_truncated { text.starts_with(title_key) } else { text == title_key };
if repeats_title { return None; }
```

Entirely your call whether that's worth it for the frequency — I'd lean yes only because the two copies drifting apart later is the more likely cost, not this specific case firing.

## Two smaller notes, neither blocking

**List markers are rejected here but stripped there.** `page_descriptor` calls `strip_list_marker` and keeps the content; `usable_summary` rejects `- `, `* `, `+ ` outright. Stricter is fine and matches the prompt's instruction — but the doc comment says "anything the reader would discard is dropped at the boundary", and the reader wouldn't discard these, it would strip and use them. Worth a word, since the comment is otherwise precise enough that I trusted it.

**Ordered-list markers slip through.** `strip_list_marker` also handles `1. `, but `usable_summary` doesn't check for it, so `"1. Bounded the queue"` is accepted where `"- Bounded the queue"` is refused. Same outcome for the reader either way, so this is consistency rather than correctness.

## What I checked and found nothing wrong with

- `is_metadata_bullet` requires the `- **` prefix, so a summary like `**session_id:** 9f2c` is kept as content rather than dropped — I went looking for a fallback trigger there and there isn't one.
- Multi-line summaries are refused by the `\n` check before any of the prefix rules, so a summary can't smuggle a heading in on line 2.
- The prompt changes tell the model *why* a bad shape is discarded rather than just forbidding it, which is the difference between a rule a model follows and one it rationalises around.

Nice piece of work — the failure mode you documented is the kind that would have been found six months later by someone wondering why one page's listing looked wrong.

samirhvbr:
@abhisheksharma2411 — thank you, and sorry for the delay. Your review landed eight minutes before
the merge, so it went unaddressed; that is on the timing and on me, not on you. Reading
`usable_summary` against `page_descriptor` instead of on its own is exactly the right way to read it,
and building and running the three suites before commenting is more than I had any right to expect.

I went through all three points against the code. **Two of the three hold, and one of those is worse
than you framed it. The third is narrower than it looks, and I can now say why rather than just
that it is narrow.**

## The ordered marker is the real defect, and it is not only a consistency issue

You filed this as the smaller of your two notes. It is the one that actually degrades a page.
Reproduced against the real filter, with the dashed form beside it:

| model-supplied summary | `usable_summary` | resulting descriptor |
|---|---|---|
| `- vamos revisar o deploy` | **rejects** | `O deploy ficou preso no worker antigo.` — body kept |
| `1. vamos revisar o deploy` | **accepts** | `1. vamos revisar o deploy` |

Same content, opposite outcome, and the one that slips through is the bad one. The path:

1. `usable_summary` (`consolidator.rs:1203-1222`) checks `- `, `* `, `+ ` but not `N. `, and
   `text == title.trim()` is false because of the `1. ` prefix — so it is admitted.
2. The `COALESCE` in `page_descriptor_expr` (`reader.rs:92-98`) then prefers it over the body.
3. `page_descriptor` calls `strip_list_marker` (`reader.rs:110-113`), which **does** handle `N. `,
   gets back the bare title, matches the title-repeat rule (`reader.rs:160-168`), and drops the only
   line it had.
4. `out` is empty, so the fallback at `reader.rs:180-182` echoes the raw input whole.

The descriptor becomes the title with a stray list marker, and the body-derived text it displaced is
gone. So it is not "same outcome for the reader either way" — it is the exact failure the guard was
written to prevent, reached through the one marker the guard forgot.

## The truncated-title half is drift insurance, and here is the mechanism

You called it narrow and said you would lean yes anyway on the drift argument. I agree with your
conclusion, and I can now give the reason rather than a frequency guess: **`usable_summary` is never
reached with a truncated title today.** `truncate_for_title` — the only thing that appends the `…`
your case depends on — lives in `ai-memory-hooks` (`payload.rs:889`, called at `:779` and `:796`),
and pages written on that path get their summary from `session_summary`, which never calls
`usable_summary`. Both call sites (`consolidator.rs:612`, `:1239`) take a **model-supplied** title
off `ConsolidatedPageUpdate` / `ConsolidatedPage`. So reaching it needs a model to emit both an
ellipsis-suffixed title and a summary opening with it.

That makes it unreachable-in-practice rather than rare — which is a better argument for your fix, not
a worse one. Restating half a predicate is a bug waiting for the other half to move, and the cost of
mirroring it is six lines.

## You are right about the doc comment, and it is our text that is wrong

This one I want to acknowledge plainly, because you said you trusted it:

> the doc comment says "anything the reader would discard is dropped at the boundary", and the reader
> wouldn't discard these, it would strip and use them.

Correct. `strip_list_marker` strips the marker and **keeps** the content. Two places state otherwise
and both are mine: the doc comment at `consolidator.rs:1190-1201`, and the test comment at
`consolidator.rs:1663-1668` ("the reader's filter drops every line"), which is accurate for headings
and metadata bullets and wrong for list items. Rejecting list markers at the boundary is still the
behaviour I want — it matches what the prompt asks the model for — but the comment has to say
"stricter than the reader, deliberately" instead of claiming to mirror it.

## On the fix shape

I would take your patch almost as written — mirror the two-branch title rule, add `N. `, fix the two
comments. One caveat on where a shared predicate would live, since it came up: exporting
`page_descriptor` from `ai-memory-store` is not as cheap as it looks. `mod reader` is private with a
curated `pub use` (`store/src/lib.rs:24`, `:52`), so it would be new public API on the store crate,
and `ai-memory-core` is the plausible rival home. That is @akitaonrails's call, not mine.

Since this is about ten lines and the PR is merged, I will carry it as a rider on the next change
that touches `ai-memory-consolidate` rather than opening a PR for it alone — unless @akitaonrails
would rather see it standalone, in which case I will send it now.

Thanks again. The ordered-marker case is the kind of thing that would have sat there for a year
looking like a rounding error.

Issue #386, opened 2026-08-08
Lifecycle-only session creates a fallback handoff and page, and its SessionStart consumes a pending real handoff


gabrielscharb:
**Version**

`ai-memory 1.24.0` (built from `9e5939c`)

**What happened**

A lifecycle-only session — a `SessionStart` followed by a `SessionEnd` with no user prompt and no tool activity in between — causes two separate problems:

**(a) It produces a handoff and a session page from lifecycle events alone.**

In a project with no prior history, one lifecycle-only session yields a handoff whose summary is the no-prompt fallback (`Session ended; 2 observations recorded.`, with `open_questions` and `next_steps` both empty) plus a session page whose stored body is 381 bytes of nothing but the two boundary observations.

**(b) Its `SessionStart` consumes a pending real handoff server-side, before any user work exists.**

`SessionStart` fetches the pending handoff, and that GET is destructive. The code already treats "fetch then discard" as a known silent-loss mode:

```rust
// crates/ai-memory-cli/src/commands/hook.rs:~610
// Only fetch the handoff for agents that inject the session-start
// hook's stdout as context. Grok ignores it, so fetching here would
// consume the handoff server-side (the GET is destructive) and then
// discard the result — silently losing it.
if agent_kind.session_start_injects_handoff() { ... }
```

That guard is per *agent kind*. It cannot cover this case: Claude Code does inject, so it passes the guard — but the session it injects into may never record any work. In my reproduction the handoff moved to `accepted` **8 ms** after the `session-start` observation was persisted, before any prompt could exist.

**What I expected**

- (a) A `SessionEnd` for a lifecycle-only session should not produce a handoff or a session page.
- (b) A pending handoff should not be permanently consumed by a session that turns out to record no work.

**Where the relevant code is**

The `SessionEnd` branch reads observations, writes the page, then builds the handoff. The handoff specifically is gated by `!managed`; the branch overall is *not* ungated — `session_end_disposition()` runs first and can `DropInvalid` / `AlreadyEnded` (`router.rs:2182`), and the admission chain can refuse the handoff afterwards (`router.rs:~2502`). What is missing is any gate on session *content*:

```rust
// crates/ai-memory-hooks/src/router.rs:2442  (SessionEnd branch)
let observations = state.reader.observations_for_session(session_id).await?;
let new_page = synthesize_session_page(ws, proj, session_id, &observations);
// ... write_page ...
[...]

PR #489, opened 2026-08-25
fix(hooks): stop harness scaffolding becoming the session page title (#484)

### LINKED ISSUE #484: Session page titles are unfiltered first-prompt text, so harness scaffolding becomes the page title
`derive_title` takes the first `UserPrompt` observation's title and returns it unchanged. That title
is the raw prompt payload cut at 80 characters. Nothing filters it, so anything the harness puts in
the prompt payload — IDE context blocks, a shell prompt echoed into a paste, a model id — becomes the
title the page is indexed and displayed under.

`page_descriptor` beside it has been fixed twice (#463, #477). `derive_title` has not been touched
since `747c32d` ("M3: hooks crate + auto-capture"): `git log -S "fn derive_title" -- crates/ai-memory-hooks/src/synth.rs`
returns exactly one commit.

## What I measured

Live instance on 1.32.0. **All 1,835 session pages** on disk, 158 projects — not a sample.

The corpus splits in two and they have to be counted separately:

| population | pages | share |
|---|---|---|
| lifecycle-only (exactly 2 observations — the #386 population) | 1,505 | 82.0% |
| **real sessions (>2 observations)** | **330** | 18.0% |

Every lifecycle-only page has a non-content title, but that is a symptom of being a stub, not this
defect, so I excluded them. Among the 330 real sessions:

**21 of 330 (6.4%) carry a title that is not user content.**

| class | n | example |
|---|---|---|
| IDE context block | 16 | `<ide_opened_file>The user opened the file /home/samir/x/…` |
| shell prompt echo | 4 | `┌─[samir@samirb3 12:56:36] ~/x/ai-usagebar/gnome-extension` |
| model id | 1 | `claude-opus-5[1m]` |

**It is still producing**, not a legacy population: the two most recent are 2026-08-21 and
2026-08-22, and the rate is flat — 12 of 140 (8.6%) in July, 9 of 187 (4.8%) in August.

## Two distinct causes

1. **The prompt payload is taken verbatim.** `payload.rs:776-779` —
   `extract_content(raw, &["prompt","message","text"]).map(truncate_for_title)`. `truncate_for_title`
   (`payload.rs:889-899`) only cuts at 80 chars and appends `…`. There is no filter of any kind.
   `synth.rs:66-78` then returns that title unchanged.
2. **A `SessionStart` can supply the title.** `payload.rs:772` —
   `HookEvent::SessionStart => extract_string(raw, &["model", "title"])`. When no `UserPrompt`
   observation carries a title, `derive_title` falls through to any observation's title, and a model
   id is what is there. (`extension_title_hint` at `payload.rs:797` ends
   `.unwrap_or_else(|| source_event.to_string())`, which is the same shape for extension events.)

## The predicate already exists — in this repo, on the adjacent path

`crates/ai-memory-workstream/src/transcript.rs:1833`:

```rust
AgentKind::ClaudeCode => trimmed.starts_with("<system-reminder>"),
```

and `:1843` also drops `<user_info>`. The comment immediately above it carries your own measurement:

> A single session's reminders measured 42KB against 270 bytes of real input, so importing them both
> leaks harness internals into the portable ledger and evicts real conversation from the startup
> packet budget.

So the product already knows these wrappers are not user content — on the opt-in `ai-memory run`
path — and does not know it on the hook path, which is the documented default every install runs.

## Where a bad title lands

- **FTS column 0.** `V17__pages_fts_index_path.sql:32-33` — `pages_fts(title, body, path_search)`.
- **The reranker's candidate text.** `server.rs:1658-1665` copies `hit.title` into `RerankCandidate`.
- **The input to `page_descriptor`'s title-repeat filter.** `reader.rs:158-165` suppresses a body line
  that repeats the title; it cannot help when the title itself is the noise.
- **The `duplicate` lint key.** `lint.rs:317-322` lowercases `frontmatter.title`.

One thing that raises the stakes: on an install with no LLM and no embedding provider — the
documented default — FTS is the only stream that can score anything. My instance's `/admin/status`
reports `embedding_rows: 0` and `links_from_latest_pages: 0` over 2,364 pages, and across a dozen
explained queries every page hit scored on `rrf.fts` alone, with `vector`, `entity` and `graph` at
0.0 on every one. The title is not one signal among four there; with `body`, it is the index.

## What I am asking

Two shapes, and the choice is yours:

- a shared scaffolding predicate in `ai-memory-core`, used by both writers — `ai-memory-hooks` does
  not depend on `ai-memory-workstream`, but both depend on `ai-memory-core`, which already hosts text
  helpers in `sanitize.rs`; or
- a hooks-local filter, leaving `transcript.rs` alone.

I have not written code. Tell me which one and I will send it with tests.

## What this does not fix, stated up front

- Like #477, a filter is **new-writes-only**. The 6.4% is stock; the fix changes flow. I am not
  proposing to rewrite existing titles.
- The `session-start` half is a different problem and probably not this one. Nearly all of those on
  my instance are the 2-observation pages #386 closed at the source — the most recent is 2026-08-13,
  one day after that merge, and none in the 12 days since. That population is historical debt on old
  instances, not a live writer defect, and I would leave it out of scope unless you disagree.
- I have one corpus. The three classes I found are what my agents happen to emit; another harness may
  produce different scaffolding, which is an argument for a shared predicate over a hardcoded list.

akitaonrails:
Closes #484. Reported by @samirhvbr.

## The defect

`derive_title` returns the first user prompt unchanged. A prompt payload is whatever the harness put there, so IDE context blocks, an echoed shell prompt, and a bare model id all became the title the page is **indexed and displayed under**. `page_descriptor` sitting next to it has been fixed twice (#463, #477); `derive_title` had not been touched since `747c32d`.

Measured across a live instance — all 1,835 session pages, not a sample — **21 of the 330 real sessions (6.4%)**, at a flat rate month over month (8.6% July, 4.8% August) rather than a decaying legacy population. The 1,505 lifecycle-only stubs were excluded as #386's population: a symptom of being a stub, not of this defect. That split is what makes the number actionable — 21 of 1,835 would have been rounded away.

## Shape, not a blocklist

The predicate asks *"does this read as something a person wrote?"* rather than matching known offenders. Three classes from one corpus is a sample of **that operator's harnesses**, not of the problem; a list is wrong on the fourth harness and fails silently, because a bad title just looks like a title.

It lives in `ai-memory-core` so the workstream path — which already drops `<system-reminder>` and `<user_info>` with a narrower version of the same idea at `transcript.rs:1833` — can share it later.

**It errs toward keeping text.** A false positive silently discards a real title; a false negative only reproduces today's behaviour. So `help` and `continue` survive as titles, and `< 5ms is fine for this path` is prose, while `claude-opus-5[1m]` is not — no whitespace anywhere *and* identifier punctuation.

## Fallback order

Usable prompt → any usable observation title → the session's own identity. The literal `"session"` now survives only for a session with no observations at all: collapsing every filtered page onto one shared string trades a bad title for an ambiguous one.

## Tests

- **The three measured classes**, each asserted to yield the *next real prompt* instead.
- **All-scaffolding session** — the case the issue asked for explicitly; asserts it does not collapse to `"session"`.
- **Descriptor interaction** — the title is what `page_descriptor` filters body lines against, so changing which string becomes the title changes which lines survive. Asserts the promoted prompt still appears in the body, and that the *skipped* one is still recorded, since it remains session history.
- Predicate unit tests including false-positive guards.

Control verified: disabling the filter makes `harness_scaffolding_does_not_become_the_title` fail with the `<ide_opened_file>` case.

## Scope

New writes only. The existing 21 are not rewritten, and the `session-start` half is deliberately out of scope — per the reporter's evidence its most recent instance is one day after #386 merged with none in the twelve days since, so that is closed at the source.

`cargo fmt` ✅ · `clippy -D warnings` ✅ · `cargo test --workspace --all-targets` → **2580 passed, 0 failed**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Session page titles were the first user prompt taken verbatim, so whatever
+  the harness put in that payload became the title the page is indexed and
+  displayed under — IDE context blocks, a shell prompt echoed into a paste, a
+  bare model id. Measured across one live instance: 21 of 330 real session
+  pages (6.4%), at a flat rate month over month rather than a decaying legacy
+  population. Titles that read as harness scaffolding are now skipped in
+  favour of the next real prompt, and a session whose every candidate is
+  scaffolding falls back to its own identity rather than a shared literal.
+  The predicate tests the *shape* of the text rather than matching a list of
+  known offenders, since three classes from one corpus is a sample of that
+  operator's harnesses and a blocklist would fail silently on the fourth.
+  New writes only — existing titles are unchanged. Reported by @samirhvbr
+  (#484).
+
 ## [1.32.0] - 2026-08-24
 
 ### Added
--- a/crates/ai-memory-core/src/lib.rs
+++ b/crates/ai-memory-core/src/lib.rs
@@ -15,6 +15,8 @@ pub mod ids;
 pub mod observation;
 pub mod page;
 pub mod routing_skills;
+pub mod scaffolding;
+pub use scaffolding::looks_like_scaffolding;
 pub mod routing_snippet;
 pub mod sanitize;
 pub mod slots;
--- a/crates/ai-memory-core/src/scaffolding.rs
+++ b/crates/ai-memory-core/src/scaffolding.rs
@@ -1,0 +1,134 @@
+//! Does a candidate title look like harness scaffolding rather than something
+//! a person wrote? (#484)
+//!
+//! Session page titles come from the first user prompt, taken verbatim. But a
+//! prompt payload is whatever the harness put there, and harnesses inject:
+//! IDE context blocks, a shell prompt echoed into a paste, a bare model id.
+//! Measured across one instance's 330 real session pages, 6.4% carried a
+//! title that was not user content, and the rate was flat month over month
+//! rather than decaying — a live writer defect, not legacy data.
+//!
+//! Deliberately a **shape** test, not a list of known offenders. Three
+//! classes observed on one corpus is a sample of that operator's harnesses,
+//! not of the problem; a blocklist is wrong on the fourth harness and fails
+//! silently, because a bad title just looks like a title. `transcript.rs`
+//! already carries a narrower version of this idea for `<system-reminder>`
+//! and `<user_info>` on the workstream path.
+
+/// Characters that open a decorated shell prompt rather than prose.
+/// Box-drawing and the powerline separators commonly echoed into a paste.
+const PROMPT_DECORATION: [char; 9] = [
+    '\u{250c}', // ┌
+    '\u{251c}', // ├
+    '\u{2514}', // └
+    '\u{2500}', // ─
+    '\u{2502}', // │
+    '\u{256d}', // ╭
+    '\u{2570}', // ╰
+    '\u{279c}', // ➜
+    '\u{276f}', // ❯
+];
+
+/// `true` when `candidate` reads as harness scaffolding rather than user
+/// prose, and so should not become a page title.
+///
+/// Errs toward keeping text: a false positive silently discards a real title,
+/// while a false negative only reproduces today's behaviour.
+#[must_use]
+pub fn looks_like_scaffolding(candidate: &str) -> bool {
+    let trimmed = candidate.trim();
+    if trimmed.is_empty() {
+        return true;
+    }
+
+    // A markup/context block: `<ide_opened_file>…`, `<system-reminder>…`,
+    // `<user_info>…`. Requires a closing bracket so a bare comparison like
+    // "< 5ms is fine" stays prose.
+    if let Some(rest) = trimmed.strip_prefix('<')
+        && let Some(tag) = rest.split('>').next()
+        && !tag.is_empty()
+        && tag.len() <= 64
+        && tag
+            .chars()
+            .all(|c| c.is_ascii_alphanumeric() || c == '_' || c == '-' || c == '/')
+        && rest.contains('>')
+    {
+        return true;
+    }
+
+    // A decorated shell prompt echoed into the paste.
+    if trimmed.starts_with(PROMPT_DECORATION) {
+        return true;
+    }
+
+    // A bare identifier rather than a sentence: no whitespace anywhere, and
+    // carrying identifier punctuation. Real prompts essentially always
+    // contain a space; requiring the punctuation too keeps a one-word prompt
+    // like "help" or "continue" as a legitimate title.
+    if !trimmed.contains(char::is_whitespace)
+        && trimmed
+            .chars()
+            .any(|c| c.is_ascii_digit() || c == '[' || c == ']')
+        && trimmed.chars().any(|c| c == '-' || c == '_' || c == '[')
+    {
+        return true;
+    }
+
+    false
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn the_three_observed_classes_are_scaffolding() {
+        // From the #484 corpus survey: 16 IDE blocks, 4 prompt echoes, 1
+        // model id across 330 real session pages.
+        assert!(looks_like_scaffolding(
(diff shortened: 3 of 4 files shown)


samirhvbr:
You asked for eyes on the PRs rather than a rebuild, so: I read all three. #487 and #488 are right,
and on #488 you improved my proposal — I had not considered that keying the hydrate on an empty
title is what keeps FTS's term-centred `snippet()` from being downgraded to a generic descriptor.
"Hydrate everything" is exactly the simplification I would have written, and the test you pinned it
with is the one I would have been missing.

One finding on this one. I checked it by compiling `looks_like_scaffolding` and running it rather
than by reading it, because the branch in question is the kind that reads correctly.

## The identifier branch cannot separate a model id from a terse user reference

The branch fires on *single token + contains a digit or bracket + contains `-`, `_` or `[`*. Run
against both populations:

| model ids (want `true`) | | terse user references (want `false`) | |
|---|---|---|---|
| `claude-opus-5[1m]` | ✅ true | `pr-477` | ❌ **true** |
| `claude-fable-5` | ✅ true | `issue-484` | ❌ **true** |
| `gpt-4o-mini` | ✅ true | `fix-484` | ❌ **true** |
| `gpt-5` | ✅ true | `commit-a0bc43d` | ❌ **true** |
| `o3-mini` | ✅ true | `main-2.rs` | ❌ **true** |
| `gemini-2.5-pro` | ✅ true | `step-1` | ❌ **true** |
| `qwen3-coder` | ✅ true | `task_7` | ❌ **true** |
| `llama-3.3-70b` | ✅ true | `branch-x` | ✅ false |
| `deepseek-v3` | ✅ true | `v1.32.0` | ✅ false |
| `kimi-k2` | ✅ true | `README` | ✅ false |

**10 of 10 model ids caught, and 7 of 11 terse references caught with them.** That is not a tuning
problem. `gpt-5` and `pr-477` are the same string shape; the difference between them is semantic,
and no lexical rule reaches it. This is the one branch where the predicate errs toward *discarding*
text, against the principle stated three lines above it.

The `<tag>` branch has a smaller version of the same thing: a first prompt of `<div>` or `<T>` is
classified as scaffolding. Rare enough that I would leave it.

## The information the lexical test threw away

The model-id class has a known structural origin. `payload.rs:772`:

```rust
HookEvent::SessionStart => extract_string(raw, &["model", "title"]),
```

So a model id does not arrive as an anonymous string — it arrives because a `SessionStart`
observation's title was populated from the harness's `model` field, and `derive_title`'s second loop
accepts any observation's title including that one.

I checked both model-id titles in my corpus on disk rather than assuming it, and both came that way.
The one inside a real session is the clearer of the two — 17 observations, and its raw block opens:

```
- `session-start` @ 2026-08-21T19:50:41Z — claude-opus-5[1m]
- `user-prompt`   @ 2026-08-21T20:07:11Z —
- `pre-tool-use`  @ 2026-08-21T20:07:18Z — tool non-file
```

The `user-prompt` carried an **empty title**, so the first loop skipped it on `!obs.title.is_empty()`
and the second loop took the `SessionStart`. The prompt was never scaffolding; it was absent.

So the class can be handled where it is unambiguous instead of guessed at afterwards — either skip
`ObservationKind::SessionStart` in the fallback loop, or stop `best_title_hint` using `model` as a
title hint at all, which is a weaker hint now that `Session {id}` exists as the last resort.

**Doing that lets the identifier branch be deleted.** Every model id above is still caught, by
construction rather than by shape, and all seven false positives go away with it. The predicate then
carries only the two branches that test genuine shape — a markup block and a decorated prompt —
which is also a cleaner thing for the workstream path to share later.

One gap in that, so it is not oversold: `extension_title_hint` (`payload.rs:794`) also reads `name`,
so an extension-sourced event whose `name` happened to be a model id would not be covered by a
`SessionStart` guard. I have never seen one and cannot show you an example — flagging it as the
place the structural argument stops being airtight, not as a known case.

## What I cannot tell you

How often a terse identifier is actually somebody's first prompt. My 21 cases were all IDE blocks,
prompt echoes and the one model id — no terse reference appeared, so I have no rate for this and it
is a code-reading finding, not a measured one. The consequence is bounded either way: a false
positive falls through to the next usable title and then to `Session {id}`, so the page gets a worse
title, never a broken one. If you would rather keep the branch until someone actually loses a title
to it, that is a defensible call and this comment is on the record either way.

Happy to send the change if you want it — it is a deletion plus one guard.

PR #490, opened 2026-08-25
fix(hooks): exclude the model id by kind, not by identifier shape

samirhvbr:
Follow-up to #489, as offered in the review there.

The identifier branch of `looks_like_scaffolding` fires on *a single token containing a digit or
bracket, plus a `-`, `_` or `[`*. That is the shape of a bare model id. It is equally the shape of a
terse human reference, and the two are not separable by any lexical rule — `gpt-5` and `pr-477`
differ only in meaning.

Compiled the shipped function and ran it against both populations:

| model ids (want `true`) | | terse references (want `false`) | |
|---|---|---|---|
| `claude-opus-5[1m]` | true | `pr-477` | **true** |
| `claude-fable-5` | true | `issue-484` | **true** |
| `gpt-4o-mini` | true | `fix-484` | **true** |
| `gpt-5` | true | `commit-a0bc43d` | **true** |
| `o3-mini` | true | `main-2.rs` | **true** |
| `gemini-2.5-pro` | true | `step-1` | **true** |
| `qwen3-coder` | true | `task_7` | **true** |
| `llama-3.3-70b` | true | `branch-x` | false |
| `deepseek-v3` | true | `v1.32.0` | false |
| `kimi-k2` | true | `README` | false |

**10 of 10 model ids caught, and 7 of 11 references caught with them.** That branch is the one place
the predicate errs toward *discarding* text, against the principle stated three lines above it.

## What this does instead

`best_title_hint` fills a title from the harness's `model` field at exactly one site, and that site
is gated on the kind:

```rust
HookEvent::SessionStart => extract_string(raw, &["model", "title"]),
```

Every other arm reads `prompt`/`message`/`text`/`tool`/`summary`. So `derive_title` now skips
`ObservationKind::SessionStart` as a title source and the identifier branch is deleted. The
predicate is left with the two branches that test genuine shape, and all seven false positives go
with the branch.

## What is given up, since it is not nothing

**This is narrower than what v1.32.1 shipped.** That version kept a model id out of a title from any
position; this keeps it out of the position it actually arrives in. A model id somebody *typed as a
prompt* is now kept — deliberately, because at that point it is what the user wrote and it is
indistinguishable from `pr-477`. Concretely, against the fixture in
`an_all_scaffolding_session_falls_back_to_its_identity` as merged:

```
[UserPrompt "<ide_opened_file>x", UserPrompt "claude-opus-5[1m]"]
  v1.32.1:  "Session {id}"
  this PR:  "claude-opus-5[1m]"
```

The loss is unreachable through any shipped harness — no installer or plugin sends the model under
`prompt`, `message`, `text` or `summary` — but it is a real narrowing and I would rather state it
than have you find it.

**The skipped field is `["model", "title"]`, not just `model`.** OpenCode's `session.created` puts
`info.title` in the same slot, and the repo's own test pins it (`title_hint == Some("New session")`).
Its plugin subscribes to `created`/`idle`/`deleted`/`compacted` and never `session.updated`, and
`startSession` runs once per id, so the only value that can arrive is the creation-time name — the
shared literal, never a rename. The honest framing is that a `SessionStart` title is whatever the
harness *named the session*, and no harness names it with something a person typed.

## What it buys beyond the two pages I found

Skipping the kind also excludes the router's own default. `title_hint.unwrap_or_else(|| kind.as_str()
.to_string())` writes the literal `session-start` whenever a harness sends neither `model` nor
`title` — the in-repo Kiro v3 fixture is exactly that payload — and the shape predicate never caught
it, because it has no digit. Compiled against both versions:

```
[SessionStart "session-start", UserPrompt ""]   v1.32.1: "session-start"   this PR: "Session {id}"
[SessionStart "New session",   UserPrompt ""]   v1.32.1: "New session"     this PR: "Session {id}"
```

Both are strings every affected page would share. That is a better argument for the change than my
n=2 anecdote, and it is now the test.

## The limitation worth naming

**The page that motivated #484 does not get its identity back — it gets `tool file`.** Claude Code
tool observations are titled by `safe_tool_title` as the literals `tool file` / `tool non-file`, so
that page retitles from `claude-opus-5[1m]` to `tool file`: one shared literal traded for another.
`Session {id}` is reached only when a session has no tool observation at all. The 1.32.1 entry
promised a fallback to identity rather than a shared literal; this change does not extend that
promise, and the new test encodes the outcome rather than hiding it. Making the fallback loop also
skip a title equal to its own `kind.as_str()` would close it, but that is your design call and not
this fix.

Nothing is destroyed either way: `render_raw_observation` still prints every observation's title, so
the model id remains visible in the page's raw-observations section. Only the title slot changes.

## The anecdote, checkable rather than trusted

Both model-id-titled pages in the 1,835-page corpus from #484 were read on disk. The one inside a
real session:

```
obs=17
- `session-start` @ 2026-08-21T19:50:41Z — claude-opus-5[1m]
- `user-prompt`   @ 2026-08-21T20:07:11Z —
- `pre-tool-use`  @ 2026-08-21T20:07:18Z — tool non-file
```

The `user-prompt` title is empty, which is why the fallback ran at all, and that is reproducible
rather than mysterious: `truncate_for_title` keeps only the characters before the first newline, so
a prompt beginning with `\n` yields `Some("")` — which passes straight through the `unwrap_or_else`
kind-name fallback in the router. The prompt was never scaffolding. It was absent, and no filter on
the string could have caught that.

## Tests

Two of yours changed. Both fixtures put the model id on a `UserPrompt` observation, and nothing
populates a `UserPrompt` title from a `model` field, so I moved it to `SessionStart` in both. That
preserves each test's intent and makes them model the real path — but it does rewrite an assertion
you merged this morning, so say the word if you read it differently and I will restore them and
narrow the branch instead.

New: the router-default case above, the production shape, and one asserting `pr-477` / `issue-484` /
`commit-a0bc43d` survive as titles. Each fails on `main` as it stands.

Gates: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
`cargo test --workspace` (2,584 passed / 0 failed), `cargo deny check`,
`scripts/check-changelog-sections.sh`.

Refs #484, #489.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,9 +21,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   written whenever a harness sends neither `model` nor `title`. The trade is
   narrower than a shape rule and stated as such: a model id a user *typed* is
   now kept, because at that point it is what they wrote. Where the prompts
-  are unusable a page falls to the next observation title, usually the shared
-  literal `tool file`, and to `Session {id}` only when there is no tool
-  observation at all. (#484)
+  are unusable a page falls to the next observation title, which the router
+  defaults to the observation's own kind, so `Session {id}` is reached only
+  by a session carrying nothing else at all — measured over 332 real sessions
+  it never was, and 5 of them landed on `tool file` / `tool non-file` against
+  327 keeping a real title. `looks_like_scaffolding` is re-exported from
+  `ai-memory-core`, so its narrowing is a public-API behaviour change even
+  though `derive_title` is its only caller in tree. (#484)
 
 ### Docs
 - Documented why registering ai-memory through Kimi Code's own
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Stopped the scaffolding filter discarding terse prompts. The identifier
+  branch added in #484 fired on any single token carrying a digit and a
+  hyphen, underscore or bracket, which is the shape of a bare model id and
+  equally the shape of `pr-477`, `issue-484` or `commit-a0bc43d`: measured
+  against the two populations, it matched 10 of 10 model ids and 7 of 11
+  plausible terse references, so a page could lose a real title. The class is
+  now excluded at its source instead — `derive_title` skips `SessionStart`
+  outright, the only kind whose title `best_title_hint` fills from the
+  harness's `model` field — and the branch is gone. That also excludes the
+  router's default `title_hint.unwrap_or(kind)`, the literal `session-start`
+  written whenever a harness sends neither `model` nor `title`. The trade is
+  narrower than a shape rule and stated as such: a model id a user *typed* is
+  now kept, because at that point it is what they wrote. Where the prompts
+  are unusable a page falls to the next observation title, usually the shared
+  literal `tool file`, and to `Session {id}` only when there is no tool
+  observation at all. (#484)
+
 ### Docs
 - Documented why registering ai-memory through Kimi Code's own
   `kimi mcp add` breaks every model turn: that command writes the plain
--- a/crates/ai-memory-core/src/scaffolding.rs
+++ b/crates/ai-memory-core/src/scaffolding.rs
@@ -14,6 +14,17 @@
 //! silently, because a bad title just looks like a title. `transcript.rs`
 //! already carries a narrower version of this idea for `<system-reminder>`
 //! and `<user_info>` on the workstream path.
+//!
+//! **A bare model id is not tested for here, on purpose.** It has the same
+//! shape as a terse human reference — `gpt-5` and `pr-477` differ only in
+//! meaning — so any lexical rule that catches one catches the other, and this
+//! predicate errs toward keeping text. Every model id that reached a title
+//! in the surveyed corpus arrived on a `SessionStart` observation, which is
+//! the only kind whose title `best_title_hint` fills from the harness's
+//! `model` field, so `derive_title` skips that kind outright instead.
+//!
+//! That is narrower than a shape rule, deliberately: a model id somebody
+//! *typed* is now kept, because at that point it is what the user wrote.
 
 /// Characters that open a decorated shell prompt rather than prose.
 /// Box-drawing and the powerline separators commonly echoed into a paste.
@@ -61,19 +72,6 @@ pub fn looks_like_scaffolding(candidate: &str) -> bool {
         return true;
     }
 
-    // A bare identifier rather than a sentence: no whitespace anywhere, and
-    // carrying identifier punctuation. Real prompts essentially always
-    // contain a space; requiring the punctuation too keeps a one-word prompt
-    // like "help" or "continue" as a legitimate title.
-    if !trimmed.contains(char::is_whitespace)
-        && trimmed
-            .chars()
-            .any(|c| c.is_ascii_digit() || c == '[' || c == ']')
-        && trimmed.chars().any(|c| c == '-' || c == '_' || c == '[')
-    {
-        return true;
-    }
-
     false
 }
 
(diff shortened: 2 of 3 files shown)


abhisheksharma2411 (commented):
Verified the load-bearing claim rather than trusting the docstring, and it holds exactly. `best_title_hint` maps the `["model", "title"]` field on **only** `HookEvent::SessionStart`:

```rust
HookEvent::SessionStart => extract_string(raw, &["model", "title"]),
HookEvent::UserPrompt   => extract_content(raw, &["prompt", "message", "text"])…
```

so excluding by kind really does close the same hole the shape rule was covering, and it closes it at the source rather than downstream of it. `cargo test -p ai-memory-core -p ai-memory-hooks` → **163 and 270 passing, 0 failed**.

The method here is the right one and worth naming: a lexical rule that matched **10 of 10 model ids and 7 of 11 legitimate terse references** is not a filter, it is a coin toss weighted toward data loss. Replacing it with a structural fact about where the value comes from is strictly better than tuning the regex, and the changelog stating the residual trade — a model id somebody *typed* is now kept — is exactly the right thing to write down.

## One fragility worth naming

Skipping the whole `SessionStart` kind in the second loop also discards OpenCode's `info.title`, and the justification is a behavioural claim about someone else's code:

> Its plugin subscribes to `created`/`idle`/`deleted`/`compacted` and never `session.updated`, and starts a session once per id, so the only value that can arrive is the creation-time name — in practice the shared literal "New session", never a rename.

That is well-researched and I believe it today. But it is pinned to a third party's subscription list, which can change in a release with no signal here, and if OpenCode ever surfaces a renamed session the code silently discards a genuinely good title. Nothing in the repo would fail. Worth either a test asserting the current expectation so a future surprise is visible, or a line in the docstring marking it as an assumption about an external component rather than an invariant.

## A question about the downgrade path, not an objection

The changelog is admirably explicit that where prompts are unusable a page now falls to the next observation title — "usually the shared literal `tool file`" — and to `Session {id}` only when there is no tool observation.

So for the affected sessions the title goes from `claude-opus-5[1m]` to `tool file`. Both are useless as titles, but they fail differently: the model id is *wrong* (it looks like content and is not), whereas `tool file` is *uninformative but honest*, and — the part I think matters — it is **shared**, so N such pages collide into an indistinguishable run in any listing, where N distinct model-id titles at least sorted apart.

I am not arguing for the old behaviour. I am asking whether `Session {id}` would be the better fallback than a shared literal in that specific case, since it is at least unique per page and immediately reads as "this page has no title" rather than as a title. Entirely your call — you have the corpus and I do not.

## Small note

`looks_like_scaffolding` is re-exported from `ai-memory-core` (`pub use scaffolding::looks_like_scaffolding`), so this is a public-API behaviour change, not only an internal one: any consumer calling it directly stops seeing bare identifiers classified as scaffolding. Only `synth.rs` uses it in-tree (two call sites, both checked), so the blast radius is nil here — just flagging that the changelog entry reads as an internal fix and the symbol is public.

samirhvbr:
@abhisheksharma2411 — thank you. All three land, and one of them is a sentence I got wrong.

Pushed in 4745518.

## The downgrade path — you asked for the number, so here it is

You said this was my call because I have the corpus and you do not. That is fair, so rather than
answer with an opinion I simulated this change's `derive_title` over the 1,835 session pages on the
live instance from #484, using each page's own raw-observations list as the input:

| outcome for the 332 real sessions | n | |
|---|---|---|
| keeps a real user title | 327 | 98.5% |
| lands on a shared literal | **5** | 1.5% |
| reaches `Session {id}` | **0** | 0% |

The five are four `tool non-file` and one `tool file`.

So your instinct about collision is right in kind and small in degree: those five pages do collapse
into two indistinguishable runs. But the more important line is the third one — **`Session {id}` is
never reached**, which means the sentence in my changelog was false. I had written that a page falls
to identity "only when there is no tool observation at all". The router defaults an observation's
title to its own kind (`title_hint.unwrap_or(kind.as_str())`), so `stop` and `session-end` always
carry one; identity requires a session with nothing else whatsoever. Copilot flagged the same
sentence from the other direction. Corrected.

On whether `Session {id}` should win over a shared literal for those five: I think you are right in
principle — unique-and-obviously-untitled beats shared-and-looks-like-a-title — but at n=5 out of 332
I did not want to reach for it inside this PR, because the rule that gets there has to reject
`tool file` specifically, and `title_falls_back_through_kinds` deliberately accepts a tool name as a
title today. That is @akitaonrails's design call, and it is now a decision with a number attached
rather than a preference.

## The OpenCode assumption

You are right that I pinned a behavioural claim to a third party's subscription list, and that
nothing here would fail if it changed. The docstring now says so explicitly — that it is read off
OpenCode's plugin as of v1.32.1, not an invariant this repo holds, and that a future release
surfacing a renamed session would discard a good title silently. I did not add a test: a test here
would assert *our* expectation of *their* event vocabulary, which passes forever regardless of what
OpenCode does, so it would look like coverage without being any. Marking it as an assumption seemed
more honest than pinning it.

## The public API note

Taken, and added to the entry. `looks_like_scaffolding` is re-exported from `ai-memory-core`, so
narrowing it is a public-API behaviour change even though `derive_title` is its only caller in tree.
The changelog now says that rather than reading as an internal fix.

Thanks for checking the load-bearing claim against `best_title_hint` rather than against my
docstring — that is the second time you have gone to the code instead of the prose, and both times
it moved something.

akitaonrails:
Merged as `0b32a81`. You found a real defect in code I wrote and merged yesterday, and the case is airtight.

I reproduced your table by compiling the shipped predicate and running it against both populations rather than reading the diff:

```
MODEL  claude-opus-5[1m] -> true    REF  pr-477        -> true
MODEL  gpt-5             -> true    REF  issue-484     -> true
MODEL  o3-mini           -> true    REF  fix-484       -> true
MODEL  kimi-k2           -> true    REF  commit-a0bc43d-> true
…10 of 10                           …7 of 11
```

**10 of 10 and 7 of 11, exactly as you reported.** `pr-477` and `issue-484` are prompts someone would plausibly type in *this repository*, and my branch discarded them — against the principle stated three lines above it, as you point out. That is the worse direction to fail in, since a discarded title is silent.

The other two claims verified too:

- `payload.rs:774` — `HookEvent::SessionStart` is the **only** arm reading `model`. Every other arm reads `prompt`/`message`/`text`/`tool`/`stepIdx`. So excluding by kind is precisely targeted where a shape rule was a proxy.
- The OpenCode plugin subscribes to `session.created`/`idle`/`deleted`/`compacted` and **zero** occurrences of `session.updated`, so no rename can reach that slot. Your reasoning about what is given up holds.

Two things about how this was written that I want to name, because they are rarer than the fix:

1. **You stated the narrowing rather than letting me find it** — including the concrete before/after against my own test fixture, showing v1.32.1 produced `Session {id}` where this produces `claude-opus-5[1m]`. That is the honest framing of a real regression in coverage, and burying it would have been easy.
2. **The doc comment flags its own residual risk**: *"That last part is an assumption about an external component, not an invariant"* — that if a future OpenCode release surfaces a renamed session here, this discards a good title with nothing in the tree failing. Writing down what would break silently is worth more than the code.

Gate under the pinned 1.95 toolchain: fmt, clippy `-D warnings`, changelog, **2585 passed / 0 failed**.

Thank you — that is four corrections from you today, three of them to my work.

PR #491, opened 2026-08-25
fix(web): block remote Markdown image fetches

tenhishadow:
## What changed

- apply a dedicated same-origin policy to rendered Markdown images
- preserve external clickable links and relative/root-relative images
- add regression coverage for schemes, protocol-relative URLs, UNC-style paths, alt text, and local images
- document the behavior in the README and changelog

## Why

Stored Markdown was sanitized for script execution, but an absolute image URL was still emitted unchanged. Merely viewing a page could therefore contact an arbitrary remote host and disclose the viewer IP, user agent, and view timing without a click.

The root cause was sharing the link URL allow-list with image sources even though links and auto-fetched subresources have different privacy boundaries.

## Impact

External links remain clickable. Relative and root-relative images continue to render. External and scheme-bearing image targets are neutralized before HTML rendering.

## Validation

- `git diff --check`
- `cargo deny check`
- independent adversarial review: P0/P1/P2 = 0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   text. A newer CLI against an older server renders the rest of `status` and
   omits the section rather than failing to parse.
 
+### Fixed
+- Prevented stored Markdown from automatically fetching external image URLs
+  when viewed in the web UI, while preserving clickable external links and
+  same-origin relative images (#491).
+
 ### Docs
 - Documented the order of magnitude reported for lifecycle-hook overhead,
   including the fact that a completed tool call normally pays for both pre- and
--- a/README.md
+++ b/README.md
@@ -760,7 +760,10 @@ Useful entry points:
   browser view of the markdown wiki. `--enable-web` also mounts a
   read-only JSON frontend API at `/api/v1` (workspaces, projects, pages,
   recent, briefing, search) so custom web UIs can read the memory without
-  opening SQLite or wiki files directly:
+  opening SQLite or wiki files directly. Rendered pages keep external links
+  clickable, but image sources must be relative or root-relative; absolute
+  external images are neutralized so viewing stored Markdown cannot act as a
+  remote beacon:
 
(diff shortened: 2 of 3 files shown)


akitaonrails:
Please mark this ready for review — I would like to merge it.

I verified the vulnerability on `main` before reading your diff, and your root-cause analysis is exactly right:

```rust
Event::Start(Tag::Link  { .. }) => … dest_url: safe_url(dest_url),   // markdown.rs:250
Event::Start(Tag::Image { .. }) => … dest_url: safe_url(dest_url),   // markdown.rs:261
```

Both go through the same `safe_url`, and `safe_url` **allows** `http://` and `https://`. So an absolute image URL is emitted unchanged and the browser fetches it on render.

**"Links and auto-fetched subresources have different privacy boundaries" is the whole bug in one sentence.** A link is a user decision; an image is not. Sharing one allowlist between them silently converts stored Markdown into a tracking-pixel channel — viewer IP, user agent and view timing to an arbitrary host, no click required.

Two things make this worse in this project specifically than it would be in a generic wiki:

- The rendered content is **agent-authored**, so a page can acquire a remote image without any human having typed it.
- `/web` is typically bound on a LAN or reached through a reverse proxy, so the leaked IP is often an internal address, and the fetch confirms a specific page was viewed at a specific time. At least one user of this project has stated legal confidentiality obligations (#446); a beacon that fires on view is not a theoretical concern for them.

Things I will be checking when it comes out of draft, so you can pre-empt them:

1. **`srcset` and any other subresource attribute**, if the renderer can emit them — the same reasoning applies to every auto-fetched URL, not only `src`.
2. **Protocol-relative `//host/x.png`** — your description mentions it; I want the test to assert the *rendered output*, not just the helper's return value.
3. That **relative and root-relative images still render**, since that is the legitimate case and the regression risk of this change.
4. That external **links** are untouched, which your summary says and I will confirm in the diff.

No rush on my account — but the fix is small and the exposure is live, so I would rather review it now than have it sit in draft.

tenhishadow:
@akitaonrails done

Issue #493, opened 2026-08-26
macOS native hook-drain never transmits: spooled events are silently never delivered (v1.32.1)


swhite1122:
**Version:** v1.32.1 native macOS aarch64 release binary (checksum verified)
**Platform:** macOS 15 (Darwin 25.6.0), Apple Silicon
**Setup:** native binary, `serve --transport http --bind 127.0.0.1:<port>` under launchd, `AI_MEMORY_AUTH_TOKEN` set on the server, non-default data dir via `--data-dir`, hooks installed with `install-hooks --agent claude-code --capture-mode allowlist --apply` (also reproduced with a default denylist store)

## Summary

Native hook capture spools events correctly, but `hook-drain` never delivers them: it makes **no network request at all**. Events either sit in `hook-spool/` indefinitely or, in some sequences, are consumed without ever being stored. The server itself is fine — hand-POSTing the spooled envelope's own `url`/`body` to `/hook` returns `202 queued` and the observation is stored.

## Repro

1. `ai-memory --data-dir <dir> hook --event post-tool-use --agent claude-code --server-url http://127.0.0.1:<port>` with a valid Claude Code PostToolUse JSON on stdin → envelope appears in `<dir>/hook-spool/` with the correct self-addressed `url`, `body`, `auth_mode`, `attempts` fields. ✅
2. `ai-memory hook-drain --data-dir <dir>` (foreground, file aged >60s) → exits 0, silent, spool file untouched or consumed, nothing stored, `logs/hook-drain.log` stays empty.
3. **Listener probe:** point the store's `server_url` at `nc -l 59999` and run `hook-drain` with an aged spool file present → **zero bytes ever arrive**. The drain never opens a connection.
4. Control: `curl -X POST "<envelope url>" -H "Authorization: Bearer <token>" --data-binary "<envelope body>"` → `202 queued`, observation visible in SQLite. ✅

Tried with the bearer token provided via env (`AI_MEMORY_AUTH_TOKEN`), via `--auth-token`, and via `bearer_token` in `config.toml` — no change; the probe shows no request regardless, so auth never enters the picture.

Session-start's synchronous cleanup drain and the detached session-end helper show the same behavior (helper process spawns, delivers nothing, logs nothing).

## Impact

[...]

PR #495, opened 2026-08-26
fix(hooks): report what a hook-drain pass did instead of exiting silently

samirhvbr:
Partial for #493 — the observability half, not the delivery fix. I cannot reproduce the macOS
symptom and this does not attempt to.

## The part of #493 that is reproducible anywhere

`run_drain` discarded its own result:

```rust
Ok(hook_spool::LockedDrainResult::Drained(_))
| Ok(hook_spool::LockedDrainResult::LockBusy) => {}
```

The drain already computes `sent` / `remaining` / `dropped`, and all three were thrown away. So three
passes that mean opposite things were byte-identical on both streams and in the exit code:

- every queued event delivered;
- every queued event **discarded undelivered** — past `MAX_ATTEMPTS`, or older than `MAX_AGE_MS`,
  both of which `load_live_entry` handles by deleting the file;
- **nothing attempted at all**, because another drainer held the spool lock and
  `DrainLockWait::Bounded(30s)` expired.

A drain silently losing capture is therefore indistinguishable in the field from one that works.
That is the shape of the #493 report itself: exit 0, no output, an empty `hook-drain.log`, and no way
for the reporter to tell which of the three they were in. Whatever the macOS root cause turns out to
be, it was diagnosed blind because the drain says nothing about its own outcome.

## What this changes

A pass that leaves nothing behind stays silent, so `hook-drain.log` keeps receiving only warnings and
a healthy instance never grows it — the rotation budget in `hook_drain_process` still holds. A pass
that leaves events queued or drops them, and a pass that found the lock held, each emit one stderr
line.

stderr, not stdout: the drainer's stdout is contractually empty, and the detached helper already
redirects stderr into `logs/hook-drain.log` (`stderr_file: spec.stderr_log.clone()`).

The wording says **"acknowledged"**, not "delivered". `sent` counts what the server ACKed via
`delete_accepted_indices`, and an ack also covers an item the server accepted and then dropped by
capture policy. Reporting those as delivered would assert storage the drain cannot observe — which
is the precise confusion #493 was written in, so it seemed worth not repeating in the fix.

**No delivery behaviour changes.** This only stops the existing behaviour being unfalsifiable.

## What it does not do

It does not explain why the drain makes no network request on macOS. @swhite1122's listener probe is
convincing and I have no way to reproduce it — the three cases above are what this makes
distinguishable, and if the macOS symptom is none of them, this narrows it rather than solving it.

One asymmetry worth flagging while looking, in case it is relevant: `spool_health` filters spool
files through `created_ms_from_name`, and the comment there says AppleDouble sidecars are *"Not
hypothetical on macOS"* — `._<original>.json` on filesystems without extended attributes, with a
dedicated test case. `list_entries`, which is what `drain` walks, applies no such filter: it takes
every `*.json`. `Vec<PathBuf>::sort` orders by bytes, so `._…` (0x2E) sorts before any digit and such
a file is processed first. It would fail to parse and be dropped, which on its own only costs a
phantom entry — but the two paths disagree about what counts as a spool file, and only one of them
was hardened. I have not connected that to the reported symptom and am not claiming it is the cause.

Gates: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
`cargo test --workspace` (2,588 passed / 0 failed), `cargo deny check`,
`scripts/check-changelog-sections.sh`.

Refs #493.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- Made `hook-drain` say what a pass actually did. The drain already returned
+  `sent`/`remaining`/`dropped` counts and `run_drain` discarded all three, so
+  three passes that mean opposite things — everything delivered, every queued
+  event discarded undelivered at the retry cap or the spool TTL, and nothing
+  attempted at all because another drainer held the lock — were byte-identical
+  on both streams and in the exit code. A drain that silently loses capture was
+  therefore indistinguishable in the field from one that works, which is how
+  #493 was reported: exit 0, no output, an empty `hook-drain.log`, and no way
+  to tell which of the three had happened. A pass that leaves nothing behind
+  stays silent, so `hook-drain.log` keeps receiving only warnings and a healthy
+  instance never grows it; a pass that leaves events queued or drops them, and
+  a pass that found the lock held, now each emit one stderr line. stderr, not
+  stdout: the drainer's stdout is contractually empty and the detached helper
+  already redirects stderr into the log. This changes no delivery behaviour —
+  it only stops the existing behaviour being unfalsifiable. (#493)
 - Stopped the scaffolding filter discarding terse prompts. The identifier
   branch added in #484 fired on any single token carrying a digit and a
   hyphen, underscore or bracket, which is the shape of a bare model id and
--- a/crates/ai-memory-cli/src/commands/hook.rs
+++ b/crates/ai-memory-cli/src/commands/hook.rs
@@ -317,11 +317,53 @@ pub async fn run_drain(data_dir: Option<PathBuf>) -> anyhow::Result<()> {
     )
     .await
     {
-        Ok(hook_spool::LockedDrainResult::Drained(_))
-        | Ok(hook_spool::LockedDrainResult::LockBusy) => {}
+        Ok(outcome) => {
+            let _ = write_drain_report(&mut std::io::stderr(), &outcome);
+        }
         Err(err) => eprintln!("ai-memory hook-drain warning: failed to acquire drain lock: {err}"),
     }
     Ok(())
+}
+
+/// Report a drain pass on stderr when it ended with events undelivered.
+///
+/// Silent on a clean pass (nothing queued, or everything delivered) so
+/// `hook-drain.log` stays warning-only, matching the rotation budget in
+/// [`hook_drain_process`]. stderr, never stdout: the drainer's stdout is
+/// contractually empty and the detached helper redirects stderr into the log.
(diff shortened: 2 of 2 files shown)


samirhvbr:
Both inline comments checked against the code. One was right and is fixed in `cf11efa`; the other I
am declining, with the reasoning below so it is on the record rather than silently ignored.

## The LockBusy wording — correct, and it was the fix committing the error the fix is for

> the LockBusy report says it "left queued events untouched", but at this point the code can't know
> whether there were any events queued

Right, and the mechanism is sharper than "can't know": `drain_until_quiescent` returns `LockBusy`
from its **first statement**, before the lock is held and therefore before anything is listed —

```rust
let Some(_lock) = acquire_drain_lock(spool, wait)? else {
    return Ok(LockedDrainResult::LockBusy);
};
```

— so the queue count is not merely unverified there, it is structurally unavailable.

What makes this worth more than a wording tweak is that it is the same move this PR argues against.
The `Drained` arm deliberately says **"acknowledged"** rather than "delivered", because `sent` counts
server acks and an ack can cover an item the server accepted and then dropped by capture policy;
calling that a delivery would assert storage the drain cannot observe. The `LockBusy` line then
turned around and asserted the spool contents. Same error, two lines apart.

It now reports only what the pass did — *attempted no delivery and left the spool untouched* — which
holds whether the spool has one event or none. Pinned with an assertion that the report does not
mention queued events: the old wording contained that word, so the assertion fails against it rather
than merely passing alongside it.

## The discarded I/O result — declining, and here is why

> `run_drain` discards the `write_drain_report` I/O result (`let _ = ...`) (…) it's better to at
> least best-effort surface write failures (and ideally lock stderr once)

Two parts, and neither survives contact with this call site.

**Surfacing a stderr write failure has nowhere to go.** The only channels this command has are stdout
— contractually empty, and the detached helper does not capture it — and stderr, which is the thing
that just failed. Reporting a failed stderr write to stderr is either a no-op or a recursion, and
returning it from `run_drain` would turn a best-effort observability line into a nonzero exit from a
hook, on a path whose whole contract is never to block or fail the agent.

**Locking once buys nothing here.** `write_drain_report` contains two `writeln!` calls, but they are
in different `match` arms, so exactly one executes per invocation, and `run_drain` invokes it once
per process. There is no interleaving to prevent and no repeated lock acquisition to amortise.

The irony in the observation is fair — a function about not swallowing outcomes swallowing its own
write result — but the swallowing is the correct behaviour on this path, and making it visible would
cost more than it explains.

Gates after the fix: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -D warnings`,
`cargo test --workspace` (2,588 passed / 0 failed), `cargo deny check`,
`scripts/check-changelog-sections.sh`.

Issue #492, opened 2026-08-26
Native ai-memory client rejects Caddy's internal CA (UnknownIssuer) even after Path 2's trust-install step — hook capture fails silently


alanmatiasdev:
**Version**
Output of `ai-memory --version`:
```
ai-memory 1.32.1
```
(client: `ai-memory-bin 1.32.1-1`, AUR; server: Docker image `akitaonrails/ai-memory:1.32.1`)

**What happened**

Following the [Path 2 — Caddy with internal CA (LAN-only)](docs/https-via-proxy.md#path-2--caddy-with-internal-ca-lan-only)
guide exactly (Caddy `tls internal`, root cert extracted and installed into
the client OS trust store), `curl` and the Claude Code MCP client both trust
`https://memory.home.lan` correctly — but the native `ai-memory` binary
rejects the same certificate:

```
Error: HTTP request to https://memory.home.lan/admin/open-sessions failed

Caused by:
    0: error sending request for url (https://memory.home.lan/admin/open-sessions?workspace=default&project=home-server&agent=claude-code&all=false&all_owners=false)
    1: client error (Connect)
    2: invalid peer certificate: UnknownIssuer
```

Reproduced with `finalize-session`, but it affects every command the native
binary makes over HTTPS — including `ai-memory hook ...`, which is what
`install-hooks` wires into every lifecycle hook. In practice this means
lifecycle capture silently fails end-to-end on any install that follows
Path 2: events queue in the local `hook-spool/` (queuing doesn't touch the
network) but the session-end drain that actually POSTs them to the server
fails with `UnknownIssuer` every time, so nothing ever reaches
`pages`/`sessions`/`observations` — with no visible error surfaced to the
agent or the user. I only found this by manually running `finalize-session`
and reading its output; a normal Claude Code session gives no signal that
capture is broken.

**Root cause (found via `Cargo.toml`)**

```
reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls", "stream"] }
```

The `rustls-tls` feature alone makes `reqwest` validate certificates against
`webpki-roots` (a static, bundled copy of Mozilla's public CA list) — it
never reads the OS trust store. `rustls-tls-native-roots` (pulling in
`rustls-native-certs`) is required for that, and it isn't enabled. That's why
`curl` (OpenSSL, reads `/etc/ssl/certs`) accepts a CA that the `ai-memory`
binary itself will never accept, regardless of what's installed system-side.

This also contradicts the docs:

[...]

PR #496, opened 2026-08-26
fix(tls): trust the platform CA store instead of bundled webpki roots (#492)

akitaonrails:
Refs #492. Reported by @alanmatiasdev.

**⚠️ This changes TLS behaviour for every outbound HTTPS request in the product. I have tested it but I am not merging it without your sign-off** — see *Blast radius* below.

## The bug

The workspace built reqwest with `rustls-tls`, which compiles in the **Mozilla webpki root store and never consults the OS trust store**. So a CA the operator installed locally — Caddy's `tls internal`, a corporate MITM appliance, any private PKI — was invisible to `ai-memory` while `curl` and the agent CLIs trusted it. That asymmetry is what made this read as an ai-memory bug rather than a PKI one.

Every HTTPS request failed with `invalid peer certificate: UnknownIssuer`, **including `ai-memory hook`**. Because spooling touches no network, capture looked healthy: events queued locally and the drain silently failed to deliver them, forever. Anyone following **this project's own Path 2 guide** ended up with a dead capture path and no signal — the reporter only found it by running `finalize-session` by hand.

## Verified against a private CA the system store does not know

| build | CA supplied | result |
|---|---|---|
| `rustls-tls` (before) | yes | `invalid peer certificate: UnknownIssuer` — **the reporter's exact error** |
| `rustls-tls-native-roots` (after) | yes | handshake succeeds, request reaches HTTP-level parsing |
| after | no | `client error (Connect)` — correctly still rejected |

The middle row is the control that matters: the same CA handed to the **pre-change** binary is still rejected, so the trust comes from the feature change and not from how the CA was supplied. My first attempt at that control was invalid — a failed `git stash` meant nothing rebuilt and I re-tested the new binary — so I reverted `Cargo.toml` explicitly and rebuilt before trusting it.

I used `SSL_CERT_FILE` to point at the test CA rather than modifying a real system trust store. `rustls-native-certs` honours both; the reporter's case is the OS-store half, which I did not exercise directly.

## Blast radius — why I want your call

This affects **every** outbound HTTPS call: LLM providers, embeddings, hooks, MCP clients, the importer.

- **Container: safe.** The runtime image is `debian:bookworm-slim` with `ca-certificates` installed, so the store is populated.
- **Native installs: safe** in any normal configuration — this is the same store `curl` uses.
- **The risk** is an environment with an empty or unreadable trust store, which previously worked off the bundled roots and would now have none. That is unusual, and such a host would already fail `curl`, but it is a real behaviour change rather than a pure bug fix.

The conservative alternative is to keep webpki roots and document that private CAs are unsupported — which I think is worse, since the project ships a guide that walks users straight into this.

## Gate

`cargo fmt --all -- --check` ✅ · `cargo clippy --workspace --all-targets -- -D warnings` ✅ · `cargo test --workspace --all-targets` → **2588 passed, 0 failed** · `cargo deny check` → advisories, bans, licenses, sources all ok (the new transitive `rustls-native-certs` passes policy).

All under the pinned 1.95 toolchain.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- The native client now trusts CAs from the platform trust store. It was built
+  with reqwest's `rustls-tls`, which bundles the Mozilla webpki roots and
+  ignores the OS store, so a CA the operator had installed locally — Caddy's
+  `tls internal`, a corporate MITM appliance, any private PKI — was invisible
+  to `ai-memory` even though `curl` and the agent CLIs trusted it. Every HTTPS
+  request failed with `invalid peer certificate: UnknownIssuer`, including
+  `ai-memory hook`, so following this project's own HTTPS-via-proxy Path 2
+  guide left lifecycle capture failing end to end: events spooled locally
+  (queuing touches no network) and the drain could never deliver them.
+  Switched to `rustls-tls-native-roots`; the runtime image installs
+  `ca-certificates`, so the server path keeps a populated store. Reported by
+  @alanmatiasdev (#492).
+
 - Made `hook-drain` say what a pass actually did. The drain already returned
   `sent`/`remaining`/`dropped` counts and `run_drain` discarded all three, so
   three passes that mean opposite things — everything delivered, every queued
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -664,6 +664,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
 name = "core-foundation-sys"
 version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1343,10 +1353,10 @@ dependencies = [
  "hyper",
  "hyper-util",
  "rustls",
+ "rustls-native-certs",
  "tokio",
  "tokio-rustls",
  "tower-service",
- "webpki-roots",
 ]
 
 [[package]]
@@ -1937,6 +1947,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
+name = "openssl-probe"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c87def4c32ab89d880effc9e097653c8da5d6ef28e6b539d313baaacfbafcbe"
+
+[[package]]
 name = "option-ext"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


abhisheksharma2411 (commented):
Happy to give you the sign-off you asked for on the change itself — and I found one thing in the blast-radius section that I think you'll want, because it's a path you named that this doesn't actually reach.

**The feature change resolves the way you intend.** Checked the resolved graph rather than the manifest, since a workspace this size can have a transitive dep quietly re-enable something:

```
reqwest v0.12.28
├── reqwest feature "rustls-tls-native-roots"
├── reqwest feature "rustls-tls-native-roots-no-provider"
├── reqwest feature "__rustls" / "__rustls-ring" / "__tls"
└── reqwest feature "json" / "stream"
```

No `webpki-roots` anywhere in the resolved set, and nothing re-adds `rustls-tls`. Your control experiment is the right one and I'd have asked for exactly it — especially the note about the failed `git stash` invalidating the first attempt. That failure mode has bitten me three times in this programme and it always produces a *false green*, so calling it out explicitly is worth more than the result itself.

Also worth confirming since it's easy to miss: `companions/ai-memory-importer` is **not** a workspace member, so it genuinely needed its own edit rather than inheriting the workspace dep. Good catch.

**The part I'd change is the blast-radius list, not the code.** It says this affects *"LLM providers, embeddings, hooks, MCP clients, the importer."* MCP clients are the odd one out — they don't go through this reqwest at all. There are **two independent copies of reqwest** in the tree:

```
reqwest v0.13.4
└── rmcp v1.7.0
    ├── ai-memory-cli
    └── ai-memory-mcp
```

And `rmcp` resolves that copy with **no TLS feature whatsoever**:

```
reqwest v0.13.4
├── reqwest feature "json"
└── reqwest feature "stream"
```

`transport-streamable-http-client-reqwest` pulls in `__reqwest = ["dep:reqwest"]` and stops there. So the workspace's root-store choice — before or after this PR — has never applied to the MCP transport.

I built exactly that feature set to see what it does rather than infer it:

```
HTTPS FAILED: error sending request for url (https://example.com/)
  caused by: client error (Connect)
  caused by: invalid URL, scheme is not http
HTTP  OK: status 200 OK
```

So the streamable-HTTP MCP client can't speak HTTPS at all today — it isn't a root-store problem there, it's that no TLS backend is compiled in. **Pre-existing and not caused by this PR**, but it does mean the sentence promising this covers MCP clients isn't true yet, and that's the sentence you're asking people to sign off on.

**The good news is that the fix is one feature flag, and it lands better than the 0.12 side.** `rmcp` exposes:

```toml
reqwest = ["__reqwest", "reqwest?/rustls"]
```

and in reqwest **0.13** the `rustls` feature pulls `rustls-platform-verifier` rather than a bundled root set — so the webpki-vs-native distinction you're fixing here has already been resolved upstream in that version. Enabling `rmcp/reqwest` would give the MCP transport OS-trust behaviour by construction, which is the same end state this PR establishes for 0.12.

Whether that belongs in this PR or a follow-up is your call — there's a decent argument for keeping this one narrow given you're explicitly asking for sign-off on blast radius. I'd just soften the MCP line here and open the follow-up, rather than widen the diff.

On the remaining risk you flagged — the empty/unreadable trust store. Agreed that's the real regression surface and I don't think it's avoidable while still fixing #492; `curl` parity is the right bar. The one thing I'd add to the changelog is that the failure is now a *connect-time* error rather than silent spooling, because the thing that made #492 nasty wasn't the failure, it was that capture looked healthy while the drain died. Worth saying out loud that this change doesn't alter that: a drain failure is still invisible unless someone runs `finalize-session` by hand. That's arguably the more valuable follow-up of the two.

Sign-off from me on the change as written, with the MCP sentence adjusted.

piradata:
como vcs tão fazendo pra ia gerar essas prs e as proprias ias validarem o code? tem um trigger no github? tava tentando fazer  algo assim com o cursor no azure devops

PR #497, opened 2026-08-26
fix(mcp-bridge): give the streamable-HTTP client a TLS backend

abhisheksharma2411:
Refs #492. This is the MCP follow-up I mentioned while reviewing #496 — entirely your call whether it belongs there, here, or nowhere.

## The bug

`ai-memory-cli` enables rmcp's `transport-streamable-http-client-reqwest`, which pulls reqwest in through `__reqwest = ["dep:reqwest"]` and stops. That leaves the second reqwest in the tree with **no TLS backend at all**:

```
reqwest v0.13.4
├── reqwest feature "json"
└── reqwest feature "stream"
```

reqwest built without a TLS feature refuses any non-`http` scheme. I built exactly that feature set rather than inferring it:

```
HTTPS FAILED: error sending request for url (https://example.com/)
  caused by: client error (Connect)
  caused by: invalid URL, scheme is not http
HTTP  OK: status 200 OK
```

So `ai-memory mcp-bridge` against an `https://` server URL fails at connect, every time.

**What makes it worth fixing rather than documenting:** the bridge is the Claude Code stdio bridge to a *remote* MCP server, and `upstream_config` attaches a bearer token via `config.auth_header(token)`. With HTTPS structurally unavailable, the only scheme that worked was the one that puts that token on the wire in cleartext. The failure also can't be worked around by configuration — no `.ai-memory.toml` key or env var can add a TLS backend that wasn't compiled in.

This is a **second, independent copy** of reqwest: the workspace client is 0.12, and its root-store configuration — including whatever #496 settles — has never applied to this path.

## The fix

rmcp already exposes the feature:

```toml
reqwest = ["__reqwest", "reqwest?/rustls"]
```

and on reqwest **0.13** the `rustls` feature pulls `rustls-platform-verifier` rather than a bundled root set. So this lands on the platform trust store by construction — the same end state #496 is establishing for 0.12, reached without a root-store decision of its own.

## Verification

Feature resolution, before and after:

```
before:  json, stream
after :  json, stream, rustls, __rustls, __rustls-aws-lc-rs, __tls

cargo tree -i rustls-platform-verifier
  rustls-platform-verifier v0.7.0
  └── reqwest v0.13.4
      └── rmcp v1.7.0
          ├── ai-memory-cli
          └── ai-memory-mcp
```

The same standalone probe with `rustls` added:

```
HTTPS OK: status 200 OK
```

Build and gates:

```
cargo build   -p ai-memory-cli                 Finished, 36.04s
cargo clippy  -p ai-memory-cli --all-targets   no warnings
cargo test    -p ai-memory-cli                 94 passed, 0 failed, 1 ignored (8 binaries)
```

`Cargo.lock` moves because `rustls-platform-verifier` and its dependencies enter the graph.

## What I did not do

**No live handshake against a real MCP server** — I don't have one behind a private CA to point at. What is proven here is that the scheme rejection is gone and the platform verifier is compiled in; the last mile is the same thing #496 exercises with `SSL_CERT_FILE`, and I'd rather say so than imply I ran it.

**Left `ai-memory-mcp` alone.** It takes rmcp through `rmcp.workspace = true` for the *server* side, and feature unification means it now sees the client TLS features too. That's a consequence rather than a decision, and if you'd rather the client transport were scoped to `ai-memory-cli` only, that's a bigger change to the workspace dependency shape than this warrants.

Happy for this to be folded into #496 instead if you'd prefer one TLS change rather than two.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,11 +41,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   server URL failed at connect with `invalid URL, scheme is not http`, and the
   only URL that worked was plaintext `http://` — which is also the transport
   the bridge attaches its bearer token to. Enabling rmcp's `reqwest` feature
-  adds `reqwest?/rustls`, which on reqwest 0.13 is `rustls-platform-verifier`,
-  so the bridge now verifies against the platform trust store like every other
-  outbound request in the workspace. The workspace-client half of #492 landed
-  in #496; this is the MCP transport, which is a second, independent copy of
-  reqwest ([#497]).
+  the bridge now uses rmcp's `reqwest-tls-no-provider`
+  (`reqwest?/rustls-no-provider`), which supplies the platform certificate
+  verifier without pinning a crypto provider, and installs `ring` — already
+  compiled for this binary via the workspace's reqwest 0.12 — as the process
+  default. rmcp's plain `reqwest` feature would have pinned `aws-lc-rs`,
+  adding a compiled C dependency to every build for a bridge most users never
+  enable. The workspace-client half of #492 landed in #496; this is the MCP
+  transport, which is a second, independent copy of reqwest ([#497]).
 
 
 ## [1.33.0] - 2026-08-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -177,6 +177,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   exists rather than reading one from a file ai-memory would have to write and
   protect.
 
+- Gave `ai-memory mcp-bridge` a TLS backend so it can reach an `https://` MCP
+  server. `ai-memory-cli` enabled rmcp's `transport-streamable-http-client-reqwest`
+  feature, which pulls reqwest in via `__reqwest = ["dep:reqwest"]` and stops
+  there — no `rustls`, no `native-tls`, no TLS at all. reqwest with no TLS
+  feature refuses any non-`http` scheme, so every bridge run against an HTTPS
+  server URL failed at connect with `invalid URL, scheme is not http`, and the
+  only URL that worked was plaintext `http://` — which is also the transport
+  the bridge attaches its bearer token to. Enabling rmcp's `reqwest` feature
+  the bridge now uses rmcp's `reqwest-tls-no-provider`
+  (`reqwest?/rustls-no-provider`), which supplies the platform certificate
+  verifier without pinning a crypto provider, and installs `ring` — already
+  compiled for this binary via the workspace's reqwest 0.12 — as the process
+  default. rmcp's plain `reqwest` feature would have pinned `aws-lc-rs`,
+  adding a compiled C dependency to every build for a bridge most users never
+  enable. The workspace-client half of #492 landed in #496; this is the MCP
+  transport, which is a second, independent copy of reqwest ([#497]).
+
 ### Docs
 
 - `docs/windows.md` gains **Scenario E**, a persistent-server story for native
@@ -453,22 +470,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   returned, and processing swallows its errors, so a store rejecting every
   event — read-only database, exhausted disk, a failed migration — still
   reported a fresh write time and looked healthy (#516).
-- Gave `ai-memory mcp-bridge` a TLS backend so it can reach an `https://` MCP
-  server. `ai-memory-cli` enabled rmcp's `transport-streamable-http-client-reqwest`
-  feature, which pulls reqwest in via `__reqwest = ["dep:reqwest"]` and stops
-  there — no `rustls`, no `native-tls`, no TLS at all. reqwest with no TLS
-  feature refuses any non-`http` scheme, so every bridge run against an HTTPS
-  server URL failed at connect with `invalid URL, scheme is not http`, and the
-  only URL that worked was plaintext `http://` — which is also the transport
-  the bridge attaches its bearer token to. Enabling rmcp's `reqwest` feature
-  the bridge now uses rmcp's `reqwest-tls-no-provider`
-  (`reqwest?/rustls-no-provider`), which supplies the platform certificate
-  verifier without pinning a crypto provider, and installs `ring` — already
-  compiled for this binary via the workspace's reqwest 0.12 — as the process
-  default. rmcp's plain `reqwest` feature would have pinned `aws-lc-rs`,
-  adding a compiled C dependency to every build for a bridge most users never
-  enable. The workspace-client half of #492 landed in #496; this is the MCP
-  transport, which is a second, independent copy of reqwest ([#497]).
 
 
 ## [1.33.0] - 2026-08-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   returned, and processing swallows its errors, so a store rejecting every
   event — read-only database, exhausted disk, a failed migration — still
   reported a fresh write time and looked healthy (#516).
+- Gave `ai-memory mcp-bridge` a TLS backend so it can reach an `https://` MCP
+  server. `ai-memory-cli` enabled rmcp's `transport-streamable-http-client-reqwest`
+  feature, which pulls reqwest in via `__reqwest = ["dep:reqwest"]` and stops
+  there — no `rustls`, no `native-tls`, no TLS at all. reqwest with no TLS
+  feature refuses any non-`http` scheme, so every bridge run against an HTTPS
+  server URL failed at connect with `invalid URL, scheme is not http`, and the
+  only URL that worked was plaintext `http://` — which is also the transport
+  the bridge attaches its bearer token to. Enabling rmcp's `reqwest` feature
+  adds `reqwest?/rustls`, which on reqwest 0.13 is `rustls-platform-verifier`,
+  so the bridge now verifies against the platform trust store like every other
+  outbound request in the workspace. The workspace-client half of #492 landed
+  in #496; this is the MCP transport, which is a second, independent copy of
+  reqwest ([#497]).
 
 
 ## [1.33.0] - 2026-08-28
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,6 +56,7 @@ dependencies = [
  "jsonc-parser",
  "reqwest 0.12.28",
  "rmcp",
+ "rustls",
  "secrecy",
  "serde",
  "serde_json",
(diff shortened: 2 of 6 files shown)


abhisheksharma2411:
Rebased onto `a2cd416` now that #496 has merged and v1.32.2 is cut — this had gone to `mergeable_state: dirty`.

Two things the conflict resolution needed beyond a mechanical merge, both worth stating:

**The CHANGELOG entry moved.** `[Unreleased]` was emptied by the v1.32.2 release, so a naive merge dropped this entry *inside the released section*. It's now back under `## [Unreleased]` where it belongs, and the reference is `#497` rather than `#492` — with a line saying the workspace-client half landed in #496 and this is the separate MCP copy, so the two entries don't read as duplicates of each other.

**`Cargo.lock` was regenerated rather than hand-merged.** Took main's lockfile and let cargo re-resolve, so the 164 added lines are the platform-verifier subtree entering the graph and nothing else.

**Re-verified on the new base rather than assuming the rebase carried it** — #496 changed the workspace client's TLS features, so the two could have interfered:

```
reqwest 0.12.28 (workspace, #496):
  rustls-tls-native-roots, rustls-tls-native-roots-no-provider, __rustls, __rustls-ring, __tls

reqwest 0.13.4 (rmcp, this PR):
  rustls, __rustls, __rustls-aws-lc-rs, __tls

rustls-platform-verifier v0.7.0
  └── reqwest v0.13.4 └── rmcp v1.7.0 └── ai-memory-cli
```

Both halves resolve independently and neither pulls webpki roots. Gates on the rebased head:

```
cargo build  -p ai-memory-cli                 Finished, 22.98s
cargo clippy -p ai-memory-cli --all-targets   no warnings
cargo test   -p ai-memory-cli                 103 passed, 0 failed, 1 ignored (9 binaries)
```

abhisheksharma2411:
Rebased onto `ec9bdaa` — v1.33.0 cut this morning renamed `## [Unreleased]`, so this had gone to `dirty` again. New head `a157715`.

Neat timing: the conflict was in exactly the class `#517` just added a guard for. So rather than eyeball it, I resolved the entry back under the current `[Unreleased]` and let your own check adjudicate:

```
$ bash scripts/check-changelog-frozen.sh
CHANGELOG: no released section was modified.
```

The changelog diff is additive only — no deletions anywhere, released sections untouched.

`cargo check -p ai-memory-cli` clean on the rebased head. The substantive change is unchanged from the last review pass: `rmcp` gains the `reqwest` feature so the streamable-HTTP transport actually has a TLS backend, with the reasoning inline in `Cargo.toml`.

akitaonrails:
The bug is real and I want it fixed — but please try a slimmer route before I take this, because the dependency cost is the sticking point rather than the change itself.

## Confirmed first

Your diagnosis holds. `transport-streamable-http-client-reqwest` pulls reqwest 0.13 in through `__reqwest = ["dep:reqwest"]` and stops, so that second reqwest has no TLS backend and refuses any non-`http` scheme. Building the exact feature set to prove it rather than inferring was the right call.

And removing TLS instead is not an option, which I checked before asking you to do more work: the session-aware bridge exists **for remote servers**, and `docs/https-via-proxy.md` tells operators to front ai-memory with Caddy. A bridge that cannot speak `https` breaks precisely the deployment we recommend, and the alternative — a TLS-terminating proxy on every client machine — is not a reasonable ask.

## What I would like you to try

Enabling rmcp's `reqwest` feature brings `reqwest?/rustls`, which on 0.13 is `__rustls-aws-lc-rs` + `rustls-platform-verifier`. That is where `aws-lc-rs`, `aws-lc-sys`, `cmake`, `jni` and friends come from — a C toolchain and an Android JNI stack in every build of a project that is otherwise pure Rust. For a bridge most users never enable, that is a large tax.

reqwest 0.13.4 offers a lighter door (from its own manifest):

```toml
rustls             = ["__rustls-aws-lc-rs", "dep:rustls-platform-verifier", "__rustls"]
rustls-no-provider = ["dep:rustls-platform-verifier", "__rustls"]
```

`rustls-no-provider` gives the platform verifier **without** pinning a crypto provider. And we already have one: `ring 0.17.14` is in the lockfile via `rustls 0.23.40`, pulled by the reqwest 0.12 the rest of the workspace uses. So the provider is free — it is already compiled.

That would need the process to install a default provider once at startup, something like `rustls::crypto::ring::default_provider().install_default()` before the bridge builds its transport, and a clear error if the install fails rather than a panic deep in a handshake.

**The open question is whether rmcp lets us get there.** If its `reqwest` feature hard-wires `reqwest?/rustls`, enabling it may drag aws-lc in regardless of what we ask for, and the escape would be depending on reqwest 0.13 directly with `rustls-no-provider` and letting rmcp use that resolved copy. I tried a quick local probe of that and hit a feature-resolution conflict I did not chase down — you are closer to this than I am.

## What decides it

If `rustls-no-provider` + ring works, I will take it happily. **If you establish that rmcp forces aws-lc and there is no way around it, say so and I will take your original patch** — a broken bridge for every HTTPS user is worse than a heavier build, and I would rather have the honest finding than a clever workaround that fights the ecosystem.

Either way please include in the PR: what the lockfile gains, and a note in `docs/https-via-proxy.md` that the session-aware bridge now reaches HTTPS endpoints — that is the doc a person hitting this reads first.

Thanks for the empirical write-up. Building the failing feature set rather than reasoning about it is what made this straightforward to act on.

abhisheksharma2411:
`rustls-no-provider` + ring works. Pushed as `4ef076c`.

**rmcp does not force aws-lc.** rmcp 1.7 already exposes the door you wanted, so the direct-reqwest escape hatch isn't needed — `rmcp-1.7.0/Cargo.toml:98`:

```toml
reqwest-tls-no-provider = [
    "__reqwest",
    "reqwest?/rustls-no-provider",
]
```

The branch now enables that instead of `reqwest`, adds `rustls` with `default-features = false, features = ["ring", ...]`, and installs the provider once before the transport is built. `default-features = false` is load-bearing there — rustls' own default set includes `aws_lc_rs`, so a plain `rustls = "0.23"` would have reintroduced exactly what we're removing.

## What the lockfile gains

Against `main`, measured rather than estimated:

| | packages added vs main | aws-lc |
|---|---|---|
| `reqwest` (what you reviewed) | 15 | `aws-lc-rs`, `aws-lc-sys` |
| `reqwest-tls-no-provider` (now) | 10 | none |

Gone: `aws-lc-rs`, `aws-lc-sys`, `cmake`, `dunce`, `fs_extra`. `grep -c aws-lc Cargo.lock` is now 0.

Still added: `rustls-platform-verifier`, `rustls-platform-verifier-android`, `webpki-root-certs`, `jni`, `jni-macros`, `jni-sys`, `jni-sys-macros`, `combine`, `simd_cesu8`, `simdutf8`.

## One correction to the premise, since it cuts against my own patch too

The JNI stack is **lockfile-only in both versions** — it isn't "in every build". `rustls-platform-verifier-0.7.0/Cargo.toml:126` gates it:

```toml
[target.'cfg(target_os = "android")'.dependencies.jni]
```

`Cargo.lock` is target-agnostic, so those crates are listed but never compiled off Android. Resolving the host graph:

```
$ cargo tree -p ai-memory-cli -e normal --target aarch64-apple-darwin
  jni          → absent, both versions
  aws-lc-sys   → present with `reqwest`, absent now
  cmake        → absent from the host build graph in both (build-deps too)
```

So the real saving is `aws-lc-sys` — a C library that genuinely compiles — not the Android crates, and not cmake. 277 → 275 crates in the host build. Smaller than the lockfile diff suggests, but it is the part that costs build time, and `rustls-platform-verifier` comes with `rustls-no-provider` regardless, so there's no version of this that drops it.

## Verification

```
$ cargo tree -f "{p} [{f}]" --target aarch64-apple-darwin
reqwest v0.12.28 [__rustls,__rustls-ring,__tls,json,rustls-tls-native-roots,...]
reqwest v0.13.4  [__rustls,__tls,json,rustls-no-provider,stream]
```

The 0.13 copy had `json,stream` and no `__tls` at all before — that absence was the bug. Note 0.12 resolves `__rustls-ring`, which is the ring you pointed at: it was already being compiled, so the provider really is free.

- `cargo test -p ai-memory-cli` — 722 + 103 across the integration binaries, 0 failed
- `cargo clippy -p ai-memory-cli --all-targets` — clean
- `cargo fmt --check` — clean
- `scripts/check-changelog-frozen.sh` — "no released section was modified"

New test `install_crypto_provider_is_idempotent_and_leaves_a_default` covers both calls succeeding and a default being present afterwards, since the second call models another component having installed one first.

## Docs

Added a **The session-aware MCP bridge and HTTPS** section to `docs/https-via-proxy.md`: the bridge reaches `https://` server URLs, a second local proxy per client machine isn't needed, and — worth saying plainly — it uses the platform verifier, so a self-signed cert or a private CA that isn't in the system trust store is rejected.

CHANGELOG entry rewritten to describe this route rather than the old one.

akitaonrails:
Approved, with one maintainer commit on top for a comment that overstated the benefit. Sorry this sat so long — the wait was on my side, and the revised route was worth it.

I audited the **diff**, not the description: the body still argues the `reqwest?/rustls` route, which `4ef076c6` replaced. Worth updating if you like, though it doesn't affect the merge.

## Verified rather than taken on trust

| claim | check |
|---|---|
| bridge's reqwest had no TLS backend | `json, stream` only |
| the fix compiles TLS in | `cargo tree -e features` → `__rustls` + `__tls` via `reqwest-tls-no-provider` |
| aws-lc-rs stays out | **0** entries in the merged lockfile |
| `ring` already present via reqwest 0.12 | 1 entry on main |

Gate on the merge with current `main`: `fmt`, `clippy -D warnings`, **2689 tests, 0 failures**, and `cargo deny` → `advisories ok, bans ok, licenses ok, sources ok` for the crates that enter.

The security framing is the part that made this worth doing properly. The bridge attaches a bearer via `config.auth_header(token)`, and with HTTPS structurally unavailable the only scheme that worked was the one that puts that token on the wire in cleartext — and no config key could add a TLS backend that was never compiled in. That is a real exposure, not a papercut.

The design is right, too: the provider is installed once *before* the transport is constructed, which matters because reqwest's no-provider path reads the process default at client-build time. The idempotency test covers the "someone else installed one first" case properly.

## The one correction

The Cargo.toml comment said the plain route "pins aws-lc-rs and drags in a C toolchain and an Android JNI stack", which reads as though this route avoids all three. It avoids the first two.

Both routes go through `rustls-platform-verifier` — your own body notes reqwest 0.13's `rustls` feature pulls it — and its dependencies include `jni` and `rustls-platform-verifier-android`. I confirmed against the merged lockfile: those crates enter *with this change*. They're target-gated and never compiled for our platforms, but they aren't a difference between the two options.

Avoiding aws-lc-rs is a strong enough reason by itself: it puts cmake/nasm in every build of this workspace, and we ship Windows binaries. I only tightened the claim so the next person weighing a TLS route here doesn't go looking for an option that doesn't exist.

## On what you declined to claim

You said you hadn't done a live handshake and wouldn't imply otherwise. That was the right call, and it's why I could audit this quickly — I knew which parts were measured. Nothing here needs it: the scheme rejection is decided by compile-time features, and those are now verifiably present.

Leaving `ai-memory-mcp` alone was also correct. Feature unification giving the server side the client TLS features is a consequence of the workspace shape, and reshaping that to avoid it would be a much larger change than this fix justifies.

akitaonrails:
CI caught a placement problem on the merge, so I pushed a second maintainer commit.

Your branch predates v1.38.0. The entry was written under `## [Unreleased]`, and merging main moved that heading down without moving the entry with it — git reanchored the addition by context and landed it inside the frozen `[1.33.1]` section, with no conflict to notice. `scripts/check-changelog-frozen.sh` exists for exactly this failure and is what failed the run.

Moved to `[Unreleased] → Fixed`, content unchanged. Merged main into your branch at the same time so the lockfile resolves against current `main`.

Gate on the merged result: `fmt`, `clippy -D warnings`, **2691 tests, 0 failures**, `cargo deny` clean. Merging once CI is green.

PR #499, opened 2026-08-26
feat(cli): add `ai-memory workstreams` checkout-local discovery

victorcesc:
## What it adds

`ai-memory workstreams` — a read-only listing of the managed workstreams that `run --workstream` can select from the current checkout, backed by a new `POST /workstream/recent` route.

```
$ ai-memory workstreams
Managed workstreams for default/ai-memory:
* default  56 minutes ago  [claude-code, open-code]
    id: 01a02b8d-31d9-7532-9f4f-e47e3642c4db
```

- `*` marks what a bare `ai-memory run` would resume; the rest follow by recent activity.
- Each row carries the linked harnesses and the stable id that `workstream-search --workstream-id` accepts.
- `--json`, plus `--workspace` / `--project` / `--limit` (1–100, default 20).
- The Docker wrapper intercepts it alongside `run` / `show` / `continue`, since repository identity is a host resource.

## Design notes

**Privacy.** Metadata only — checkout paths, repo/worktree fingerprints and native session ids are deliberately absent; the fingerprints are query *inputs*, not something a caller should get back. The endpoint test asserts their absence in the serialized body.

**Scope.** `Capability::NormalRead`, matching the sibling read routes, resolved through `lookup_existing_scope` — no-create, failing closed with 404.

**Query.** One statement, no N+1; `EXPLAIN QUERY PLAN` shows both accesses index-backed. `LIMIT` sits inside the CTE *before* the `LEFT JOIN` onto native sessions, so a workstream with many harnesses cannot inflate the row budget.

## Two decisions worth reviewing

1. **`is_current` as the leading ORDER BY key** — the selected workstream leads even when another has a newer `updated_at`. The right default for a selector list, I think, but pure recency is defensible. Pinned by a test that fails if reversed.
2. **The route name** — `/workstream/recent`, beside `/workstream/runs` and `/workstream/{id}/events`.

## Verification

- `TAILWIND_SKIP=1 cargo test --workspace` — 2593 passed, 0 failed
- `cargo deny check`
- `scripts/managed-workstream-acceptance.sh` — deterministic phase passed
- Live smoke against a real data dir: ordering, `--json`, `--limit`, worktree isolation, and the id chaining into `workstream-search`

Tests added at three levels: store (ordering, harness join, `--limit`, cross-workspace isolation), endpoint (worktree scoping, 404, no leaked session ids), CLI (renderer). The new store test and the acceptance block were both mutation-checked — each fails when the behaviour it pins is removed.

**Not verified:** the Tailwind asset build (skipped locally; no web assets changed), and `check-native-packaging.sh`, which needs `systemd-analyze` — wrapper routing is covered by `managed_host_commands_use_native_path_and_remote_server_without_docker` instead. Reaching my acceptance block on macOS also required excising its two pty adoption steps: the script is macOS-incompatible on two counts that predate this branch (`mapfile` needs bash 4+; `script -qefc` is util-linux syntax BSD rejects). Neither affects CI, which does not invoke it. Not proposing a fix here.

## Deliberately out of scope

- **No MCP tool.** This needs the caller's repository identity, which an MCP client cannot supply the way a host CLI can — such a tool would quietly answer about the wrong checkout. Tool count stays 18.
- **No selection or mutation.** Listing only; switching stays `run --workstream NAME`.
- **No migration.** Reads existing columns, so it is rollback-safe against any store at V49+.

CHANGELOG entry is under `[Unreleased]` → `### Added`, referencing (#499).

Context: I keep several workstreams open at once, roughly one per task in flight. Coming back to one was already possible — `opencode session list`, find the row, copy the opaque session hash, hand it to `ai-memory run` — and I used that for a while. It works, but you pay the hash-copying every single time you switch, which quietly discourages the branching that `--new` exists to support. Workstreams already have names I chose; nothing surfaced them. This lists them, so switching is `run --workstream <name>` instead of a hash hunt. It also spans harnesses, which a per-harness `session list` cannot.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Added `ai-memory workstreams`, a read-only checkout-local list of recent
+  managed workstreams with the current selection first, linked harnesses,
+  timestamps, and stable ids. Human-readable and `--json` output share the same
+  bounded POST query, which follows `run`'s exact workspace, project,
+  repository, and worktree identity without returning checkout paths,
+  fingerprints, or native session ids. The Docker shell wrapper routes the
+  command through its native host client so repository identity remains
+  correct. (#499)
+
+
 ## [1.33.1] - 2026-08-28
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -222,6 +222,9 @@ priors are at the [bottom](#influences-and-prior-art).
 
   # Kiro defaults to v2; select its incompatible v3 engine explicitly once.
   ai-memory run kiro --v3
+
+  # List the workstreams that can be selected from this checkout.
+  ai-memory workstreams
   ```
 
 - **"Pick the project instead of remembering where it lives."** Start from a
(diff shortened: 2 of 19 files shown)


victorcesc:
Could you add the `windows` label when you get a chance?

This does not change path handling itself, but it adds a new caller of `inspect_repository`, whose repo/worktree fingerprints are SHA-256 hashes of `cwd.canonicalize()` output and `git rev-parse --path-format=absolute` results — verbatim paths and git plumbing, the two areas `windows.yml` names as worth opting in before the merge rather than after.

victorcesc:
Heads up: I have `ai-memory rename-workstream` working on top of this branch. Names are fixed at `run --new` time and have no correction path today.

It selects by current name or by the stable id this PR prints, which is why it stacks — `workstreams` is currently the only surface that emits a workstream id.

I kept it out of this PR because the scope note says listing only. It is metadata-only: the ledger, managed runs, and linked sessions all key on the workstream id, and it leaves `selected_at` and `updated_at` untouched, so relabelling does not reorder this listing or change what a bare `ai-memory run` resumes.

Happy to open it separately once this merges, or fold it in here if you would rather review both together.

PR #514, opened 2026-08-28
feat(config): add EMBEDDING_API_KEY for embedding-only credentials

djalmajr:
## What changed

`EMBEDDING_API_KEY` — an optional, embedding-only credential resolved ahead of `OPENAI_API_KEY` and `LLM_API_KEY`.

- `openai` embedder: `EMBEDDING_API_KEY` → `OPENAI_API_KEY` → `LLM_API_KEY` (the last still only when `AI_MEMORY_EMBEDDING_BASE_URL` is set)
- `openai-compat` embedder: `EMBEDDING_API_KEY` → `LLM_API_KEY`, still keyless when neither is set
- `voyage` and `google`/`gemini` are untouched — they already name their own key
- Both `NotConfigured` messages name the new variable, since that error is where an operator hits the missing-key path

With the variable absent, key resolution is byte-identical to before. No existing install changes behaviour.

## Why

Embeddings are already independently configurable — `AI_MEMORY_EMBEDDING_PROVIDER`, `_MODEL`, `_DIM` and `_BASE_URL` each have their own setting. There was no key to go with them, and that one gap defeats the other four: `openai_embedding_api_key` returned `OPENAI_API_KEY` before the base-URL check ran, so pointing `AI_MEMORY_EMBEDDING_BASE_URL` at a second provider sent it whichever credential the chat model happened to use.

The existing `LLM_API_KEY` fallback only fires when `OPENAI_API_KEY` is absent — which also takes the `openai` chat provider down, since it reads that same variable. So there was no configuration that ran the LLM on `api.openai.com` and embeddings on a different OpenAI-compatible endpoint.

Reproduction on `main`:

```
AI_MEMORY_LLM_PROVIDER=openai
OPENAI_API_KEY=<openai key>
AI_MEMORY_EMBEDDING_PROVIDER=openai
AI_MEMORY_EMBEDDING_BASE_URL=<other openai-compatible endpoint>
AI_MEMORY_EMBEDDING_MODEL=<model on that endpoint>
```

The embedding request goes to the second endpoint carrying the OpenAI key and 401s, with nothing to override it.

That combination is the ordinary one for a local or cheaper embedding engine (Ollama, vLLM, LM Studio, OpenRouter, Workers AI) alongside a hosted chat model.

## Test plan

- [x] `cargo test --workspace` passes (65 suites, 0 failures)
- [x] Manual test: with only `OPENAI_API_KEY` set, resolution is unchanged; adding `EMBEDDING_API_KEY` sends it to the embedding endpoint while the chat provider keeps using `OPENAI_API_KEY`. Four unit tests cover the precedence order, the unchanged fallbacks, and both error strings.

## Commit attribution

## CHANGELOG (merge gate)

- [x] Added a `CHANGELOG.md` `[Unreleased]` entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `EMBEDDING_API_KEY`, an optional embedding-only credential resolved ahead of
+  `OPENAI_API_KEY` and `LLM_API_KEY`. Embeddings are already independently
+  configurable — `AI_MEMORY_EMBEDDING_PROVIDER`, `_MODEL`, `_DIM` and
+  `_BASE_URL` each have their own setting — but there was no key to go with
+  them, so pointing `AI_MEMORY_EMBEDDING_BASE_URL` at a second provider sent it
+  whichever credential the chat model happened to use. `openai_embedding_api_key`
+  returned `OPENAI_API_KEY` before the base-URL check ran, and the `LLM_API_KEY`
+  fallback only fired when `OPENAI_API_KEY` was absent — which also takes the
+  `openai` chat provider down, since it reads that same variable. `voyage` and
+  `google`/`gemini` were unaffected: they already name their own key.
+  Scoped to the two embedders that borrowed another role's key. `openai` now
+  resolves `EMBEDDING_API_KEY` → `OPENAI_API_KEY` → `LLM_API_KEY` (the last still
+  only with a custom base URL); `openai-compat` resolves `EMBEDDING_API_KEY` →
+  `LLM_API_KEY` and stays keyless when neither is set. With the new variable
+  absent, resolution is byte-identical to before. Both `NotConfigured` messages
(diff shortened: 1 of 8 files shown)


lucasliet:
today if you use LLM_API_KEY env to llm model you can use OPENAI_API_KEY for embedding, like
ai-memory:
    image: akitaonrails/ai-memory:latest
    environment:
      AI_MEMORY_LLM_PROVIDER: openai-compat
      AI_MEMORY_LLM_BASE_URL: https://openrouter.ai/api/v1
      AI_MEMORY_LLM_MODEL: moonshotai/kimi-k2.6
      LLM_API_KEY: ${OPENROUTER_API_KEY}
      #
      AI_MEMORY_EMBEDDING_PROVIDER: openai
      AI_MEMORY_EMBEDDING_BASE_URL: https://api.githubcopilot.com/embeddings
      AI_MEMORY_EMBEDDING_MODEL: text-embedding-3-small
      AI_MEMORY_EMBEDDING_DIM: "1536"
      OPENAI_API_KEY: ${COPILOT_OAUTH_TOKEN}
      
but it would be good to actually set one for embedding specifically

Issue #530, opened 2026-08-29
Windows: no native persistent-service story (Scheduled-Task + Start-Process wrapper gets silently killed by the Job Object)


gabrielscharb:
## Summary

There's no first-party way to run `ai-memory serve` persistently on **native Windows** (no WSL2, no Docker Desktop). Linux gets a proper daemon story via the packaged systemd units (`packaging/systemd/ai-memory.service` / `ai-memory-user.service`, both with `Restart=on-failure` + `RestartSec=5s`). Windows has nothing equivalent:

- `docs/windows.md` Scenario A is WSL2 (Docker inside Linux).
- Scenario B is Docker Desktop on native Windows, where all the resiliency (`--restart unless-stopped`) comes from Docker Desktop itself, not from `ai-memory`.
- Scenarios C and D (prebuilt binary / source build on native Windows) both just say to run `ai-memory.exe serve` **in a foreground terminal**. Nothing about keeping it alive across logons/reboots/crashes.

There's also no `windows-service` crate (or any SCM integration) anywhere in the workspace, so `ai-memory.exe` can't be registered directly as a Windows Service — the SCM expects a process that speaks its control-code protocol, and a plain console binary gets killed shortly after `sc create`/`New-Service` starts it.

## Concrete failure mode we hit

To fill the gap, we built an ad-hoc Scheduled Task ("run at logon") whose action is `powershell.exe -File start-server.ps1`, where the script does:

```powershell
$alreadyRunning = Get-NetTCPConnection -LocalPort 49374 -State Listen -ErrorAction SilentlyContinue
if ($alreadyRunning) { exit 0 }
Start-Process -FilePath $exe -ArgumentList @(...) -WindowStyle Hidden -RedirectStandardOutput ... -RedirectStandardError ...
```

This silently fails after every reboot. Root-caused it today:

1. Machine reboots (confirmed via `Win32_OperatingSystem.LastBootUpTime`).
2. Task Scheduler's logon trigger fires the task (`Microsoft-Windows-TaskScheduler/Operational` events 119/129/100/200 all timestamped together).
3. `start-server.ps1` runs, calls `Start-Process` to launch `ai-memory.exe` — it logs its one startup line (`ai_memory: ai-memory starting version=... bind=127.0.0.1:49374`).
4. In the **same second**, the Task Scheduler operational log records the task instance completing successfully (event 201/102, return code 0) — because `Start-Process` is non-blocking and the wrapper script has nothing left to do.
[...]

PR #531, opened 2026-08-29
feat(admin): add read-only GET /admin/audit-log

djalmajr:
## What

Adds `GET /admin/audit-log`, a read-only paginated reader for the `audit_log`
table that has existed since `V01__init.sql` (plus `author_id` in V16) but had
no HTTP surface. Answering "who expired alice's token last week?" or "which user
soft-deleted that gotcha page?" required querying SQLite directly.

## Shape

```
GET /admin/audit-log?workspace=&project=&op=&before_id=&limit=
-> { "events": [ { id, at, op, workspace, project, page_path, author_username, detail } ] }
```

- Keyset cursor (`before_id`, `id DESC`) rather than offset, so a live trail
  cannot skip or duplicate rows while new events land mid-pagination.
- `limit` clamps to `1..=200`, default 50.
- Workspace/project filters are by name, matching the other scoped readers.

## Design notes

- **LEFT JOINs, `Option` everywhere.** `audit_log` deliberately has no foreign
  keys on `workspace_id`/`project_id`/`page_id` (V05 documents this: append-only
  event log, orphan rows are expected and acceptable). Resolved names are
  therefore nullable rather than assumed present.
- **`detail` is returned but not advertised.** The single writer
  (`ops::audit`) inserts the literal `'{}'` on every call, so the column carries
  no information today. It is exposed for schema fidelity; the CHANGELOG entry
  says so explicitly rather than implying a payload.
- **Same capability as its neighbours.** Registered in the same router group and
  gate as `audit-contamination`, `open-sessions` and `checkpoints` — no new
  capability introduced.
- **Not an MCP tool**, so `MEMORY_INSTRUCTIONS`, `ai_memory_core::SNIPPET_BODY`
  and the documented tool count are untouched.

## Tests

- store: name resolution through all four joins; anonymous write yields
  `author_username: None`; `before_id` pages deterministically; `limit` clamps.
- mcp: route appears in the existing admin route inventory test; behavioural
  test through the router pins the JSON shape and the scope filter.

## Gates

Run on this branch, rebased onto `main` at v1.36.0:

- `git diff --check` — clean
- `TAILWIND_SKIP=1 cargo test --workspace` — **2631 passed, 0 failed**
- `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings` — clean

## Note

The CHANGELOG entry references `(#530)` as a placeholder; happy to amend it to
this PR's real number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Added `GET /admin/audit-log`, a read-only paginated reader for the existing
+  `audit_log` table. Events resolve workspace, project, page path, and author
+  username through LEFT JOINs (orphans stay `null` — the log has no foreign
+  keys by design), page by keyset (`before_id`) so a live trail cannot skip or
+  duplicate rows, and clamp `limit` to 1..=200 (default 50). The `detail`
+  column is returned for schema fidelity; the only writer still stores the
+  literal `{}`. (#531)
+
 ## [1.36.0] - 2026-08-29
 
 ### Added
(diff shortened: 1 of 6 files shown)


djalmajr:
CHANGELOG amended to reference this PR (#531) instead of the placeholder.

Issue #511, opened 2026-08-27
Add ZCode as an MCP client (install-mcp --client zcode)


bcosta19:
## Feature

Add [ZCode](https://z.ai) as a supported MCP client: `ai-memory install-mcp --client zcode`.

## ZCode MCP configuration facts (verified against the ZCode client)

- **User-scope config file:** `~/.zcode/cli/config.json`, server map under the nested key `mcp.servers` — the same shape already supported for the `zero` client (`JsonMcpLocation::NestedMcpServers`).
- **Workspace scope** (`.zcode/config.json`, `zcode.json`, `.agents/mcp.json` fallback) also exists; the user-scope file is the natural install default, consistent with other clients.
- **Entry schema is strict:** `type: "http"` + `url` + optional `headers`/`enabled`/`timeoutMs`. Any unknown key causes the server to be dropped silently — so the generated entry must be exactly `{ "type": "http", "url": "...", "headers": { "Authorization": "Bearer ..." } }`.
- **HTTP transport is streamable/stateless-friendly**: ai-memory's default stateless `/mcp` endpoint works as-is; bearer auth goes in `headers`.
- Servers from every scope auto-connect at session start.

## Scope

- `McpClient::Zcode` variant (name `zcode`) in `cli.rs`.
- Default config path `~/.zcode/cli/config.json` in `mcp_config_path`.
- `JsonMcpLocation::NestedMcpServers` + generic entry builder arm.
- `render_zcode` snippet.
- Uninstall: sweep list + JSON pointer `["mcp","servers"]`.
- `infer_installed_mcp_config` arm in `install_hooks.rs`.
- Tests (render matches, auth token threading, shape keys, apply preserves siblings + idempotent, CLI parse, uninstall round-trip) and docs (`docs/mcp-install.md` section + one-shot list, README support matrix, `docs/install.md`, CHANGELOG).

akitaonrails:
Grouping my answer across #511 and #512 since they are the two halves of the same request.

Adding an agent is a well-worn path here — #476 added Pool recently and is the template: an `AgentKind` variant, a forward migration expanding the `sessions.agent_kind` CHECK constraint, hook-payload extraction for the harness's own field names, a `hooks/<agent>/` bundle, installer wiring, and docs. It is mechanical but not small, and the migration makes it irreversible-ish: once a store has `zcode` rows, rolling back the binary is awkward.

[...]

PR #532, opened 2026-08-29
feat(hooks): add ZCode (z.ai) lifecycle-hook integration (#512)

### LINKED ISSUE #512: Add ZCode lifecycle hooks (install-hooks --agent zcode)
## Feature

Add ZCode lifecycle-hook support so ZCode sessions get automatic capture and SessionStart handoff injection: `ai-memory install-hooks --agent zcode`.

Depends on `install-mcp --client zcode` landing first (`mcp_client_for_agent` needs `McpClient::Zcode`).

## ZCode hooks facts (verified against the ZCode client)

- **Config-file hooks** live in the same `~/.zcode/cli/config.json` under `hooks`, shaped as `{ "enabled": true, "events": { "<Event>": [ { "matcher"?: ..., "hooks": [ ... ] } ] } }`. Two differences from Claude Code's shape: an extra `events` nesting level, and the runner is disabled unless `hooks.enabled: true` is set.
- **Exactly seven events:** `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PermissionRequest`, `PostToolUse`, `PostToolUseFailure`, `Stop` (Claude-Code-compatible names). No `SessionEnd`, `PreCompact`, or subagent events → same finalize-session story as Codex/Antigravity (`ai-memory finalize-session --agent zcode`). `PermissionRequest` is a permission gate, not a capture event — skip it.
- **`type: "process"` hooks** run an argv vector without a shell — the most portable form for the native `ai-memory hook --event <e> --agent zcode` command (same exec-form approach as Zero/Command Code), cross-platform including Windows.
- **Hook stdout is consumed:** JSON output with `additionalContext` is injected into the conversation (strict output schema) — the same envelope Claude Code uses, so `AgentKind::session_start_injects_handoff()` defaulting to `true` gives automatic handoff injection for free.
- **Payloads are Claude-Code-shaped** (stdin JSON, `session_id`/`cwd`/`tool_name`/`tool_input` keys), so existing sanitizer/capture-policy paths should apply unchanged; needs a live acceptance test to confirm.

## Scope

- `AgentKind::Zcode` (`zcode`) in `ai-memory-core`: variant, `ALL`, `as_str`, `from_wire`, round-trip tests.
- Store migration extending the `sessions.agent_kind` CHECK (table rebuild, next free number).
- `AgentChoice::Zcode` + `kind()` arm.
- `ZCODE_EVENTS` + a new `HookShape` variant covering the `hooks.events.<Event>` nesting and the `enabled: true` requirement.
- `install_hooks.rs`: config path, apply/render arms, `mcp_client_for_agent` → `Some(McpClient::Zcode)`; native process-form commands (no `.sh`/`.ps1` bundle).
- `uninstall.rs` hook-file plan + `setup_agent.rs` arm; `finalize-session` works via `AgentChoice`.
- Tests + docs (README support matrix hooks column, `docs/install.md`, `docs/mcp-install.md`, CHANGELOG) + a live-client acceptance test.

bcosta19:
Closes #512

## What changed

Adds ZCode (z.ai) lifecycle-hook integration (`install-hooks --agent zcode`, alias `zai`), following the Zero/Command Code model: exec-form native commands (`type: "process"`, no shell) merged into the root `hooks` block of `~/.zcode/cli/config.json` around third-party hooks, idempotent via the `statusMessage: "ai-memory capture"` ownership marker.

- `AgentKind::Zcode` (wire `zcode`, alias `zai`) in `ai-memory-core`, added to `ALL`, serde, and `from_wire`.
- V51 forward migration expanding the `sessions.agent_kind` `CHECK` constraint and scope-pairing triggers with the `zcode` value without losing rows.
- Hook ingestion recognizes ZCode's dual casing (camelCase native + Claude Code-compatible snake_case `session_id` / `cwd` / `tool_name` / `tool_input` aliases) for session attribution and closed tool-family titles, and enforces `[capture] ignore_paths` for file tools.
- `PostToolUseFailure` is handled alongside `PostToolUse`, mapping payload `error` / `error_details` to the `ToolOutcome::Error` observation outcome.
- Six documented triggers wired: `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `PostToolUseFailure`, `Stop` (verified live against engine v0.16.5).
- `install-hooks --agent zcode --apply` merges into `~/.zcode/cli/config.json` in place, preserving existing third-party hooks and custom `maxOutputBytes` (raises to 64 KiB if unset/default to prevent handoff truncation).
- `SessionStart` stdout injection is enabled (`hookSpecificOutput.additionalContext`), delivering the prior session's handoff automatically.
- `finalize-session --agent zcode` closes sessions (no true `SessionEnd` — `Stop` is per-turn).
- Docs: `README.md` support matrix, `docs/install.md`, and `CHANGELOG.md`.

## Why

Enables automatic cross-session memory capture and SessionStart handoff injection for developers using ZCode / Z.AI coding agents with `ai-memory`.

## Test plan

- [x] `cargo fmt --all -- --check` passes cleanly
- [x] `git diff --check` passes cleanly
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes (0 warnings)
- [x] `cargo test --workspace` passes (all unit and integration tests)
- [x] `cargo test --manifest-path companions/ai-memory-importer/Cargo.toml` passes (19/19 tests)
- [x] `tests/hooks/test_lib.sh` passes (46/46 tests)
- [x] `scripts/check-native-packaging.sh` passes
- [x] Live engine validation: real turn execution across mock, headless CLI (`zai/glm-5.1`), and app-server protocol against embedded engine v0.16.5 (`zcode-bin 3.10.1-1`).

## Commit attribution

## CHANGELOG (merge gate)

- [x] I added a `CHANGELOG.md` `[Unreleased]` entry under `### Added` referencing `#512`.

## Notes for reviewers & Contract evidence (observed, not guessed)

Tested against Arch Linux (`zcode-bin 3.10.1-1`, embedded engine `/usr/lib/zcode/glm/zcode.cjs` v0.16.5) across 3 capture phases:
1. **Mock** (runs 1–8): local Anthropic-compatible mock server (`127.0.0.1:8971`).
2. **Real model, headless CLI** (runs 9–10): `zai/glm-5.1` via `https://api.z.ai/api/anthropic` with real turns.
3. **App-server (desktop surface)** (runs 11–12): engine in `app-server` mode driven by a ZCode Protocol client.

All raw stdin payloads and protocol logs were captured and verified.

### Answers to the contract questions from #512:

| Question | Verified Answer |
|---|---|
| **Which lifecycle events does it fire, under what names?** | `SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PermissionRequest`, `PostToolUse`, `PostToolUseFailure`, `Stop`. |
| **What is the JSON payload shape on stdin?** | Dual casing on every turn: native camelCase (`toolName`, `toolInput`, `sessionId`) + Claude Code snake_case aliases (`tool_name`, `tool_input`, `session_id`, `cwd`). |
| **Where does hook config live?** | User scope: root `hooks` block in `~/.zcode/cli/config.json` (`{ "enabled": true, "events": { ... } }`). |
| **Does it have a true session-end?** | No — `Stop` fires at the end of every turn. Finished sessions are closed via `ai-memory finalize-session --agent zcode`. |
| **Is `SessionStart` stdout injected into model context?** | **Yes.** Injected as `<system-reminder>` context block in the first user message to the model (`hookSpecificOutput.additionalContext`). Live-tested with canary `ZCODE-EVIDENCE-CANARY-2026-08-28`. |

### Key Findings & Contract Behaviors:

1. **`SessionStart` Stdout Injection Canary:**
   The canary output `{"hookSpecificOutput":{"hookEventName":"SessionStart","additionalContext":"ZCODE-EVIDENCE-CANARY-2026-08-28"}}` was injected into the prompt context:
   ```
   <system-reminder>
   SessionStart hook additional context:
   #1
   ZCODE-EVIDENCE-CANARY-2026-08-28
   </system-reminder>
   ```

2. **`PostToolUseFailure` Handling:**
   When a tool execution throws (e.g. invalid filesystem access), `PostToolUseFailure` fires instead of `PostToolUse`. The payload carries `error` (string) and `error_details` (`{message, type}`), which `ai-memory-hooks` maps to `ToolOutcome::Error`.

3. **`PermissionRequest` Race Condition:**
   `PermissionRequest` is deliberately not installed. In the ZCode engine, `PermissionRequest` hooks run concurrently with the interactive permission client (`racePermissionResponders`). If the client responds instantly, the losing hook is aborted, making passive capture unreliable by design.

4. **Runner `maxOutputBytes`:**
   ZCode defaults `maxOutputBytes` to 32 KiB. The installer raises this to 64 KiB only when unset or default, ensuring a retrieved handoff and briefing are not truncated before context injection.

## Relation to #511 / PR #529
PR #529 adds the MCP client configuration. Both PRs share the same user configuration file (`~/.zcode/cli/config.json`) and can land independently without conflicts.


akitaonrails:
Audited. This is strong work and I want to land it — the migration and the merge logic are exactly right, and I verified more of your contract independently than I expected to:

- **V51** is structurally identical to V50 apart from the CHECK list — table definition, all five indexes, all three triggers — and the list is a strict superset, 20 → 21 with nothing dropped. That is the precise hazard V28's own comment warns about, and you avoided it.
- `AgentKind::Zcode.as_str() == "zcode"` matches the CHECK string exactly, `ALL` is bumped to 21, and the store tests that iterate `AgentKind::ALL` will catch any future enum/DB drift.
- The config merge is scoped the way editing a shared user file has to be: siblings (`model`, `provider`, `mcp.servers`) survive, `hooks.enabled` is defaulted but never flipped, and you never touch matcher groups you didn't write.
- Against z.ai's own docs I confirmed the root `hooks` block shape, `enabled`/`timeoutMs`/`events`, **and** `maxOutputBytes` (I went looking for that one expecting it to be invented — it's documented).
- Local gate on `2f5c7f24`: fmt clean, `clippy -D warnings` clean, 2639 tests, 0 failures.

**One question before I merge, and it is the only thing holding this.**

`session_start_injects_handoff()` returns `true` for ZCode. That is not a cosmetic flag: the native hook *fetches* the handoff, and the fetch is destructive — the server marks it accepted. If ZCode does not actually surface that stdout, every ZCode user silently loses handoffs, one at a time, unrecoverably. That asymmetry is why the predicate's own doc says unproven agents fail safe.

Your canary evidence is specific and I'm inclined to believe it. What I could not confirm from public sources is the **surface**:

- z.ai's docs confirm `additionalContext` is injected into model context, but do not state it for `SessionStart` specifically on the **native agent**.
- [zai-org/feedback#32](https://github.com/zai-org/feedback/issues/32) reports that hooks configured in `~/.zcode/cli/config.json` do not fire for the native ZCode Agent at all — working for external CLI subagents but not the native one. It is closed as duplicate / not planned with no visible resolution, so I can't tell whether it's stale, user error, or live.

You mention capture on three surfaces including "app-server/desktop". If the canary landed in the model's context on the **native agent** — not only the headless CLI — then that is the missing piece and I'll merge on your word plus whatever excerpt is convenient. If your capture was headless-only, I'd rather land this with the flag `false` and flip it in a follow-up once confirmed: ZCode users still get the handoff on demand via `memory_handoff_accept`, and nothing is silently destroyed while we find out.

Also worth knowing, not a change request: the predicate is a negative list, so a new enum variant gets `true` by simply existing, while the doc above it says unknown agents fail safe. Unknown *wire values* do fail safe via `Other`; a new *variant* does not. That's pre-existing and not yours to fix here — I'm noting it because your PR is the first to depend on it.

One housekeeping note: #529 landed the MCP half, so this will need a merge for `CHANGELOG.md` and `docs/install.md`. Happy to do that myself once the question above is settled.

akitaonrails:
Withdrawing my question — I went and read z.ai's official docs instead of making you produce the capture dossier, and they answer it. Your flag is right, and the docs turn out to corroborate more of this PR than I expected.

## What the docs establish

**1. `SessionStart` → `additionalContext` is injected.** [zcode.z.ai/en/docs/hooks](https://zcode.z.ai/en/docs/hooks) documents the contract and gives the exact envelope:

```json
{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "…"}}
```

`session_start_handoff_envelope` (`hook.rs:435`) emits precisely that shape. So `session_start_injects_handoff() == true` is correct, and I can confirm it without your capture.

**2. The envelope is load-bearing, and you used it.** The docs are explicit that *"plain stdout without JSON structure is treated as diagnostics only and does not enter model context."* Had ai-memory printed the raw handoff body for ZCode the way it does for Kimi Code, the fetch would have consumed the handoff destructively and ZCode would have discarded it — the exact silent-loss failure I was worried about. It doesn't, because the envelope is the default path and only Antigravity opts out.

**3. User scope isn't a preference, it's the only option.** *"Project-level hooks are not executed in the current version. For security reasons, any `hooks` configuration in `<workspace>/.zcode/config.json` or `<workspace>/zcode.json` is ignored as a whole."* Your `~/.zcode/cli/config.json` choice is the only one that works.

**4. The seven events match** the documented set exactly, in the documented order.

**5. `maxOutputBytes` — I owe you credit for this one.** I initially flagged it as a key you might have invented. It's documented, the default cap is 32768 bytes, and your comment says you raise it *"so a fetched handoff is not truncated mid-JSON before injection."* A handoff over 32 KiB would be cut mid-envelope, the JSON parse would fail, and the handoff would be gone — silently, and only for large ones. That is a failure mode I did not think of and would not have caught in review.

## On feedback #32

I raised [zai-org/feedback#32](https://github.com/zai-org/feedback/issues/32) (native agent not firing hooks) as the blocker. I no longer think it applies:

- The docs state **"The file must set `hooks.enabled: true`"**. Your merge does exactly that when the key is absent, and warns rather than flipping it when a user has explicitly set `false`. That requirement is the most economical explanation for a report of hooks silently never firing — and for it being closed as not-planned/duplicate rather than fixed.
- #32 was filed 2026-06-22. ZCode **3.8.1** (2026-08-20) shipped hooks work, and the current release is **3.10.1** (2026-08-28) — which is the build you tested. The report predates two months of hooks development.
- The hooks documentation draws **no distinction** between the native agent and CLI surfaces; hooks are described as session-general throughout.

Documentation, your live capture, and the code all agree. That's enough for me to merge on.

## Merged with maintainer commits

`CHANGELOG.md` and `docs/install.md` both needed a merge — #529 landed the MCP half and v1.37.0 shipped in between, so main already lists ZCode in the client table. Your entry moved to `[Unreleased]`; the released sections are untouched.

Thank you for the rigour here, and for the split into #529/#532 — landing MCP independently is what let me verify the config path against a shipped feature before the hooks half went in. The `maxOutputBytes` detail in particular is the kind of thing that only comes from actually running the thing.

bcosta19:
Apologies for not replying before the merge. I was assembling a careful response because my first packaged-desktop test appeared to contradict the documented behavior. I was preparing evidence to recommend the fail-safe false value rather
 than let the PR merge with an unproven desktop path.

 The missing detail was on my side: I had refreshed the ZCode hook configuration with the new ai-memory binary, but I had forgotten to upgrade the actual systemd server. The hooks were running through a local ai-memory 1.37.0 binary while
 /usr/bin/ai-memory and the active server were still on 1.34.0.

 The desktop was emitting the hook events, but they remained in the local hook spool because the old server had not applied V51 and could not ingest agent_kind = zcode. I saw an empty database and incorrectly attributed it to the desktop
 not firing hooks.

 I have now corrected the installation and repeated the test:

 - upgraded ai-memory-bin from 1.34.0 to 1.38.0;
 - restarted the server and confirmed V51__sessions_zcode_agent_kind was applied;
 - refreshed the ZCode hooks to use /usr/bin/ai-memory;
 - confirmed hooks.enabled: true, all six installed events, maxOutputBytes: 65536, and the existing MCP configuration preserved;
 - first verified the embedded engine against an isolated mock model;
 - then launched a completely fresh ZCode 3.10.1 Electron process and ran a real tool-using turn.

 The packaged desktop persisted the complete sequence:

 ```text
SessionStart
UserPromptSubmit
PreToolUse
PostToolUse
Stop
 ```

 The spool drained cleanly afterward.

 So no follow-up correction is needed for #532. Your conclusion from the official contract was correct, and the merged implementation works on the packaged Electron desktop as intended. Thank you for doing the additional verification, and
 apologies for the delayed response and the false lead caused by my outdated local server.

PR #533, opened 2026-08-30
feat(admin): add password sessions and separated API credentials

djalmajr:
## Summary

- add password-based admin login, recovery, revocable web sessions, CSRF protection, and root/user lifecycle management
- separate native `aim_` API credentials from human identities while preserving brownfield token hashes byte-for-byte
- add private session introspection for the `mcp-auth` edge and keep machine-only routes Bearer-only
- update CLI, admin routes, architecture, install, usage, and user documentation

## 1.x compatibility decision

- keep deprecated GET-only HTTP Basic / `ai_memory_auth` browser compatibility until a human password or completed bootstrap activates human auth
- preserve `user add|expire|revive|rotate-token` and their admin endpoints as deprecated 1.x shims
- keep V53 mirror triggers load-bearing until those shims are removed in 2.0

## Migrations

- V52 adds human roles, password state, bootstrap state, and web sessions; its scoped FK guard rejects author-reference damage without blocking unrelated pre-existing violations
- V53 creates native API credentials and mirrors the deprecated user-token write path for rollback and 1.x compatibility

## Verification

- `git diff --check`
- `TAILWIND_SKIP=1 cargo test --workspace` — 2705 passed, 2 ignored
- `cargo deny check`
- `cargo test --manifest-path companions/ai-memory-importer/Cargo.toml` — 19 passed
- focused V52/V53 migration tests — 13 passed
- `scripts/check-changelog-frozen.sh upstream/main`
- staged gitleaks scan — no findings


akitaonrails:
Approved after a full audit, with one maintainer commit for a blocking collision.

## The blocker: migration version collision

`V52__purged_sessions_tombstones` and `V53__page_embed_failures` landed on main after this branch opened. The merge produced **four migration files across two version numbers**, and git reports no conflict because the filenames differ. It surfaces only at runtime:

```
UNIQUE constraint failed: refinery_schema_history.version
```

on a **fresh database** — the merged tree could not open a store at all. I reproduced that before touching anything.

Renumbered `human_auth` → V54 and `api_credentials` → V55, content unchanged. That broke 12 tests that pin version numbers, which is the useful half: `run_to`/`open_to` targets, `schema_version` assertions, the rollback assertion, and the test names all shifted. The pre-migration fixture point moves 51 → 53, and I checked that is safe — main's V52/V53 create `purged_sessions` and `page_embed_failures`, touching neither `users` nor any table your migrations rewrite.

Worth noting for next time: refinery failing loudly here is the good outcome. A runner that skipped the duplicate instead would have left a store half-migrated with no error.

## Security review

The primitives are right, and several choices are better than the minimum:

- **Argon2id** at `m=19456, t=2, p=1` with a 16-byte OS-CSPRNG salt and PHC storage.
- **The KDF runs off the writer thread** behind a 2-permit semaphore with a 32-deep queue. Reserving the slow KDF for passwords while tokens stay SHA-256+pepper for the hot-path unique lookup is the correct split, and the comment says so.
- **Dummy verify for unknown and disabled users**, so "no such user" costs what a real miss costs.
- **Session secrets are CSPRNG and stored as SHA-256**, never raw, with a CSRF hash alongside and an explicit `expires_at`.
- **Cookies**: `HttpOnly` on the session, deliberately *not* on the CSRF cookie for double-submit, `SameSite=Strict`, `Secure` gated on config — with tests asserting each flag rather than trusting them.
- **`subtle::ConstantTimeEq`** on both bearer comparisons.

The proxy path is where this class of change usually goes wrong, and it is built correctly: identity headers are honoured only after the caller proves a dedicated `actor_proxy_bearer`, and the browser Basic/cookie path passes `allow_proxy=false`, so a session can never assert an arbitrary identity through headers. That separation is the whole ballgame and it is explicit in the signature.

I controlled the central guarantee rather than reading it: reintroducing the fall-through so a rejected bearer resolves to anonymous fails `rung3_unknown_bearer_with_multiuser_returns_401_not_anonymous`. The guard has teeth.

## Gate

`fmt`, `clippy -D warnings`, **2751 tests, 0 failures**, both changelog guards, on the merged result against current `main`.

## One process note

53 files and ~5,000 lines spanning migrations, the auth middleware, a new 2,151-line module, the CLI and the web mount is a lot to land as one unit — the collision above is a symptom, since a long-lived branch of this size is exactly what a release cut can invalidate underneath. Not asking you to split it now; the work is coherent and reviewed. But for future changes in this area, landing the schema and the store layer ahead of the HTTP surface would shorten the window where main can move under you.

PR #534, opened 2026-08-30
feat(cli): add interactive workstream resume picker

pvpmartins:
## What changed

- Added `ai-memory resume`, an interactive selector for recent managed workstreams from the current checkout and validated paths in the client-local project registry.
- Up/Down (or `j`/`k`) moves between workstreams. Left/Right cycles the highlighted row through `auto` and supported harness executables found in the host `PATH`; each row remembers its harness choice while navigating.
- Enter delegates the displayed workstream/harness combination to the existing managed `run` path. `auto` remains the default and preserves existing newest-usable-session discovery.
- Reused the checkout-local `/workstream/recent` query, path/scope revalidation, and terminal picker. Updated the Docker host wrapper, README, managed-workstream/install/architecture docs, packaging coverage, and `[Unreleased]` changelog.

## Why

#499 made managed workstreams discoverable, but selecting one still required copying its name into a separate `run --workstream` command from the correct checkout. This adds the `/resume`-style flow: choose a named workstream from anywhere and optionally continue it in another installed harness without exposing host paths to the server.

## Test plan

- [x] `cargo test --workspace` passes (`752` CLI tests passed, `1` ignored; every workspace crate suite passed)
- [x] `cargo deny check` passes (only the existing unmatched-license allowance and yanked transitive `chacha20` warnings)
- [x] Manual test: built the optimized v1.38.0 binary, opened `ai-memory resume` against a real local workstream, observed Right change `auto` to `claude`, Left change it back to `auto`, and pressed `q`; the picker exited `0` without launching a harness.

Focused tests cover Left/Right key mapping, cycling and wraparound in both directions, the `auto` sentinel, explicit harness resolution, per-workstream remembered choices, row rendering, parser constraints, ordering, and terminal-text sanitization.

## Commit attribution

- [x] I verified the name and email on every commit in this PR and corrected any unintended identity before requesting merge. The single commit uses `Paulo Vitor Martins <85913786+pvpmartins@users.noreply.github.com>`.

## CHANGELOG (merge gate)

- [x] I added a `CHANGELOG.md` `[Unreleased]` entry for the new user-facing command and keyboard behavior.
- [x] No default changed: each row starts at `auto`, matching the existing automatic managed-run selection behavior.

## Notes for reviewers

The server still never returns or receives a host checkout path. Discovery joins the current checkout and this client's private `client-projects.json` links to the existing fingerprint-scoped listing endpoint, then canonicalizes each path and verifies its resolved workspace/project before querying or launching. Duplicate workstreams reached through both discovery sources are removed by stable workstream id.

The horizontal behavior is an opt-in extension of the shared terminal selector: existing `show` selection remains vertical-only, while `resume` supplies the Left/Right handler and footer hint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -326,6 +326,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and a session with nothing else to name drops the clause instead of filling
   it. The predicate lives beside the writer and is derived from the same serde
   representation, so a new `ToolFamily` variant cannot escape it. (#527)
+
 ## [1.36.0] - 2026-08-29
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- Added `ai-memory resume`, an interactive picker for recent managed
+  workstreams from the current checkout and validated client-local links. It
+  launches the selected named workstream with the established automatic harness
+  selection while the server continues to receive only repository/worktree
+  fingerprints, extending the checkout-local discovery introduced by
+  `workstreams`. Left/Right cycles `auto` and the supported harnesses found in
+  `PATH`, so the selected workstream can be continued directly in another agent.
+  (#499)
+
 - `ai-memory handoffs --expire-all --confirm` and `POST /admin/handoffs/expire`
   clear a stale open-handoff backlog for one scope. The listing added in
   v1.35.0 made a backlog visible and gave `memory_handoff_cancel` the ids it
@@ -317,7 +326,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and a session with nothing else to name drops the clause instead of filling
   it. The predicate lives beside the writer and is derived from the same serde
   representation, so a new `ToolFamily` variant cannot escape it. (#527)
-
 ## [1.36.0] - 2026-08-29
 
 ### Added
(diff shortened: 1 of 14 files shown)


akitaonrails:
Approved, with one maintainer commit for a CI blocker.

**The blocker:** the branch dropped a blank line separating the end of `[1.37.0]` from the `## [1.36.0]` heading. Whitespace only, but it sits inside an already-released section, so `check-changelog-frozen.sh` rejects it — correctly, since the guard cannot distinguish a stray edit from a rewritten entry and should not try. Restored; your own entry was correctly placed at the top of `[Unreleased] → Added`.

GitHub was reporting this as conflicting, but your head already sits on current `main`, so no merge was needed.

## Verified

**The path-privacy claim holds.** `list_for_checkout` sends `workspace`, `project`, `repo_fingerprint`, `worktree_fingerprint`, `limit` — no path in the body or the query. The checkout path stays client-side and only feeds local fingerprinting, so joining the client-local registry to a fingerprint-scoped endpoint keeps the invariant intact across a command that now spans multiple checkouts.

**Terminal sanitization is load-bearing, not decorative.** Names and error text both go through `terminal_text` before rendering. I controlled it: dropping the sanitizer from the label construction fails `choice_sanitizes_server_control_characters`. Passing error strings through it too is the part that is easy to forget — `{error:#}` can carry server-supplied content into a raw-mode terminal.

Delegating Enter to the existing managed `run` path rather than spawning a harness in `resume.rs` is the right call: it keeps process execution on one audited surface instead of adding a second.

Gate: `fmt`, `clippy -D warnings`, **2705 tests, 0 failures**, both changelog guards.

## One observation, no action needed

`terminal_text` maps `char::is_control()` to `?`, which covers `ESC` and therefore ANSI sequences — the thing that matters. It does not catch Unicode bidi overrides (`U+202E` is category `Cf`, not `Cc`), so a name containing one could visually reorder a picker row. That is pre-existing in `show` rather than anything you introduced, the names are checkout-scoped, and selection is positional, so the practical risk is slight. Mentioning it because you clearly thought about this surface.

PR #535, opened 2026-08-31
docs(windows): flag the Job Object autostart trap, add Scenario E (WinSW)

### LINKED ISSUE #530: Windows: no native persistent-service story (Scheduled-Task + Start-Process wrapper gets silently killed by the Job Object)

gabrielscharb:
## Summary

Implements the direction you gave in #530: **(c)** as priority, **(a)** as the primary fix, WinSW over NSSM.

- **Known Trap** section in `docs/windows.md`, placed right after the "Rule Of Thumb" block (high in the doc, not buried) and cross-referenced from "Current Harness Caveats" — both a top-to-bottom read and a scan for gotchas land on it. Names the job-object teardown and the `LastTaskResult = 0` green-status symptom explicitly, plus how to confirm it after the fact via Task Scheduler event IDs 200/201 against the server's own startup log line.
- **Scenario E** in `docs/windows.md` — WinSW wrapping the existing binary as a real Windows Service, `<onfailure action="restart" delay="5 sec"/>` as the native analog of the systemd units' `Restart=on-failure`. No Rust changes, no new dependency.
- `CHANGELOG.md` entry under `[Unreleased]` / `### Docs`.

## Testing

Per your ask, here's exactly what was tested vs. documented from WinSW's own behavior:

**Machine:** Windows 11 25H2, build 26200.9168. (Flagging one oddity so it doesn't look like an error later: this build's own registry `ProductName` still reads "Windows 10 Pro" — a known cosmetic string bug, not an actual Windows 10 install. Noted inline in the doc too.)

**WinSW:** v2.12.0 (`WinSW-x64.exe`).

**Verified directly** on this machine, against a throwaway service id/port (not the real running instance): download, XML config, `install`, `start` — `Get-Service` reports `Running` / `StartType Automatic`, the bound port is reachable, `stop` + `uninstall` removes it cleanly with no residue.

**Not independently reproduced:** the `<onfailure>` restart-on-crash cycle itself. The wrapped process runs as `LocalSystem` by default, and my session (a member of the local group that can apparently install/start services here, but without an elevated/UAC-admin token) got `Access is denied` trying to force-kill it to simulate a crash. So that leg is documented as WinSW's own long-standing SCM recovery-action behavior (linked in the doc) rather than something I confirmed end-to-end myself. Said so explicitly in the doc, with a pointer back to this issue for anyone who validates it on a setup where they can.

No Rust code touched, so the four required gates (`cargo fmt`/`clippy`/`test`/`deny`) are unaffected — didn't see a need to add the `windows` CI label for a docs-only change, but say the word if you want it run anyway.

Closes #530.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Docs
+- Documented the Windows Scheduled-Task-plus-`Start-Process` autostart trap:
+  the wrapper script's Job Object is torn down (and the spawned
+  `ai-memory.exe` killed with it) the instant the script exits, while
+  `LastTaskResult` keeps reporting `0`. Added Scenario E to `docs/windows.md`
+  covering WinSW as the native-Windows service wrapper (the systemd analog
+  for Windows), tested against Windows 11 25H2 (build 26200.9168) with
+  WinSW v2.12.0 (#530).
+
 ## [1.38.0] - 2026-08-30
 
 ### Added
--- a/docs/windows.md
+++ b/docs/windows.md
@@ -25,6 +25,43 @@ Command](#native-hook-command-claude-code-on-windows). Other agents use native
 single command strings matching their hook schema. PowerShell/Git Bash script
 bundles are compatibility fallbacks and do not enforce capture-policy v1.
 
+## Known Trap: A Naive Autostart Task Dies Silently
+
+If you already wired a Scheduled Task to run `ai-memory serve` at logon and
+it doesn't seem to actually stay up, read this before anything else — it's
+the most common way to lose the server on native Windows, and it hides
+itself.
+
+The usual shape: a Scheduled Task with a "run at logon" trigger whose action
+is `powershell.exe -File start-server.ps1`, where the script calls
+`Start-Process` to launch `ai-memory.exe` detached and hidden. This looks
+correct — it starts the server, you watch it come up — right up until the
+next reboot.
+
+**Symptom:** `Get-ScheduledTaskInfo` reports `LastTaskResult = 0` every
+single time, the task always "succeeds," yet the server is down and
+Event Viewer's Application log shows no crash for `ai-memory.exe`.
+
+**Root cause:** `Start-Process` returns immediately, so the wrapper script
+has nothing left to do and exits right after spawning the child. Task
+Scheduler runs the task's action inside a Job Object, and the spawned
+`ai-memory.exe` child never broke away from that job. The instant the
+action process (`powershell.exe`) exits, Windows tears down the Job Object
+— killing every process still in it, including the child that was supposed
+to keep running. That's a clean `TerminateProcess` call, not a fault, so it
+leaves nothing in Application-Error to search for. You can confirm this
+after the fact by cross-referencing `Microsoft-Windows-TaskScheduler/Operational`
+event IDs 200 (action started) / 201 (action completed) against the
+server's own startup log line: if the task "completes" within the same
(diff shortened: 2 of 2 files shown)


gabrielscharb:
Fechando para refazer: os testes que fiz nao usaram uma sessao elevada, entao nao confirmei de fato o ciclo de crash-and-recover do WinSW (so documentei como comportamento conhecido dele). Vou reabrir com o teste completo depois de validar com o usuario.

PR #538, opened 2026-08-31
feat(cli): add `ai-memory rename-workstream` for checkout-local renames

victorcesc:
## What it adds

`ai-memory rename-workstream` — a checkout-local rename for managed workstreams, backed by a new `POST /workstream/rename` route.

```
$ ai-memory rename-workstream --from typo-nmae --to refactor-db
Renamed workstream 'typo-nmae' to 'refactor-db'.
    id: 01a04092-982e-7471-b7dc-0e97f886d6ff
```

Selects by current name or by the stable id `workstreams` prints. `--json`, plus `--workspace` / `--project`. The Docker wrapper intercepts it alongside `run` / `show` / `continue` / `workstreams`.

## Why

Follow-up to #499. Names are fixed at `run --new` time and had no correction path, so a typo outlived the work it labelled — the only escape was starting a new workstream and abandoning the ledger attached to the old one. Discovery made the names visible; this makes them fixable.

## Design notes

**Metadata only.** `workstream_events`, `managed_runs`, and `workstream_native_sessions` all key on `workstreams.id`, so the rename is a single-row update with nothing to cascade. `selected_at` and `updated_at` are deliberately left untouched — relabelling is not activity, so neither the listing order nor the workstream a bare `ai-memory run` resumes moves as a side effect of fixing a typo.

**Scope.** `Capability::NormalWrite`, where the sibling discovery read uses `NormalRead`, resolved through `lookup_existing_scope` — no-create, fails closed. Both selectors repeat the checkout predicate: `workstreams.id` is globally unique, so without it a caller holding an id from another checkout could retitle a workstream their request never named. An out-of-scope id reads as absent rather than renamable.

**Refusals.** The destination is validated exactly like a `--new` name. A name another workstream in the same checkout already holds is refused as a typed conflict instead of a bare `UNIQUE` failure. Renaming a workstream to the name it already has writes nothing and is not an error.

## Verification

- `cargo fmt --all -- --check`, `git diff --check`, `cargo deny check`
- `TAILWIND_SKIP=1 cargo test --workspace` — 2658 passed, 0 failed
- 8 new tests. The listing-order one uses three workstreams rather than two, because with two the renamed row has nothing to swap with and a bumped `updated_at` would pass unnoticed.
- `scripts/managed-workstream-acceptance.sh` gains a rename leg: rename, the three refusals, then rename back and assert the listing is byte-identical to before the round trip, so any touched timestamp or moved row surfaces as a diff.

Two caveats on my side, both pre-existing on `main` and unrelated to this change: the acceptance script's deterministic phase does not run on macOS (`script -qefc` is util-linux syntax), and `check-native-packaging.sh` needs `systemd-analyze`. So I exercised the new leg against a purpose-built fixture instead — plus two mutations, to confirm the refusal assertions fail when they should — and verified the wrapper routing with the same fake docker/native harness that script uses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Added `ai-memory rename-workstream`, a checkout-local rename for managed
+  workstreams selectable by current name or by the stable id `workstreams`
+  prints. Names were fixed at `run --new` time and had no correction path, so
+  a typo outlived the work it labelled. The rename is metadata only: the
+  ledger, managed runs, and linked native sessions all key on the workstream
+  id, and `selected_at` and `updated_at` are deliberately left untouched, so
+  neither the listing order nor the workstream a bare `ai-memory run` resumes
+  moves as a side effect of relabelling. The destination is validated exactly
+  like a `--new` name and refused with a named conflict when another
+  workstream in the same checkout already holds it; renaming a workstream to
+  the name it already has writes nothing and is not an error. Both selectors
+  repeat the checkout predicate, so an id belonging to another workspace,
+  project, or worktree reads as absent rather than renamable. The Docker
+  wrapper routes the command through its native host client, since repository
+  identity is a host resource.
+
 ## [1.38.0] - 2026-08-30
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -229,6 +229,9 @@ priors are at the [bottom](#influences-and-prior-art).
   # List the workstreams that can be selected from this checkout.
   ai-memory workstreams
 
+  # Fix a name you regret; the ledger and the current selection stay put.
+  ai-memory rename-workstream --from typo-nmae --to refactor-db
+
   # List open cross-agent handoffs, oldest first, with the id
   # `memory_handoff_cancel` needs to clear a stale one.
   ai-memory handoffs
--- a/bin/ai-memory
+++ b/bin/ai-memory
@@ -13,6 +13,8 @@
 #   ai-memory show ...      Use that client for host project/harness discovery.
 #   ai-memory continue ...  Use that client to resume the newest linked checkout.
 #   ai-memory workstreams   Use that client to inspect the host checkout identity.
+#   ai-memory rename-workstream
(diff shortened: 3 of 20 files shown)


akitaonrails:
Approved. I merged current `main` into your branch to clear two conflicts, both mechanical:

- `docs/ARCHITECTURE.md` — the subcommand block; `purge-session` and `compact` landed on main while this was open. Resolved to main's list plus `rename-workstream`, which keeps `architecture_lists_every_visible_cli_subcommand` satisfied.
- `CHANGELOG.md` — your `### Added` entry re-anchored against the entries that landed since. Both kept.

## Verified

The scope containment is the part that mattered, and it holds. Both selectors carry the full four-part predicate (`workspace_id`, `project_id`, `repo_fingerprint`, `worktree_fingerprint`), and the id form needs it precisely because `workstreams.id` is globally unique. I controlled that rather than reading it: replacing the scope predicate in the id lookup with a tautology fails `rename_cannot_reach_a_workstream_outside_the_resolved_scope`. The test has teeth.

The transaction shape is right too — selector resolution, conflict check and `UPDATE` share one transaction, so there is no window between "no other workstream holds this name" and the write. The final `UPDATE ... WHERE id = ?` is safe because the id was resolved inside scope in that same transaction.

Leaving `selected_at` and `updated_at` alone is the correct call, and using three workstreams in the ordering test rather than two is the kind of detail that decides whether a test proves anything — with two, a bumped timestamp has nothing to swap with.

Gate on the merged result: `fmt`, `clippy -D warnings`, **2704 tests, 0 failures**, both changelog guards.

## One flake, not yours

The first full-workspace run failed `ops::tests::ensure_project_with_id_warns_when_creating_homonym_across_workspaces`. It passes alone on your branch, passes alone on `main`, the whole store suite passes on both, and the workspace rerun is green at 2704. It captures tracing output, which is global state, so workspace-wide parallelism is the likely trigger. Pre-existing and unrelated to this change — filing it separately so it does not get attributed here.

## One nit, not blocking

`rename` stores `input.new_name.trim()`, while the create path stores the name untrimmed. So `--new " foo "` yields `" foo "` but `--to " foo "` yields `"foo"`, and the rename conflict check would not see those as colliding. Your behaviour is the better one; the untrimmed create is the pre-existing oddity. Not worth holding this for.

PR #547, opened 2026-08-31
fix(hooks): resolve a symlinked binary before probing for hooks/

### LINKED ISSUE #546: install-hooks: hooks/ discovery fails when the binary is invoked through a symlink
## Summary

`install-hooks` cannot find the bundled `hooks/<agent>` directory when `ai-memory` is invoked through a symlink — e.g. `~/.local/bin/ai-memory -> <repo>/target/release/ai-memory`, which is exactly the layout the quick start encourages (`~/.local/bin` on `PATH`).

Both candidate paths that could match are derived from `std::env::current_exe()` without canonicalization, and on macOS that returns the symlink path itself rather than its target.

## Environment

- ai-memory 1.38.0, source build (`cargo build --release --workspace`)
- macOS arm64 (Darwin 25.6.0)
- `~/.local/bin/ai-memory` → `<repo>/target/release/ai-memory`

## Repro

```bash
ln -sf <repo>/target/release/ai-memory ~/.local/bin/ai-memory
cd <repo>          # cwd is irrelevant, see below
ai-memory install-hooks --agent claude-code --project-strategy repo-root --apply
```

Actual:

```
Error: could not locate hooks directory. Tried:
["/Users/me/hooks/claude-code",
 "/Users/me/.local/bin/hooks/claude-code",
 "/usr/local/share/ai-memory/hooks/claude-code",
 "/usr/share/ai-memory/hooks/claude-code",
 "/Users/me/Library/Application Support/ai-memory/hooks/claude-code"]
```

Expected: `<repo>/hooks/claude-code` is found, as it is when the same binary is invoked as `./target/release/ai-memory`.

## Root cause

`resolve_hooks_dir` builds its candidates from `hook_source_candidates(sub, repo_root_guess(), exe_dir_guess(), dirs::data_local_dir())` (`crates/ai-memory-cli/src/commands/install_hooks.rs:4174-4186`). Both guesses read `std::env::current_exe()` as-is:

- `repo_root_guess()` (`:4219-4225`) assumes `<repo>/target/{debug,release}/<bin>` and walks three parents up. Through the symlink that is `~/.local/bin/ai-memory` → `/Users/me`, so candidate #1 becomes `/Users/me/hooks/claude-code`.
- `exe_dir_guess()` (`:4230-4234`) yields `~/.local/bin`, so the release-tarball candidate becomes `~/.local/bin/hooks/claude-code`.

Neither ever points at the checkout, so the command fails no matter where it is run from.

A secondary usability problem: the failure looks like a "wrong directory" error, but **cwd is never consulted**. `cd`-ing into the repo root — the obvious first thing to try, and what the macOS docs imply by running the command from the checkout — changes nothing.

## Suggested fix

Canonicalize the executable path before deriving either candidate:

```rust
fn current_exe_resolved() -> Option<PathBuf> {
    let exe = std::env::current_exe().ok()?;
    Some(std::fs::canonicalize(&exe).unwrap_or(exe))
}
```

Two optional improvements on top:

1. Probe the cwd's git repo root (`<git root>/hooks/<sub>`) as an additional candidate, so running from the checkout works regardless of how the binary was invoked.
2. Mention `--hooks-dir <parent>` in the error message — it is the escape hatch, and the current message does not hint that it exists.

## Related (smaller, same file)

`resolve_hooks_dir` passes `dirs::data_local_dir()` for the last candidate, ignoring the resolved `data_dir`. With `AI_MEMORY_DATA_DIR=~/.cache/ai-memory` the process logs `data_dir=/Users/me/.cache/ai-memory`, yet the probe looks in `~/Library/Application Support/ai-memory/hooks/...`.

The same mismatch shows up on success: with that env var set, `--apply` reports

```
✓ staged 18 hook script(s) → /Users/me/Library/Application Support/ai-memory/hooks/claude-code
```

i.e. scripts are staged into the platform-default data dir rather than the one in use. Harmless for a `posix-native` install (the rendered commands call the binary directly and never read those scripts), but it silently populates a directory the operator is not using. Happy to split this into its own issue if you prefer.

## Workaround

```bash
ai-memory install-hooks --agent claude-code --hooks-dir <repo>/hooks --apply
# or invoke the real binary
<repo>/target/release/ai-memory install-hooks --agent claude-code --apply
```

alvadorn:
Fixes #546.

## Problem

`install-hooks` builds its hook-bundle candidates from where the binary lives — `repo_root_guess()` and `exe_dir_guess()`, both reading `std::env::current_exe()` as-is. Through the layout the quick start encourages (`~/.local/bin` on `PATH`, pointing into a source build), `current_exe()` returns the symlink itself, so:

- the repo-root guess walks three parents up from `~/.local/bin/ai-memory` and lands on `$HOME`;
- the tarball guess stops at `~/.local/bin`.

Neither ever reaches the checkout, so every candidate misses and the command fails:

```
Error: could not locate hooks directory. Tried:
["/Users/me/hooks/claude-code", "/Users/me/.local/bin/hooks/claude-code", ...]
```

The cwd is never consulted, so `cd`-ing into the repo root — the obvious thing to try, and what the macOS guide implies — does not help.

## Fix

Canonicalise the executable path once (`current_exe_resolved`), keeping the original when it cannot be resolved (deleted binary, or a filesystem that refuses), and derive both guesses from it. The parent-walk itself moves into `repo_root_from_exe(&Path)` so it can be tested against an injected root instead of the real `current_exe()`.

The failure message now also names `--hooks-dir` as the explicit escape hatch, which it never mentioned.

## Tests

- `hooks_are_found_through_a_symlinked_binary` (unix): builds `<tmp>/repo/target/release/ai-memory` plus a `<tmp>/home/.local/bin/ai-memory` symlink and asserts the first candidate is the checkout's `hooks/claude-code`. Fails before the change.
- `resolve_exe_path_keeps_an_unresolvable_path`: covers the fallback branch.

Verified end-to-end on macOS as well — through a symlink, from an unrelated cwd:

```
$ /tmp/scratch/ai-memory install-hooks --agent claude-code
# Hook scripts: /Users/me/workspace/ai-memory/hooks/claude-code
```

Local gate: `cargo fmt --all -- --check`, `git diff --check`, `TAILWIND_SKIP=1 cargo test --workspace`, `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings` — all clean.

## Note for review — `windows` label

Labelled `windows` because this is path handling: `fs::canonicalize` returns extended-length (`\\?\C:\…`) paths on Windows, so the resolved exe path — and the hooks directory derived from it — can carry that prefix into rendered output and staged-script paths. The behaviour is unchanged for the common Windows case (no symlink to resolve, `canonicalize` still normalises), but the native Windows workflow should run before merge.

## Not included

The related inconsistency noted in #546 — `resolve_hooks_dir` passing `dirs::data_local_dir()` for the last candidate instead of the resolved `data_dir`, which also sends `stage_hook_scripts` to the platform default rather than the data dir in use — is left out to keep this change scoped. Happy to follow up in a separate PR.

### DIFF
--- a/crates/ai-memory-cli/src/commands/install_hooks.rs
+++ b/crates/ai-memory-cli/src/commands/install_hooks.rs
@@ -4183,7 +4183,10 @@ fn resolve_hooks_dir(explicit: Option<&Path>, agent: AgentChoice) -> Result<Path
             return Ok(path.clone());
         }
     }
-    anyhow::bail!("could not locate hooks directory. Tried: {:?}", candidates,);
+    anyhow::bail!(
+        "could not locate hooks directory. Tried: {candidates:?}. \
+         Pass --hooks-dir <directory containing {sub}/> to point at the bundle explicitly.",
+    );
 }
 
 fn hook_source_candidates(
@@ -4217,20 +4220,38 @@ fn hook_source_candidates(
 }
 
 fn repo_root_guess() -> Option<PathBuf> {
-    // When the binary lives under target/{debug,release}/<name>, the
-    // workspace root is two parents up.
-    std::env::current_exe()
-        .ok()
-        .and_then(|p| p.parent()?.parent()?.parent().map(Path::to_path_buf))
+    repo_root_from_exe(&current_exe_resolved()?)
+}
+
+/// When the binary lives under target/{debug,release}/<name>, the
+/// workspace root is two parents up.
+fn repo_root_from_exe(exe: &Path) -> Option<PathBuf> {
+    Some(exe.parent()?.parent()?.parent()?.to_path_buf())
 }
 
 /// Directory the running binary lives in. The release tarball ships the
 /// `hooks/` bundle right next to the binary, so a no-`--source`
 /// `install-hooks` finds it there (issue #107).
 fn exe_dir_guess() -> Option<PathBuf> {
-    std::env::current_exe()
-        .ok()
-        .and_then(|p| p.parent().map(Path::to_path_buf))
+    Some(current_exe_resolved()?.parent()?.to_path_buf())
+}
+
+/// Path of the running binary with symlinks resolved.
+///
+/// Both guesses above are relative to where the binary *lives*, not to how
+/// it was invoked. `~/.local/bin/ai-memory -> <repo>/target/release/ai-memory`
+/// is the layout the quick start suggests, and without this the repo-root
+/// guess walks up from the link's own directory to `$HOME` while the tarball
+/// guess stops at `~/.local/bin` — leaving the bundled `hooks/` unreachable
+/// from every candidate, whatever the cwd (issue #546).
+fn current_exe_resolved() -> Option<PathBuf> {
+    std::env::current_exe().ok().map(resolve_exe_path)
+}
+
+/// Canonicalise an executable path, keeping the original when the target
+/// cannot be resolved (deleted binary, or a filesystem that refuses).
+fn resolve_exe_path(exe: PathBuf) -> PathBuf {
+    fs::canonicalize(&exe).unwrap_or(exe)
 }
 
 // CLAUDE_CODE_EVENTS + build_claude_code_payload now live in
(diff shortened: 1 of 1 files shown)


akitaonrails:
Approving this, and closing my own #553 in its favour. Explaining that properly, because the overlap was my fault and I don't want the precedent to be "the maintainer's branch wins".

You reported #546, root-caused it to the exact lines, and had a fix up. I was mid-batch on other issues, picked yours up, and opened #553 before your PR appeared — I should have said on the issue that I was starting, and didn't. That's the process failure, not anything you did.

On the merits yours is the one to land regardless of order:

- **You verified on real macOS.** I can't — this bug does not reproduce on Linux, where `current_exe()` reads `/proc/self/exe` and is already resolved. My PR's evidence stopped at an injected-path unit test.
- **You raised the `\\?\` extended-length path question and I missed it entirely.** That is the one thing about this change that could bite a platform I wasn't testing.

I verified the claims rather than taking them on trust:

| claim | check |
|---|---|
| the block walks out of the checkout | at v1.38.0 the derivation lands on `$HOME` through the link |
| the new test fails before the change | removed the canonicalisation → `the checkout must be the first candidate, not the symlink's ancestor` ✅ |
| gate clean | `fmt`, `clippy -D warnings`, **2652 tests, 0 failures** on your head |

On the `\\?\` concern specifically, I traced where the resolved path can surface. `resolve_hooks_dir`'s output reaches only `# Hook scripts: {}` informational prints and the staging read — never a written agent config, since rendered hook commands take the binary path from `current_exe()` separately and the staged-script paths come from `stage_hook_scripts`. So the blast radius is a cosmetic prefix in an informational line, not a path written into someone's settings. Worth knowing rather than assuming, so thank you for flagging it.

Your `windows` label was right and I've applied it — the run needed maintainer approval first (fork PR), which is why nothing had started. It's running now. For what it's worth, the same canonicalisation passed the Windows job on my branch, so I expect green.

Two things from my branch I am deliberately *not* pushing onto yours:

1. A cwd-git-root candidate probed last, for the "`cd` into the checkout doesn't help" trap you documented. It's a safety net, not needed for #546, and it belongs in its own PR with its own argument — yours if you want it, otherwise I'll open one.
2. My error-message wording. Yours says the same thing more briefly.

Merging once the Windows job is green.

PR #550, opened 2026-08-31
feat(purge): `--compact` for purge-project and delete-workspace (#540)

### LINKED ISSUE #540: purge-project and delete-workspace overstate what they delete
Split out of #387 rather than bundled into #539, because it is a separate claim about two already-shipped commands.

## What's wrong

`purge-project`'s help says:

> Permanently delete a project and ALL its data

Measured on `main` — canary token in an observation body, then grep the raw SQLite file after the purge:

```
PURGE_PROJECT shadow_rows=4 text_on_disk=true
```

The text is still recoverable from `memory.sqlite`. `delete_workspace` takes the same path.

## What is *not* wrong

This is not a cascade or trigger bug, and it is not unique to these commands. It is the ordinary behaviour of any SQLite delete: rows go, bytes stay in free pages until the file is rewritten. Matched pairs:

| | FTS shadow rows | text in file |
|---|---|---|
| delete only | 4 | yes |
| delete + `VACUUM` | 4 | yes |
| delete + `rebuild` + `VACUUM` | 2 | no |

The `rebuild` is the operative step — an ordinary delete leaves the tokens inside the FTS5 index segments, and `VACUUM` alone does not clear them.

So the defect is the **claim**, not the deletion. "ALL its data" promises something no delete in this system does.

## Proposed fix

#539 added `ai_memory_store::Compaction` and a `--compact` flag for `purge-session`, plus a table in `docs/lifecycle-ops.md` stating exactly what each mode does and does not guarantee. Extend the same two things to these commands:

1. `--compact` on `ai-memory purge-project`, and a `compact` field on `POST /admin/delete-workspace`, reusing `Compaction::{Skip,Reclaim}`.
2. The same boundary row in the `docs/lifecycle-ops.md` matrix, and softened help text so the command does not promise byte-level removal it does not perform.

That makes the destructive-command family consistent instead of having one command that documents its limits and two that overstate them.

## Explicitly out of scope

Forensic unrecoverability. The wiki git history stores page content in its objects and commit messages, and backups taken before a purge still contain everything. `--compact` reclaims bytes from the live SQLite file; it is not erasure everywhere and must not be described as such.

samirhvbr:
Closes #540.

## What changed

- `ai-memory purge-project --compact` and `compact: true` on `POST /admin/delete-workspace` reclaim the freed bytes, reusing `Compaction::{Skip,Reclaim}` from #539.
- Both responses gained a `compacted` boolean, so a caller is told which of the two modes ran instead of assuming.
- **No default changed.** Absent, `false`, and the old request body all behave exactly as before: a logical delete. The new field is `#[serde(default)]` on both endpoints.
- The overstated claims are gone: `purge-project`'s help said "Permanently delete a project and ALL its data", and the admin module docs said "all its data" / "all of its projects". They now say what the commands remove.
- `docs/lifecycle-ops.md` carries the boundary table for the whole destructive family — `purge-session`, `purge-project`, `delete-workspace` — rather than for `purge-session` alone, with both matrix rows updated.

## Why

Your measurement on the issue: `PURGE_PROJECT shadow_rows=4 text_on_disk=true`. The rows are gone and the text is still in the file, because that is what any SQLite delete does. The defect is the claim, so this fixes the claim *and* the absence of a way to make it true.

## The part a literal reuse of #539 gets wrong

`purge_session` rebuilds `observations_fts` and `pages_fts`. Exhaustive for a session; not for a project or a workspace.

`workstreams.workspace_id` and `workstreams.project_id` are both `ON DELETE CASCADE` (`V31__managed_workstreams.sql:6-7`), `workstream_events.workstream_id` cascades out of those (`:69`), and `workstream_events_fts` is an external-content FTS5 index over `workstream_events.content` (`:92`). A managed workstream's event ledger is reachable from neither a session id nor a page, so a session purge never touches that index and a project or workspace purge always can.

Counting `*_fts_data` rows after each step, one purged project and one surviving project both holding a workstream with events:

| after | observations_fts | pages_fts | workstream_events_fts |
|---|---|---|---|
| delete only | 11 | 11 | 11 |
| + `VACUUM` | 11 | 11 | 11 |
| + the #539 batch | 3 | 3 | **11** |
| + `workstream_events_fts` rebuild | 3 | 3 | 3 |

Same shape at 4, 20 and 200 events per workstream. Rebuilding only the session's two would leave that index holding entries for rows that no longer exist, with the response still reporting `compacted: true`.

So `reclaim_after_commit` takes the index set explicitly and the two sets are named constants, rather than the correct set being a fact about today's schema that lives in one function's head.

## Enforced rather than asserted

Two tests carry the claim, because a comment saying "these are all of them" is exactly the thing that rots:

- `scope_wide_reclaim_covers_every_fts_index_in_the_schema` reads every `USING fts5` table out of `sqlite_master` and requires `SCOPE_WIDE_FTS` to equal it. An index added later fails here, routing whoever adds it into the decision instead of letting `--compact` silently skip it.
- `purge_project_reclaim_rebuilds_the_workstream_event_index` asserts *"an explicit rebuild afterwards changes nothing"*, which stays true however the reclaim is spelled, and shows the assertion has teeth by checking the `Skip` half — where the same explicit rebuild does change it.

I verified the second one fails for the right reason rather than trusting it: routing `purge_project` through `SESSION_SCOPED_FTS` produces

```
assertion `left == right` failed: Reclaim must already have rebuilt
workstream_events_fts — an explicit rebuild afterwards found more to drop,
so the purge left stale index entries behind for rows it deleted
```

then reverted.

## Scope decisions I made

- **The internal purges stay uncompacted.** `move-project`'s merge teardown and `merge-workspace`'s shell delete both call these functions, and both pass `Compaction::Skip`. Those are steps of a move, not an operator asking to reclaim bytes, and a `VACUUM` in the middle of a move is a surprise nobody requested. Say the word if you would rather they honour a flag.
- **`purge-project` keeps the paired boundary tests** — one asserting the default leaves the bytes, one asserting `--compact` removes them — mirroring how #539 paired them for `purge-session`, so the difference between what we promise and what we offer stays pinned from both sides.

## Drive-by

`purge_project`'s doc-comment was orphaned. #539 inserted `Compaction`, `PurgeSessionSummary` and `purge_session` between the block and its function, so it had been silently documenting the `Compaction` enum while `purge_project` had no docs at all. Moved back, since this change adds a paragraph to it. Nothing else in the move; only the "and all its data" phrase changed.

## One measurement of yours I could not reproduce

Raised on the issue and repeated here so it is not lost in the merge. Your table has `delete + VACUUM` still leaving the text in the file. In my harness `VACUUM` alone cleared the canary at every scale I tried — `delete only` recovers it, `+ VACUUM` does not — while the shadow rows sit unchanged. That matches your shadow-row column exactly and contradicts your text column.

It does not change this PR: the rebuild is required either way, for the stale index entries if nothing else, and the existing `purge_session_is_unsearchable_but_does_not_claim_byte_level_removal` only pins the `delete only` case, which we agree on. It does mean **the docs matrix row I wrote says "no" for bytes-in-file under `--compact` and "yes" under the default**, which is true in both our measurements — I deliberately did not write a row asserting anything about `delete + VACUUM` on its own. Happy to re-run against whatever shape you used if you want that row pinned too.

## Test plan

- [x] `cargo test --workspace` passes — 70 suites, 0 failures
- [x] `cargo deny check` — `advisories ok, bans ok, licenses ok, sources ok`
- [x] `bash scripts/check-changelog-frozen.sh upstream/main` — "no released section was modified"
- [x] Manual test: the negative verification above (routing `purge_project` through the session FTS set and confirming the regression test fails, then reverting), plus the `*_fts_data` measurement table, run against `f09382d`.

New tests: `purge_project_is_unsearchable_but_does_not_claim_byte_level_removal`, `purge_project_with_reclaim_removes_the_bytes_from_the_file`, `purge_project_reclaim_rebuilds_the_workstream_event_index`, `scope_wide_reclaim_covers_every_fts_index_in_the_schema`, `delete_workspace_reclaims_only_when_asked`, and one HTTP round-trip test per endpoint asserting `compacted` reports what actually ran across absent / `false` / `true`.

## Commit attribution

- [x] Verified: `Samir Hanna Verza <123646616+samirhvbr@users.noreply.github.com>`, the same identity as my merged commits.

## CHANGELOG (merge gate)

- [x] Added under `[Unreleased]` / `### Added`, alongside the #539 entry it extends.
- [x] No default changed — called out in "What changed" above.

## Notes for reviewers

The FTS set is the part worth your scrutiny. Everything else is mechanical plumbing of an existing enum through two more paths; that one is a judgement about what a scope-wide delete can dirty, and I would rather you check my reading of the `V31` cascades than take the test's word for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   wiki git history keeps page content in its objects and commit messages, and
   earlier backups still contain it. `docs/lifecycle-ops.md` states that
   boundary in a table so it is not inferred from the command name.
+- `ai-memory purge-project --compact` and `compact: true` on
+  `POST /admin/delete-workspace` reclaim the freed bytes, the way
+  `purge-session --compact` already did. Both commands described themselves as
+  deleting "ALL its data", which no delete in this system does: measured with a
+  canary token, the text was still recoverable from `memory.sqlite` after the
+  purge, because rows go and bytes stay in free pages until the file is
+  rewritten. The commands' help text and the admin endpoint docs now say what
+  they remove instead of promising byte-level removal, and
+  `docs/lifecycle-ops.md` carries the same boundary table for the whole
+  destructive family rather than for `purge-session` alone (#540).
+
+  Compaction is scope-aware rather than a copy of the session batch. A session
+  purge dirties `observations_fts` and `pages_fts`; a project or workspace
+  delete also cascades into `workstreams` and their `workstream_events`, whose
+  content is indexed by a third FTS table that no session purge can reach.
+  Rebuilding only the session's two would have left that index holding entries
+  for rows that no longer exist, with the response still reporting
+  `compacted: true`. A test asserts the set is exhaustive against the schema,
+  so an FTS index added later fails there rather than being silently skipped.
+  The internal purges that back `move-project`'s merge teardown and
+  `merge-workspace`'s shell delete stay uncompacted — those are steps of a
+  move, not an operator asking to reclaim bytes.
 
 ### Fixed
 - The hook drain no longer stalls the whole spool behind one undeliverable
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -145,9 +145,12 @@ pub enum Command {
     /// `is_latest=false` (they were a multi-project mash-up) so the
     /// next consolidation can regenerate them per-project. Idempotent.
     Reorg(ReorgArgs),
-    /// Permanently delete a project and ALL its data (pages, sessions,
-    /// observations, handoffs, embeddings, on-disk wiki files).
-    /// This is irreversible — requires `--confirm`.
+    /// Permanently delete a project: its rows (pages, sessions, observations,
+    /// handoffs, embeddings, managed workstreams) and its on-disk wiki files.
+    ///
+    /// Irreversible, and requires `--confirm`. A logical delete by default —
+    /// the rows become unreachable, but their bytes stay in free pages of the
+    /// database file until it is rewritten. `--compact` reclaims them.
     PurgeProject(PurgeProjectArgs),
     /// Permanently delete ONE session and everything derived from it.
     ///
@@ -609,6 +612,19 @@ pub struct PurgeProjectArgs {
     /// out — purging is destructive and irreversible.
     #[arg(long)]
     pub confirm: bool,
+    /// Also reclaim the freed bytes: rebuild the affected FTS indexes and
+    /// VACUUM the database after the delete commits.
+    ///
+    /// Without this the purge is a logical delete — the project is gone from
+    /// the API and from search, but its bytes stay in free pages of the
+    /// database file until it is next rewritten, as with any SQLite delete.
+    ///
+    /// This rewrites the WHOLE database and needs free disk space of about
+    /// its size, so it takes minutes on a large store. It also does NOT make
+    /// the content forensically unrecoverable: the wiki git history and any
+    /// backup taken before the purge still contain it.
+    #[arg(long)]
+    pub compact: bool,
 }
 
 /// Arguments for `purge-session`.
--- a/crates/ai-memory-cli/src/commands/purge_project.rs
+++ b/crates/ai-memory-cli/src/commands/purge_project.rs
@@ -15,6 +15,8 @@ struct PurgeProjectRequest {
     confirm: bool,
     /// Purge even when a managed workstream still holds a live run lease.
     force: bool,
+    /// Rebuild the FTS indexes and VACUUM after the delete commits.
+    compact: bool,
 }
 
 /// Run the `purge-project` subcommand.
@@ -49,6 +51,7 @@ pub async fn run(config: &Config, args: PurgeProjectArgs) -> Result<()> {
             project: project.clone(),
             confirm: true,
             force: args.force,
+            compact: args.compact,
         },
     )
     .await?;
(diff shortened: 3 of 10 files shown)


samirhvbr:
Superseded by #548, which landed while this was being written. Closing.

For what it is worth, we converged on the same design independently — same flag, same `compacted` field on both responses, same `Compaction::Skip` for the `move-project` and `merge-workspace` internal callers, and the same `workstream_events_fts` finding.

Your explanation of *why* that third index matters is better than mine, and I want to say so plainly rather than quietly drop it: I attributed the surviving tokens to segment residue after a delete. You identified that the cascade never fires the `AFTER DELETE` trigger at all, because `recursive_triggers` is off — so those are live index entries for rows that no longer exist, not residue. That is a sharper claim and it explains my own measurement better than my reading did.

Your note about the first control initially passing on the hyphenated needle is the useful half of this for anyone who reads it later. I checked my own equivalent test against that failure mode after reading it, and it happens to survive for a different reason — it asserts on `%_data` row counts rather than a needle in the file, so there is no tokenization to get wrong — but that was luck, not care, and I would have written the same vacuous assertion if I had reached for a canary string.

Two things from this branch are not in #548 and are not duplicates. I will send them separately rather than as a bundle here:

1. `purge_project` has no doc-comment. The block above it is separated from the function by `Compaction`, `PurgeSessionSummary` and `purge_session`, so rustdoc renders it on the `Compaction` enum — including the "Delete a project's rows inside one transaction" line #548 just rewrote. Verified with `cargo doc --document-private-items`: the text appears on `enum.Compaction.html` and `fn.purge_project.html` has none of it.

2. `reclaim_freed_pages` names its three indexes inside a SQL string literal, so a fourth added later is skipped silently and nothing fails. Your design decision — rebuild all of them, always, because a rebuild is idempotent and the `VACUUM` dominates — is the right one and I am not proposing to change it; it just wants a guard that fails when the list stops being all of them.

PR #553, opened 2026-08-31
fix(install-hooks): resolve the executable path before deriving hooks/ candidates (#546)

### LINKED ISSUE #546: install-hooks: hooks/ discovery fails when the binary is invoked through a symlink

akitaonrails:
Closes #546. Thanks @alvadorn — the report located the exact lines and the root cause was right; I verified all four claims against current main before changing anything.

## Why it broke so completely

`std::env::current_exe()` is not consistent across platforms. Linux reads `/proc/self/exe` and gets a fully resolved path; macOS returns the path **as invoked**, symlink and all.

The part that turns this from a wrong guess into total failure is that *every* candidate is derived by walking **parents** of that path. So the symlink didn't mislead one probe — it relocated the whole search:

| candidate | through `~/.local/bin/ai-memory` |
|---|---|
| repo root (3 parents up) | `$HOME/hooks/claude-code` |
| binary sibling | `~/.local/bin/hooks/claude-code` |

Neither can ever point at the checkout, which is why `cd`-ing into the repo didn't help either — and why the reported error lists five paths, none of them plausible.

Canonicalizing first makes macOS agree with Linux, where it is a no-op. An unresolvable path keeps its original value rather than dropping the guess: an unresolved candidate is still worth probing.

## The two suggested improvements — both taken

**The cwd checkout is now probed, but *last*.** You flagged the real usability problem: the message reads like a wrong-directory error, so `cd` into the checkout is the obvious first thing to try, and it did nothing. Last placement is deliberate — it can only rescue a lookup that was already going to fail, and never displaces an installed bundle for someone who merely happens to be standing in a repo. A test pins that ordering.

**The error now names `--hooks-dir`**, and says the argument is the *parent* holding `<agent>/` — the detail most likely to cost a second failed attempt.

## Tests

The symlink test's two halves are each other's control:

```rust
// the bug: three parents up from ~/.local/bin/ai-memory is $HOME
assert_eq!(repo_root_from_exe(&link).unwrap(), tmp.path().join("home"));
// the fix: canonicalize first and the derivation lands in the repo
assert_eq!(canonicalize(repo_root_from_exe(&resolve_exe_path(link))?)?, canonicalize(&repo)?);
```

Control verified: removing the canonicalization fails it ✅

`repo_root_from_exe` is split out so the derivation can be tested against a path that isn't the test binary's own — otherwise this is untestable on Linux, where `current_exe()` already resolves and the bug cannot reproduce.

`#[cfg(unix)]` on the symlink test: Windows symlinks need privileges, and Windows' `GetModuleFileNameW` returns the real path anyway. **Labelled `windows`** regardless, since this is path handling and the repo's policy is to run the Windows job on those.

## Split out, not ignored: #554

Your "Related" item is real and I reproduced the reasoning: `resolve_hooks_dir` and `stage_hook_scripts` both use `dirs::data_local_dir()` rather than the resolved data dir, so `AI_MEMORY_DATA_DIR` is ignored for the probe *and* for where `--apply` stages scripts. Taking you up on the offer to split it — it touches every `resolve_hooks_dir` call site and changes where files get written, which wants its own review rather than riding along with a path-resolution fix.

## Gate

`fmt` ✅ · `clippy -D warnings` ✅ · **2684 tests, 0 failures** ✅ · both changelog guards ✅

https://claude.ai/code/session_01MDbhmszrjG9s5MrPrTuNtm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -153,6 +153,26 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   three destructive commands in one place instead of documenting the limits of
   one and overstating the other two.
 
+- `install-hooks` now finds the bundled `hooks/` directory when `ai-memory` is
+  invoked through a symlink (#546). Reported by @alvadorn against 1.38.0 on
+  macOS arm64 with `~/.local/bin/ai-memory -> <repo>/target/release/ai-memory`
+  — the layout the quick start encourages.
+
+  `std::env::current_exe()` is not consistent across platforms: Linux reads
+  `/proc/self/exe` and gets a fully resolved path, macOS returns the path as
+  invoked. Every candidate is derived by walking *parents* of it, so an
+  unresolved symlink did not mislead one guess — it moved the entire search to
+  `$HOME`, and no `hooks/` directory was ever reachable. Resolving the path
+  first makes macOS agree with Linux, where it is a no-op.
+
+  Two smaller fixes alongside it. The cwd's git checkout is now probed as a
+  **last** candidate: the cwd was previously never consulted, which made the
+  failure actively misleading — the error reads like a wrong-directory problem,
+  so the first thing anyone tries is `cd` into the checkout, and that changed
+  nothing. It is deliberately last, so it can only rescue a lookup that was
+  going to fail and never displaces an installed bundle. The error message also
+  now names `--hooks-dir`, the escape hatch it previously left unmentioned.
+
 ### Docs
 
 - `docs/windows.md` gains **Scenario E**, a persistent-server story for native
--- a/crates/ai-memory-cli/src/commands/install_hooks.rs
+++ b/crates/ai-memory-cli/src/commands/install_hooks.rs
@@ -4177,13 +4177,18 @@ fn resolve_hooks_dir(explicit: Option<&Path>, agent: AgentChoice) -> Result<Path
         repo_root_guess(),
         exe_dir_guess(),
         dirs::data_local_dir(),
+        cwd_repo_root(),
     );
     for path in &candidates {
         if !path.as_os_str().is_empty() && path.is_dir() {
             return Ok(path.clone());
         }
     }
-    anyhow::bail!("could not locate hooks directory. Tried: {:?}", candidates,);
+    anyhow::bail!(
+        "could not locate hooks directory. Tried: {candidates:?}\n\n\
+         Pass --hooks-dir <dir> to point at the bundle explicitly, where <dir> \
+         is the PARENT holding `{sub}/` (e.g. `--hooks-dir <repo>/hooks`)."
+    );
 }
 
 fn hook_source_candidates(
@@ -4191,8 +4196,9 @@ fn hook_source_candidates(
     repo_root: Option<PathBuf>,
     exe_dir: Option<PathBuf>,
     data_local_dir: Option<PathBuf>,
+    cwd_repo_root: Option<PathBuf>,
 ) -> Vec<PathBuf> {
-    let mut candidates = Vec::with_capacity(5);
+    let mut candidates = Vec::with_capacity(6);
     // Cargo-run from the repo.
     if let Some(root) = repo_root {
         candidates.push(root.join("hooks").join(sub));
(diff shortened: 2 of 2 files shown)


akitaonrails:
Closing in favour of #547, which fixes #546 from the reporter who found and root-caused it. Their branch carries evidence mine could not: verification on real macOS hardware, where this bug actually reproduces — on Linux `current_exe()` reads `/proc/self/exe` and is already resolved, so my evidence stopped at an injected-path unit test. They also raised the `\\?\` extended-length path question on Windows, which I missed entirely.

I opened this before their PR appeared, having picked up the issue mid-batch without saying so on the ticket first. That was the process failure. The fix is theirs.

Not lost: the cwd-git-root fallback candidate from this branch — for the "`cd` into the checkout doesn't help" trap — is a genuine improvement but a safety net rather than part of #546, so it wants its own PR and its own argument.

PR #555, opened 2026-08-31
docs(store): attach purge_project's doc-comment to purge_project, and make `reclaim_freed_pages`'s index list checkable

samirhvbr:
Two follow-ups to #548. One file, no behaviour change, no CHANGELOG entry — doc-comment placement and a test are not user-facing.

## `purge_project` has no doc-comment

The block written for it sits above `Compaction`, `PurgeSessionSummary` and `purge_session`, with no item between it and the enum. Rustdoc attaches a doc block to the *next* item, so the whole thing lands on `Compaction` — including "Delete a project's rows inside one transaction" and an execution order describing a function three items further down.

Verified rather than argued, `cargo doc -p ai-memory-store --no-deps --document-private-items`:

| | `fn.purge_project.html` | `enum.Compaction.html` |
|---|---|---|
| before | no trace of it | "Delete a project's rows inside one transaction" |
| after | the text | gone |

This is worth saying carefully, because it is not a slip anyone should feel bad about: the block reads as attached when you scroll the file, and it has been detached since #539 inserted three items into the gap. #548 rewrote that text — the "This is a *logical* delete unless `compaction` is `Compaction::Reclaim`" paragraph is new and good — with the placement invisible from the diff. Only rustdoc disagrees, and nothing was asking rustdoc.

Moved verbatim. No wording changed, so the diff is a pure relocation.

## `reclaim_freed_pages` promises "every index" where nothing can check it

Your reasoning for rebuilding all three regardless of scope is right, and I am not proposing to change it — a rebuild is idempotent, the `VACUUM` dominates, and "rebuild only what this caller should have touched" is exactly the reasoning that leaves text searchable.

But "every FTS5 index" was spelled inside a SQL string literal, where nothing compares it to the schema. A fourth index added later gets skipped in silence, with the response still reporting `compacted: true` while its entries for deleted rows survive — the same shape of defect you caught with `workstream_events_fts`, which took a person noticing that the cascade never fires the FTS delete trigger. That catch should not have to happen twice.

So the list moves to `ALL_FTS_INDEXES` and `reclaim_rebuilds_every_fts_index_in_the_schema` compares it against `sqlite_master`. A new index fails the build instead of needing to be noticed.

Verified to have teeth by dropping `workstream_events_fts` from the list:

```
assertion `left == right` failed: ALL_FTS_INDEXES is out of date.
`reclaim_freed_pages` rebuilds only what is listed there, so an FTS index
missing from it survives a --compact with entries for deleted rows still in it.
```

then restoring it.

## Context

I had #540 built independently when #548 landed, and closed it as superseded — we converged on the same flag, the same `compacted` field, the same `Compaction::Skip` for the internal callers, and the same third-index finding. Your account of *why* that index matters is better than mine: I attributed the surviving tokens to segment residue, and you identified that the cascade never fires the `AFTER DELETE` trigger because `recursive_triggers` is off, which makes them live entries rather than residue. These two items are the only things from that branch not already in yours.

## Test plan

- [x] `cargo test --workspace` passes — 70 suites, 0 failures
- [x] `cargo deny check` — `advisories ok, bans ok, licenses ok, sources ok`
- [x] `check-changelog-sections.sh` and `check-changelog-frozen.sh` both pass
- [x] Manual test: the rustdoc before/after above, and the deliberately-broken-list control on the new test.

## Commit attribution

- [x] Verified: `Samir Hanna Verza <123646616+samirhvbr@users.noreply.github.com>`, the identity on my merged commits.

## CHANGELOG (merge gate)

- [x] Deliberately none. Behaviour is byte-identical: same three indexes, same batch, same order. This is a doc-comment relocation plus a test, which CONTRIBUTING exempts.
- [x] No default changed.

## Notes for reviewers

The relocation is the half worth a second look — I moved the block without editing it, so `git show --color-moved` should render it as a pure move. If you would rather the doc-placement fix and the test guard were separate PRs, say so and I will split them.

### DIFF
--- a/crates/ai-memory-store/src/ops.rs
+++ b/crates/ai-memory-store/src/ops.rs
@@ -2959,37 +2959,6 @@ pub fn insert_wiki_migration(
     Ok(())
 }
 
-/// Delete a project's rows inside one transaction.
-///
-/// This is a *logical* delete unless `compaction` is [`Compaction::Reclaim`].
-/// The rows are gone and nothing reaches them through the API or search, but
-/// their bytes remain in free pages of the database file until it is
-/// rewritten — as with any SQLite delete. Neither mode is forensic erasure:
-/// the wiki git history and any earlier backup still hold the content.
-///
-/// Execution order:
-/// 1. Count rows in each dependent table (pages/all versions, sessions,
-///    observations, handoffs, embeddings) before the delete so we can
-///    report how many rows were removed.
-/// 2. Collect all distinct page paths stored under the project — these are
-///    the on-disk files the caller must clean up after this function returns.
-/// 3. DELETE FROM projects WHERE id = ? — the ON DELETE CASCADE clauses in
-///    V01 + V02 propagate the delete to pages, sessions, observations,
-///    handoffs, and page_embeddings automatically.
-/// 4. Commit and return the [`PurgeSummary`].
-///
-/// The `workspace_project_label` string is passed in by the caller (the
-/// admin handler has the human-readable names; the writer only has IDs) and
-/// forwarded verbatim into [`PurgeSummary::label`] for logging.
-///
-/// `force` overrides the live-managed-run guard (step 0): `workstreams`
-/// cascades out of `projects`, so purging a scope whose lease is still live
-/// would delete the lease row out from under a running agent.
-///
-/// # Errors
-/// Returns [`StoreError::ManagedRunActive`] when a managed run's lease is
-/// still live and `force` is false, or [`StoreError`] if any SQL statement
-/// fails. The transaction is rolled back automatically on error.
 /// Whether a destructive scope operation should reclaim the freed bytes
 /// afterwards. Shared by [`purge_session`], [`purge_project`] and
 /// [`delete_workspace`] so the family makes one promise, not three.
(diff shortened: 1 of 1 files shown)


akitaonrails:
Verified and merging. Both halves hold up, and I checked the history rather than taking the account on trust:

| commit | the block attaches to |
|---|---|
| v1.38.0 | `pub fn purge_project` ✓ |
| `05138e29` (#539) | `pub enum Compaction` ✗ |
| `3f15ae82` (#548) | `pub enum Compaction` ✗ |

So it detached when #539 inserted `Compaction`, `PurgeSessionSummary` and `purge_session` into the gap, and #548 then rewrote the text without noticing. Both of those are mine, and your framing is more generous than I'd have been: the block reads as attached when you scroll the file, and nothing in either diff showed otherwise. Only rustdoc disagreed, and I never asked it. I confirmed the relocation is a pure move by extracting both blocks — byte-identical, 31 lines.

The second half is the one I'd have missed indefinitely. `ALL_FTS_INDEXES` is worth more than the tidiness: my `workstream_events_fts` catch in #540 depended on a person noticing that the cascade never fires the FTS delete trigger, and you're right that this should not have to happen twice. A promise spelled inside a SQL string literal cannot be compared to anything.

I ran your control and one of my own:

- dropping `workstream_events_fts` from the list → `ALL_FTS_INDEXES is out of date` ✅
- the same truncation also fails `reclaim_clears_workstream_event_text_that_left_via_cascade` ✅

The second matters to me: it shows the refactor left the list *load-bearing* rather than turning it into a decorative constant that the guard test checks against itself.

Gate on your head: `fmt`, `clippy -D warnings`, **2682 tests, 0 failures**. CHANGELOG correctly omitted — CONTRIBUTING exempts internal refactors and test-only churn, and the behaviour is byte-identical.

No need to split; the two belong together, since the second is the durable form of the lesson the first exposed.

On the converged branch: closing yours as superseded was a generous call and I'd rather not have that be the norm. If it happens again, say so on the issue early and I'll hold off — you had it built independently, and the only reason mine landed first is that I was already mid-batch. The distinction you drew between segment residue and live index entries is the part I'd have got wrong without the trigger detail, so that reasoning came from looking closely, not from luck.

PR #557, opened 2026-08-31
Add a config knob for how hard the LLM thinks

Cardosaum:
## What changed

You can now set reasoning / thinking effort in one place:

```
AI_MEMORY_LLM_REASONING_EFFORT=low
# or in config.toml: llm_reasoning_effort = "low"
```

Leave it unset and nothing changes. The model keeps its own default.

If you do set it, ai-memory sends the native field each provider actually understands (OpenAI `reasoning_effort`, OpenRouter `reasoning`, Grok `reasoning_effort`, Claude `output_config.effort`, Codex `reasoning.effort`). Gemini and Copilot ignore it.

A typo no longer waits until the first consolidation call to 400. Load fails closed. Values a given host or model cannot take are clamped instead of forwarded raw (`ultra` becomes `max` on OpenAI; Claude `none` does not try to disable thinking on Mythos Preview / Fable 5).

## Why

Reasoning models spend hidden tokens before they emit the JSON consolidation needs. Operators already wanted a cheap "think less" switch for that path, and a real "think more" switch when they are on Opus / GPT-5 / Grok. This is that switch, without a different config key per provider.

## Test plan

- [x] `cargo clippy -p ai-memory-llm -p ai-memory-cli --all-targets -- -D warnings`
- [x] `cargo test -p ai-memory-llm -p ai-memory-cli`
- [ ] Manual: `AI_MEMORY_LLM_REASONING_EFFORT=low` on a real provider and check the request; `HIGH` / `ludicrous` must fail at config load

## CHANGELOG (merge gate)

- [x] `[Unreleased]` entry added
- [x] Default is unchanged: omit the key, keep the model default

## Notes for reviewers

The interesting bits are the per-host clamps in `types.rs` / `anthropic.rs` / `openai.rs`. They were checked against current OpenAI, Anthropic, OpenRouter, and xAI docs.


akitaonrails:
The feature work here is good and I verified it rather than taking the description on trust. One dependency question blocks merge.

## Verified

**The provider field claims are accurate.** I checked `output_config.effort` against the Messages API docs rather than assuming: it is real, and the allowed values are `low | medium | high | xhigh | max`, which matches your `ReasoningEffort` mapping. `thinking` with `type: "adaptive"` also exists, so the `uses_adaptive_thinking` branch is grounded. Getting the family gating right instead of sending one field everywhere is the difference between this working and it 400-ing on half the models.

**Fail-closed is real, not aspirational.** `load_rejects_invalid_reasoning_effort` asserts the error names `llm_reasoning_effort` and says `unknown variant`, so a typo dies at config load rather than on the first consolidation call. That was the most valuable claim in the description and it is the one you tested hardest.

**`rstest` is correctly a dev-dependency** in both crates, so it does not ship.

I merged current `main` into your branch — it was conflicting on `CHANGELOG.md` and `Cargo.lock` — and regenerated the lock. Gate on the merged result: `fmt`, `clippy -D warnings`, **2742 tests, 0 failures**, both changelog guards.

## [BLOCKING] `cognomen` as a production dependency

`cognomen` is added to `ai-memory-llm`'s real dependencies — not dev — and `cognomen-macros` is a proc macro, so it executes at compile time in every build of this workspace, including release and CI.

From crates.io:

| | |
|---|---|
| first published | 2026-08-15 (about two weeks ago) |
| total downloads | 385 |
| repository | `github.com/Cardosaum/cognomen` |

That last row is the part that should have been in the description. You wrote the crate. That is not a problem in itself — plenty of good crates start as an author scratching their own itch — but a reviewer needs to know it, because it changes what "add this dependency" means: the project would be taking a permanent runtime and build-time dependency on a two-week-old crate with one maintainer, introduced by that maintainer, without that being stated.

What it buys is `as_str` plus serde labels on four enums (`Role`, `ReasoningEffort`, one in `anthropic.rs`, one in `health.rs`). This codebase already writes exactly that by hand — `pub const fn as_str(&self) -> &'static str` in `observation.rs` and twice in `page.rs` — with `#[serde(rename_all = "...")]` alongside. So the trade is a new compile-time-executing dependency against a match arm per variant, in a codebase that has already chosen the match arm three times.

**Please drop `cognomen` and hand-write the four enums** to match the existing convention. `rstest` can stay.

I deliberately did not do this myself: it is four enums including a `kebab` one, and quietly swapping out a contributor's dependency choice is not a maintainer edit, it is a redesign. If you would rather I did it, say so and I will. If you think the dependency earns its place — you have context on the crate I do not — make that argument and I will read it properly.

## Not blocking

Your manual-verification box is still unchecked (`AI_MEMORY_LLM_REASONING_EFFORT=low` against a real provider). Neither of us can run that in CI without provider keys, and everything checkable is covered by the tests, so I am not holding merge for it — but if you do get a chance to run it against one real provider, say which, and the note is worth more than the checkbox.

Cardosaum:
I wrote [cognomen](https://github.com/Cardosaum/cognomen) for my own projects, just to stop repeating the same `as_str` / serde-label boilerplate. I figured it might fit here too. I get why you’d rather not take that dependency, so I went ahead and dropped it and wrote the enums by hand.

If you ever want to try it later, I’m happy to do a follow-up that replaces the existing match arms. The source is small if you want to poke at it first.

Issue #554, opened 2026-08-31
install-hooks ignores the resolved data_dir: probes and stages under dirs::data_local_dir() instead


akitaonrails:
Split from #546 at the reporter's suggestion — @alvadorn identified it there and offered to file it separately.

## What's wrong

`install-hooks` uses the platform default `dirs::data_local_dir()` in two places instead of the data dir the process actually resolved, so `AI_MEMORY_DATA_DIR` (and `--data-dir`) are ignored by both.

**The probe** — `resolve_hooks_dir` passes `dirs::data_local_dir()` as its last candidate (`install_hooks.rs`). With `AI_MEMORY_DATA_DIR=~/.cache/ai-memory` the process logs `data_dir=/Users/me/.cache/ai-memory`, yet probes `~/Library/Application Support/ai-memory/hooks/...`.

**The staging destination** — the same mismatch, and this one writes:

```rust
fn stage_hook_scripts(source_dir: &Path, agent_label: &str) -> Result<PathBuf> {
    let data_dir = dirs::data_local_dir()
```

so `--apply` reports

```
✓ staged 18 hook script(s) → /Users/me/Library/Application Support/ai-memory/hooks/claude-code
```

— populating a directory the operator is not using, and silently, since the line looks like success.

## Impact

Mostly benign today for a `posix-native` install: the rendered hook commands invoke the binary directly and never read the staged scripts, so nothing is broken at runtime. It is wrong in three smaller ways that will matter later:

1. It writes into a directory the operator did not choose and may not expect to exist.
2. The data-local *probe* can never find a bundle that a previous run staged into a custom data dir, so the two halves disagree with each other.
3. Docker `setup-agent` extracts the bundle into the data dir; a custom data dir puts it somewhere `resolve_hooks_dir` will not look.

## Why it wasn't fixed in #546

`resolve_hooks_dir(explicit, agent)` takes no data dir and is called from a dozen sites; `stage_hook_scripts` likewise. Threading the resolved dir through changes **where files are written**, which deserves its own review rather than riding along with a path-resolution fix.

## Suggested shape

- Thread the resolved data dir into `resolve_hooks_dir` and `stage_hook_scripts` (both already have `_in` variants taking the directory explicitly, so the testable seam exists).
- Keep `dirs::data_local_dir()` as the fallback when no data dir was configured, so current behaviour is unchanged for anyone not setting one.
- A test that a configured data dir is both probed and staged into, with the platform default untouched.

Issue #52, opened 2026-05-29
install-hooks (claude-code) stages 0 scripts and points settings.json at an empty dir (source/dest path collision)


Otavio-Machado-Santos:
## Summary

`ai-memory install-hooks --agent claude-code --apply` reported `✓ staged 0 hook script(s)` and then rewrote `~/.claude/settings.json` to point every hook at a directory that contains **no** scripts. The result is a settings file whose hook commands reference files that don't exist, silently disabling all capture.

## Environment

- ai-memory **v0.7.0**, built from source (`cargo build --release`, rustc 1.95.0)
- macOS (Apple Silicon, arm64)
- Binary invoked from `~/.local/bin/ai-memory` (i.e. **not** from inside the repo checkout)
- A previous install had created `~/Library/Application Support/ai-memory/hooks/claude-code/` but the actual scripts lived elsewhere (`~/.ai-memory/hooks/`), so the data-local dir existed but was **empty**.

## Observed

```
$ ai-memory install-hooks --agent claude-code --apply
✓ staged 0 hook script(s) → /Users/.../Library/Application Support/ai-memory/hooks/claude-code
✓ updated ~/.claude/settings.json (backup written next to it)
```

`settings.json` was then rewritten with commands like:

```
AI_MEMORY_HOOK_URL=http://127.0.0.1:49374 '/Users/.../Library/Application Support/ai-memory/hooks/claude-code/session-start.sh'
```

…but that directory is empty (`ls` shows 0 files), so every hook is a dangling reference.

## Root cause

`resolve_hooks_dir` and `stage_hook_scripts_in` can resolve to the **same path**, so staging copies from an empty dir into itself.

- `stage_hook_scripts_in` (install_hooks.rs:1292) stages **into** `data_local_dir/ai-memory/hooks/<agent>`.
- `hook_source_candidates` (install_hooks.rs:1462) lists `data_local_dir/ai-memory/hooks/<agent>` as the **last source candidate**.

When the binary runs outside the repo (`repo_root_guess()` → none) and the system paths (`/usr/local/share`, `/usr/share`) don't exist (normal on macOS), `resolve_hooks_dir` falls through to the data-local candidate — which is identical to the staging destination. If that dir pre-exists but is empty, the copy loop reads 0 files (`copied = 0`), wipes nothing, and the install still proceeds to rewrite `settings.json` pointing at the now-confirmed-empty dir.

So the source-of-truth scripts are never found, yet the command exits 0 and reports success.

[...]

PR #561, opened 2026-09-01
fix(hooks): stop putting the server bearer on hook command lines (#552)

### LINKED ISSUE #552: install-hooks writes the auth token onto the hook command line, exposing it in the process table
Found while working #542. Pre-existing, not introduced by it.

## What happens

`install-hooks --apply --auth-token <token>` renders the credential into the hook's **command line**. Against a scratch config file:

```
"command": "…/ai-memory --data-dir … hook --event session-start --agent command-code --server-url http://127.0.0.1:49374 --auth-token SEKRIT-CANARY-123"
```

Reproduced identically for `claude-code`. The token appears 4 times in the command-code config and 9 times in the claude-code one — once per registered event.

## Why it matters

Two exposures, and the second is the less obvious one:

1. **At rest.** The token sits in cleartext in the agent's config file, whose permissions ai-memory does not control and which users routinely sync, back up, or commit.
2. **In the process table.** A command line is world-readable on Linux (`/proc/<pid>/cmdline`) and via `ps` on macOS for as long as the process lives. Every hook invocation — and hooks fire on every tool call — briefly publishes the bearer to every local user on the machine. On a shared or multi-tenant host that is a credential disclosure that leaves no trace.

The code already knows the better answer. `render_shared.rs` documents it:

> `auth_token`, when set, lands in each hook's `env` block as …

so an env-block path exists; the rendered output above simply is not using it for these agents.

## Suggested direction

- Render the token into the hook entry's `env` block for every agent whose hook schema supports one, rather than onto `command`.
- For agents with no env block in their schema, prefer referencing an environment variable the user exports (`AI_MEMORY_AUTH_TOKEN`) over inlining the secret, and say so in the printed output. The manual (non-`--apply`) path already tells users to do exactly this:

  > Auth: set AI_MEMORY_AUTH_TOKEN in each hook's environment to the value passed via `--auth-token` (omitted from this printout).

  The `--apply` path should not be less careful than the instructions it prints.
- A test per agent renderer asserting the token does not appear in any `command` string, mirroring the argv-leak test added for the drain in #542.

## Scope note

Deliberately not bundled into #542. That change was about the drain recovering stranded events and touched one spawn path; this touches every agent renderer and changes what `--apply` writes into user config files, which deserves its own review and its own line in the changelog.

akitaonrails:
Closes #552.

## The exposure

`install-hooks --apply` wrote the token into the agent's own config as `--auth-token <token>` for native hooks, and as an `AI_MEMORY_AUTH_TOKEN=` shell prefix for the script hooks. Both land on a command line, so the credential was readable through `/proc/<pid>/cmdline` by any local user for as long as each hook ran — and hooks fire on **every tool call**.

The shell hooks then handed it to `curl -H "Authorization: Bearer …"`, putting it on a **second** command line. Fixing only the agent config would have left that half in place.

## The fix

Two `0600` files inside the `0700` data dir:

| file | read by |
|---|---|
| `<data_dir>/auth-token` | the native `ai-memory hook` |
| `<data_dir>/auth-header` | the shell hooks, via `curl -H @<file>` |

The second file is the point rather than duplication — building the header inline would put the credential straight back onto curl's argv. I verified `-H @file` end to end against a listener: the header arrives, and the token appears in no argv.

Renderers already took `Option<&str>` and omit the credential when it is `None`, so **one** decision covers every agent instead of twenty edits.

## What this corrects in the issue itself

I filed #552 citing a doc comment claiming the token "lands in each hook's `env` block". It does not — `grep '"env"'` matches nothing; no renderer emits one. That comment was describing a path that does not exist, and I have not built on it here.

## Backward compatibility

An explicit `--auth-token`, or `AI_MEMORY_AUTH_TOKEN` in the environment, still takes precedence, so every config written before this keeps working unchanged. Re-running `install-hooks --apply` moves an install onto the stored form. Only `--apply` persists — printing a snippet writes no credential, since writing one as a side effect of a dry run would be a surprise.

`uninstall` now clears the pair: removing the hooks removes the only readers, so leaving the files would strand a live credential.

## Tests

| guarantee | control | result |
|---|---|---|
| bearer stays out of the agent config | put it back into the rendered config | ✅ fails |

I checked first that **no existing test caught this** — reverting the change left the whole `install_hooks` suite green — which is why the regression guard is at the `run()` level rather than on a renderer.

The file modes are asserted, not assumed: moving a secret from argv to disk is only an improvement if it is not world-readable. Also covered: a rotated token replaces the old one rather than leaving both, and absent/empty/whitespace files read as "no token" instead of sending `Authorization: Bearer ` with an empty value.

## Gate

`fmt` ✅ · `clippy -D warnings` ✅ · **2774 tests, 0 failures** ✅ · both changelog guards ✅

https://claude.ai/code/session_01MDbhmszrjG9s5MrPrTuNtm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -251,6 +251,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Reported by @alvadorn alongside #546 and split out at their suggestion.
 
+- The server bearer is no longer written onto hook command lines (#552).
+  `install-hooks --apply` stored it in the agent's config as
+  `--auth-token <token>` for native hooks and as an `AI_MEMORY_AUTH_TOKEN=`
+  shell prefix for the script hooks, so it sat in `/proc/<pid>/cmdline` —
+  readable by any local user — for the lifetime of every hook, and of every
+  `curl` those hooks ran, on every tool call.
+
+  It now lives in two `0600` files inside the `0700` data dir:
+  `auth-token` for the native `ai-memory hook` path, and `auth-header` for the
+  shell hooks, which pass it to `curl -H @<file>`. The second file is the point
+  rather than duplication — building the header inline would put the credential
+  straight back on a command line.
+
+  Backward compatible: an explicit `--auth-token`, or `AI_MEMORY_AUTH_TOKEN` in
+  the environment, still wins, so configs written before this keep working.
+  Re-run `install-hooks --apply` to move an existing install onto the stored
+  form. Only `--apply` persists; printing a snippet writes no credential.
+
 ### Docs
 
 - `docs/windows.md` gains **Scenario E**, a persistent-server story for native
--- a/crates/ai-memory-cli/src/commands/hook.rs
+++ b/crates/ai-memory-cli/src/commands/hook.rs
@@ -799,10 +799,12 @@ where
     if should_spawn_background_drainer(&args.event)
         && let Err(err) = after_background_drain_event_enqueue(
             &dd,
-            // The hook's own token: current by definition, since the agent
-            // config it came from was rendered by the last `install-hooks`.
-            // The drain uses it only to retry an entry the server has already
-            // rejected with 401 (#542).
+            // The hook's own token, resolved the same way it authenticates
+            // its own request: `--auth-token`, else the environment, else the
+            // copy `install-hooks --apply` persisted under the data dir
+            // (#552). Current by construction either way. The drain uses it
+            // only to retry an entry the server has already rejected with 401
+            // (#542).
             effective_token,
             spawn_background_drainer,
         )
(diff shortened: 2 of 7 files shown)


akitaonrails:
Pushed two post-audit corrections onto this branch. Both are cross-PR drift from today's batch rather than defects in any single change, and neither alters what this PR does.

**1. A stale comment.** The note beside the drain spawn still said the token came from "the agent config … rendered by the last `install-hooks`" — which is precisely what this PR removes. It now describes the real resolution order: flag, then environment, then the persisted file.

**2. A regression test that stopped testing its own case.** `stage_hook_scripts_bails_when_source_equals_empty_dest` is the #52 guard for **source == dest**. #554 moved the destination from `<data_dir>/ai-memory/hooks/<agent>` to `<data_dir>/hooks/<agent>`, but the test kept building the old shape — so source ≠ dest, and it was passing through the different-paths guard that the *next* test already covers. The same-path branch had no coverage at all, and nothing failed to say so.

Restored to source == dest and verified it discriminates: disabling the empty-source bail fails it.

That second one is the kind of thing a post-audit exists for — each PR was individually correct, and the coverage hole only appeared in the composition.

Gate: `fmt`, `clippy -D warnings`, **2776 tests, 0 failures**.

Issue #563, opened 2026-09-01
No single-instance guard: two servers can open the same data dir


akitaonrails:
Found while assessing multi-session/multi-user support (#562). Documented there as a warning; not enforced in code.

## What happens

Nothing prevents two `ai-memory serve` processes from opening the same `data/`. The only locks in the tree are `client-projects.lock` and the spool's `.drain.lock` — neither guards the store.

SQLite survives it: WAL supports multi-process access and `busy_timeout` is set. What does not survive is everything *around* it:

- **Two writer actors.** The single-writer invariant (`AGENTS.md` #2) is per-process, not per-database. Two processes mean two writers contending through `busy_timeout` rather than one serialized queue.
- **Two git handles on the wiki.** Concurrent commits contend on `index.lock`; the loser's consolidation or session-end commit fails.
- **Two active-project pointers.** Each process keeps its own, so the isolation added in #562 is per-process and the two disagree.

## Why it matters

The arrangement is one a user naturally reaches for: point two machines at a synced folder (Dropbox, NFS, an SMB share) instead of standing up a server. It looks like it works — reads succeed, most writes succeed — and degrades in ways that surface much later, as a missing commit or a wiki whose git state does not match the index.

`docs/deploy.md` now says to run one server, but a warning in prose is weaker than a guard.

## Why it was not fixed in #562

A lock file has real failure modes of its own, and a bad one is worse than the bug:

- a stale lock after `SIGKILL`, an OOM, or a power loss locks the operator out of their own data;
- containers restart with the same volume and a different PID, so PID-based staleness checks mislead;
- an NFS/SMB mount — precisely the case this is meant to catch — is where `flock` semantics are least reliable.

## Suggested direction

`fs2::FileExt::try_lock_exclusive` on a `data/.serve.lock` held for the process lifetime is the usual answer, and the crate is already a dependency. The design work is the recovery story, not the lock:

- refuse with a message that names the holding process and the override, rather than a bare "already running";
- an explicit `--force` escape hatch, since an operator who knows the previous process is dead must not be stranded;
- decide deliberately what happens on a filesystem where locking is unsupported — refusing to start there would be worse than the problem.

Worth pairing with a startup check that the data dir is not on a known network filesystem, since that is the underlying mistake.

PR #596, opened 2026-09-02
fix(windows): backup walk skips .serve.lock so 2.0 migration survives (#593)

### LINKED ISSUE #593: OKF v0.2 backup-gated migration always fails on Windows: backup walk tars .serve.lock, which the same serve process holds under a mandatory exclusive lock (os error 33)
## Summary

The backup-gated OKF v0.2 migration (`2026_09_01T18_00_okf_v02_conformance`) **always fails on Windows when upgrading from 1.x to 2.0.x with existing data**, with `O processo não pode acessar o arquivo porque outro processo bloqueou parte do arquivo. (os error 33)` (ERROR_LOCK_VIOLATION). The pre-migration backup walk tries to `tar` the new `.serve.lock` single-instance file (#563) **while the same `serve` process holds an exclusive mandatory byte-range lock on it**, so the very first file entry read aborts the backup → the migration aborts → the server refuses to start.

The store is left half-migrated (SQL refinery migrations v57/v58 already applied, wiki not conformed, server cannot boot on either side) — effectively offline until this is resolved.

## Environment

- Windows 11 x86_64, native binary install (no Docker/WSL)
- Upgrading `ai-memory` 1.25.0 → 2.0.0 → 2.0.1 (reproduces identically on both 2.0.x)
- Existing data dir (~64 MB: `wiki/` 459 files, `db/` SQLite 48 MB, `hook-spool/`)
- Fresh stores never hit it (the backup gate skips fresh installs), so only real upgrades are affected

## Evidence

Log (first 2.0 start):

```text
INFO ai_memory::commands::serve: single-instance serve lock held lock=\\?\C:\Users\...\ai-memory\.serve.lock
INFO refinery_core::traits: current version: 58 / no migrations to apply
INFO ai_memory_wiki::migrations::runner: running wiki migration migration="2026_09_01T18_00_okf_v02_conformance" ...
INFO ai_memory_wiki::migrations::m2026_09_okf_conformance: OKF migration: taking the pre-migration backup first db_rows=429 files=459
ERROR ai_memory_wiki::migrations::runner: wiki migration failed — server cannot start until this is resolved
  error=... (os error 33)
```

The archive that lands in the backup dir is **95 bytes** — gzip headers only, i.e. the failure happens on the first regular file the walk archives.

## Root cause

1. `serve` acquires the single-instance lock via `fs2::FileExt::try_lock_exclusive` on `<data_dir>/.serve.lock` — `crates/ai-memory-cli/src/commands/serve.rs` (`SERVE_LOCK_FILE`, `acquire_serve_lock`). The lock is held for the process lifetime, and the migration runs **after** acquisition, in the same process.
2. `create_pre_migration_backup` (`crates/ai-memory-wiki/src/backup.rs`) walks the whole data dir and `tar.append_path_with_name()` every entry that is a file, excluding only the archive itself and `pre-migration-backup.json.tmp`. **`.serve.lock` is not excluded.**
3. On Windows, `LockFileEx` locks are mandatory: a read from another handle overlapping the locked range fails with `ERROR_LOCK_VIOLATION`. `read_dir` returns `.serve.lock` first (the leading `.` sorts before all names on NTFS), so `append_path_with_name` → `File::open` (succeeds, fs2 opens share-mode) → read → error 33.

This exact Windows behavior is already acknowledged a few lines away in `serve.rs` — the holder-info sidecar exists because "on Windows the exclusive lock blocks other handles from reading the locked file itself" — but the backup walk doesn't apply the same insight.

Linux/macOS are immune because POSIX locks are advisory and don't fault reads on other handles, which is presumably why CI (fresh-store flows + Unix) never exercises this. The bundled `backup.rs` tests build data dirs that contain no `.serve.lock`, so even a Windows CI wouldn't catch the serve-startup upgrade path.

## Suggested fix

Exclude the serve lock file from the backup walk (same treatment the walk already gives the archive and the receipt tmp) — e.g. share the `SERVE_LOCK_FILE` constant with the wiki crate, or skip any file name starting with `.serve.lock`. Defensively, skipping `.serve.lock*` in the walk covers the holder-info sidecar spelling changes too.

## Workaround used (unblocks the upgrade without a code change)

Start `serve` with `--force` while some *other* process holds a byte-range lock on `.serve.lock` starting above the first 64 KB (so serve's whole-file `try_lock_exclusive` contends → the `--force` "continuing unguarded" path runs without holding the lock → the tar can then read the 0-byte `.serve.lock` freely). A one-liner decoy in PowerShell:

```powershell
$fs = [IO.File]::Open("$DataDir\.serve.lock", 'OpenOrCreate', 'ReadWrite', 'ReadWrite')
$fs.Lock(1048576, 1); Start-Sleep 240; $fs.Unlock(1048576, 1); $fs.Dispose()
```

After the migration completes, restart the server normally (without `--force`) to restore the single-instance guard. Worth adding to `docs/MIGRATION-2.0.md` for Windows users if the fix doesn't ship before they hit this.

Happy to test a pre-release build from Windows if that helps.

rafaelkenedy:
## What changed

**Before:** on Windows, upgrading a store with existing data from 1.x to 2.0.x
aborted the OKF v0.2 pre-migration backup with `os error 33`
(`ERROR_LOCK_VIOLATION`). The backup gate then failed, the migration aborted,
and the server refused to start on either version — the store was left
half-migrated (refinery v57/v58 applied, wiki not conformed). Fresh installs
were unaffected (the gate skips them).

**After:** the backup walk skips `.serve.lock` and its `.holder` sidecar before
any `File::open`, the same way it already skips the half-written archive and
the receipt tmp file. The migration completes; a store already stuck recovers
by restarting on the fixed binary (the failed wiki migration is never recorded,
so it simply retries). No default changes.

Fixes #593.

## Why

Bug fix. `create_pre_migration_backup` (`crates/ai-memory-wiki/src/backup.rs`)
tars every file under the data dir, excluding only the archive and the receipt
tmp. `.serve.lock` — the single-instance guard (#563) that the *same* `serve`
process holds under an exclusive lock — was not excluded. On Windows
`LockFileEx` locks are mandatory: opening the locked file on another handle to
read it into the tar faults with `ERROR_LOCK_VIOLATION`, and `read_dir` yields
`.serve.lock` first (leading `.` sorts ahead on NTFS), so the walk died on its
first entry. Unix is immune (advisory locks don't fault reads), which is why
the Unix-only CI matrix never caught it.

The lock file name moves to `ai-memory-core` as `SERVE_LOCK_FILE`, shared by
`serve` and the wiki crate; `holder_info_path` derives the sidecar name from
it.

## Test plan

- [x] `cargo test --workspace` passes — on `x86_64-pc-windows-msvc`, Rust 1.95,
      including the `#[cfg(windows)]` test
- [x] `cargo deny check` passes (`advisories ok, bans ok, licenses ok, sources
      ok`) — and `Cargo.lock` is unchanged by this PR anyway (the fix uses
      `std`'s file locking, no new crate)
- [x] Manual test: negative control — reverted the walk guard and ran the two
      new tests. `a_held_serve_lock_does_not_fault_the_backup_on_windows` panics
      with `Os { code: 33, kind: Uncategorized }` (`ERROR_LOCK_VIOLATION`);
      `the_backup_walk_skips_the_serve_lock_and_its_sidecar` fails with
      `entries` = 4 instead of 2. Both green with the fix.

New regression tests in `backup.rs`:

- `the_backup_walk_skips_the_serve_lock_and_its_sidecar` (all platforms)
- `a_held_serve_lock_does_not_fault_the_backup_on_windows` (`#[cfg(windows)]`) —
  holds an exclusive lock on `.serve.lock` exactly as `acquire_serve_lock` does

## Commit attribution

      (`rafaelkenedy <k.rafaelalves@gmail.com>`, linked to the GitHub account.)

## Release impact

- [ ] **Minor**
- [ ] **Major (breaking)**

## CHANGELOG (merge gate)

- [x] Added a `CHANGELOG.md` `[Unreleased]` → `### Fixed` entry.
- [x] No changed default.

## Notes for reviewers

- Please add the **`windows`** label so `windows.yml` runs before merge — this
  touches file locking and path handling. (I can't set labels from a fork.)
- `SERVE_LOCK_FILE` lands in `ai-memory-core`, whose crate doc says "free of
  platform concerns" — it's a plain `&str` filename; the doc-comment notes the
  Windows hazard only to tell callers why they must skip it. Happy to move it
  to `ai-memory-store` or inline the literal in `backup.rs` if you'd rather.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- The OKF v0.2 pre-migration backup no longer aborts on Windows when
+  upgrading a store with existing data from 1.x to 2.0.x (#593). The
+  backup walks the whole data dir and archived every file, including the
+  `.serve.lock` single-instance lock (#563) — which the same `serve`
+  process holds under a *mandatory* exclusive lock on Windows, so the
+  first read faulted with `ERROR_LOCK_VIOLATION` (os error 33), the
+  backup gate failed, and the server refused to start on either version.
+  The walk now skips `.serve.lock` and its `.holder` sidecar (neither
+  carries durable state); the lock file name is shared from
+  `ai-memory-core`. Unix was immune (advisory locks). Once on the fixed
+  binary a half-migrated store recovers by restarting — the failed wiki
+  migration was never recorded and simply retries.
 - `as_of` time-travel queries and the entity retrieval stream now work
   on real stores. Both read the entity index, which was populated only
   from an LLM consolidator's `entities:` frontmatter — absent on the
--- a/crates/ai-memory-cli/src/commands/serve.rs
+++ b/crates/ai-memory-cli/src/commands/serve.rs
@@ -11,7 +11,7 @@ use ai_memory_consolidate::{
     ScheduledAutoImproveSettings, run_auto_improve_scheduler_tick, run_embedding_backfill,
     run_lint, run_sweep_with_options,
 };
-use ai_memory_core::{ActiveProject, ProjectId, Sanitizer, WorkspaceId};
+use ai_memory_core::{ActiveProject, ProjectId, SERVE_LOCK_FILE, Sanitizer, WorkspaceId};
 use ai_memory_hooks::{
     DEFAULT_HOOK_INGEST_MAX_IN_FLIGHT, HookState, ProjectCacheStore, WorkstreamState, hook_router,
     workstream_router,
@@ -67,9 +67,6 @@ const SESSION_CONSOLIDATION_POLL_INTERVAL: Duration = Duration::from_secs(15);
 /// requests are expected to finish well inside this lease.
 const SESSION_CONSOLIDATION_LEASE: Duration = Duration::from_secs(10 * 60);
 
-/// Lock file guarding a data dir against a second `ai-memory serve` (#563).
-const SERVE_LOCK_FILE: &str = ".serve.lock";
-
 /// The single-instance guard for `ai-memory serve`: an exclusive `flock` on
 /// `<data-dir>/.serve.lock` held for the process lifetime. The OS releases it
 /// when the process exits, so a crashed server never locks the operator out of
@@ -83,7 +80,7 @@ struct ServeLock {
 /// lock file because Windows' exclusive lock blocks reads of the locked
 /// file from other handles; informational only, rewritten by each holder.
 fn holder_info_path(data_dir: &Path) -> std::path::PathBuf {
-    data_dir.join(".serve.lock.holder")
+    data_dir.join(format!("{SERVE_LOCK_FILE}.holder"))
 }
 
 /// Take the single-instance serve lock for `data_dir`.
--- a/crates/ai-memory-core/src/lib.rs
+++ b/crates/ai-memory-core/src/lib.rs
@@ -40,6 +40,14 @@ pub const DEFAULT_PROJECT_NAME: &str = "scratch";
 /// event capture to it.
 pub const GLOBAL_SCOPE_PROJECT: &str = "_global";
 
+/// Name of the single-instance lock file `ai-memory serve` holds for its
+/// process lifetime (#563), directly under the data dir. It carries no
+/// durable state. On Windows the process holds a *mandatory* exclusive lock
+/// on it, so any code that walks the data dir (pre-migration backups,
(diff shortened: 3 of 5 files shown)


akitaonrails:
Resolved by #593's fix, which merged via PR #597 (Gaalbu). Your diagnosis here was independently correct and identical in root cause — thank you for the thorough write-up; both PRs nailed it. Closing as a duplicate of the merged fix; the `.serve.lock` (and `.holder` sidecar) are now skipped by the backup walk, so the Windows 2.0 upgrade no longer aborts with os error 33.

PR #606, opened 2026-09-03
feat(llm): send operator headers and identify ai-memory on every request

lucasliet:
## What changed

**`AI_MEMORY_LLM_HEADERS`** (`llm_headers` in `config.toml`) — a new config
key that attaches operator-supplied HTTP headers to every LLM chat request,
on every provider. Comma-separated `Name=Value` / `Name: Value` entries.

- Parsed and validated once at the configuration boundary into a typed
  `ExtraHeaders`, so a malformed entry fails at **startup** rather than on the
  first consolidation pass, and providers consume typed header material
  instead of operator strings — the rule provider auth already follows.
- Headers ai-memory sets itself (`authorization`, `content-type`, `x-api-key`,
  `x-goog-api-key`, `anthropic-version`, `anthropic-beta`, `openai-beta`,
  `host`, `content-length`) are **refused**, not duplicated: `reqwest`'s
  `RequestBuilder::header` appends, so a second `authorization` would break
  the request instead of overriding it.
- Values are marked sensitive on the wire and never logged — the `Debug` impl
  prints header **names** only, since this is a plausible place for an
  operator to put a token.

**Layers over 2.0.2's client headers rather than competing with them.** #610
gave the `opencode` provider an agent string plus a `LlmOperationId` under
`x-opencode-session`. Those stay the defaults, semantics untouched; an
`AI_MEMORY_LLM_HEADERS` entry for either name now takes precedence.
`OpenAiProvider::post` builds one header map per request — operator entries,
then `set_default` for the agent string and the operation-id header — and
applies it once with replace semantics. Applying the two sources separately
would append rather than replace, sending two `user-agent` values whenever an
operator configures one; a duplicate is worse than either value alone.

**⚠️ Changed default — `User-Agent`.** Every LLM chat request now sends
`User-Agent: ai-memory/<version>`. Previously only `opencode` sent one (via
#610); every other provider sent **none at all**, because `reqwest` adds no
user agent unless configured. Layered under `AI_MEMORY_LLM_HEADERS`, so an
operator entry still wins. **Copilot is excluded** and unchanged: it keeps
`GitHubCopilotChat/<version>`, the editor-plugin agent GitHub's Copilot API
expects.

**Fix — `AI_MEMORY_LLM_BASE_URL` now works with `provider=opencode`.** The
provider hardcoded the Go endpoint and the factory dropped the configured
base URL without a word, so Zen's catalogue at `https://opencode.ai/zen/v1`
was unreachable through the provider built for OpenCode; the only route was
`openai-compat`, which gives up the session header. Go remains the default,
so existing setups do not move. An override keeps the session header and the
user agent, since both endpoints correlate requests the same way.

Zen and Go are separate products, not two spellings of one, and the code said
otherwise: `OPENCODE_ZEN_BASE_URL` has always held Go's URL. It is now
deprecated in favour of `OPENCODE_GO_BASE_URL`, keeping both its value and
its export, so code compiled against v2.0 is unaffected. Model ids are per
catalogue, so an override needs an explicit `AI_MEMORY_LLM_MODEL`. The
`opencode-zen` alias selects Go like every other spelling — the endpoint
comes from the base URL, not the alias — and the docs now say so.

Docs updated: `docs/ARCHITECTURE.md` config reference, `docs/install.md`,
`docs/support-matrix.md`, `README.md`, `docker/.env.production.example`,
`docker/docker-compose.yml`, `packaging/env/ai-memory.env`.

## Why

OpenCode Zen/Go emailed operators running ai-memory against their endpoint:

> Some of your requests to OpenCode Go are missing an `x-opencode-session`
> header. If we don't have this we cannot properly optimize our service.
> Starting 09/06 requests missing this header may error.
>
> Here are your useragents that are missing this header:
>
> **Unknown client** — Your requests carry no user agent, so we can't tell
> what sends them — identify the tool and add the header there.

#610 solved that for OpenCode Go under the 09/06 deadline. This generalises
it: the next gateway with its own required header should not need another
provider-specific code path, and no provider should stay anonymous. An
unattributable request is the one a rate limiter throttles first.

### Is third-party use of OpenCode Go permitted? Yes — with conditions we meet

Source: [`packages/web/src/content/docs/go.mdx`](https://github.com/sst/opencode/blob/dev/packages/web/src/content/docs/go.mdx),
section **"Where can I use it?"**:

> OpenCode Go is designed to be used with [OpenCode](https://opencode.ai) and
> other popular coding agents that produce a similar types of requests.
>
> Traffic is monitored for abusive traffic that degrades the experience for
> other users.
>
> To ensure your account does not get flagged, make sure the tool you're using
>
> 1. does not generate abusive traffic
> 2. properly identifies itself (no broad user agents)

The same document publishes the `https://opencode.ai/zen/go/v1` endpoints
under **"Endpoints"** for direct API use.

That is why the default agent string names **ai-memory** and not OpenCode's
own (`opencode/<version>`, at
[`packages/opencode/src/session/llm/request.ts:18`](https://github.com/sst/opencode/blob/dev/packages/opencode/src/session/llm/request.ts)):
identifying the caller is the stated requirement, and impersonating another
client would defeat it.

Server-side
([`packages/console/app/src/routes/zen/util/handler.ts:103-113`](https://github.com/sst/opencode/blob/dev/packages/console/app/src/routes/zen/util/handler.ts))
the `x-opencode-*` headers feed `logger.metric({ session, request, client,
user_agent, ... })` and are stripped before the request is forwarded
upstream — correlation metadata, not authentication, so the value is a
free-form opaque id and an operator override is safe.

## Test plan

- [ ] `cargo test --workspace` — `ai-memory-llm` 205/205 green,
      `ai-memory-cli` 813 green. Two pre-existing failures in
      `logging::tests` (`readonly_log_dir_falls_back_to_temp_dir`,
      `fully_readonly_environment_degrades_to_stderr_only`) reproduce on a
      clean checkout of this base: they `chmod 0o555` a temp dir and expect
      the write to fail, which never happens as uid 0. Four `ai-memory-mcp`
      integration binaries could not be linked in my sandbox (disk quota),
      unrelated to this diff — please confirm on CI.
- [x] Wire-level tests (`extra_headers_on_the_wire.rs`) drive `build_provider`
      against wiremock and assert on the **received** request: an operator
      header arrives; the default user agent arrives as a *single* value; an
      operator user agent *replaces* rather than appends; provider auth and
      `content-type` survive alongside; a base-URL override actually
      redirects the `opencode` provider while keeping both identifying
      headers; and an operator `x-opencode-session` beats the per-operation
      default, exactly once. Asserting a value sits on the struct is not the
      same claim as asserting the right single value is on the wire.
- [x] #610's own wire tests for the defaults, including retry stability,
      still pass unchanged.

## Commit attribution

## Release impact

- [x] **Minor** — additive: new config key `AI_MEMORY_LLM_HEADERS`
- [ ] **Major (breaking)**

Targeting 2.1.0. The changed `User-Agent` default is additive in effect — a
header is added where none was sent — and overridable, so no breaking bump.
`OPENCODE_ZEN_BASE_URL` keeps its value and export under `#[deprecated]`.

## CHANGELOG (merge gate)

- [x] `CHANGELOG.md` `[Unreleased]` entries added under `### Added`,
      `### Changed` and `### Fixed`.
- [x] The changed `User-Agent` default is called out explicitly in "What
      changed" above.

## Notes for reviewers

**Caveat on the Copilot exception — flagged, not resolved.** Copilot is
excluded from the default user agent because its client and per-request
headers both carry `GitHubCopilotChat/<version>`, the agent GitHub's Copilot
API expects; `ExtraHeaders` replaces per name, so a default there would have
silently broken that provider. That means the Copilot flow keeps identifying
ai-memory **as another agent** — precisely what this PR fixes everywhere
else. It predates this change and resolving it is outside this PR's scope:
the maintainer should decide (leave as is, append an ai-memory token, or send
a distinguishing header alongside).

Other decisions worth a look:

- **Where the layering lives.** `set_default` in `post()`, not in each
  provider's constructor, because the operation-id *value* is per request
  while `ExtraHeaders` is built once. The common path (no client headers)
  still applies the operator map directly with no clone.
- **Per-request `.headers()` rather than client `default_headers`.** Keeps
  the builders infallible and matching the existing `with_timeout_secs` /
  `with_reasoning_effort` shape.
- **Comma-separated env parsing** follows the existing `allowed_hosts` /
  `cors_allow_origins` pattern, which means a header *value* cannot contain a
  comma via the env var. `llm_headers = [...]` in `config.toml` covers that;
  documented rather than worked around.
- **Dropped unit tests.** The ones that read the stored header map to assert
  the session default pinned the pre-#610 mechanism; the header is applied
  per request now. Coverage moved to the wire, not deleted.
- The `CHANGELOG` entries carry no `(#NNN)` reference yet — happy to add the
  PR number once assigned.


akitaonrails:
Audited — clean, well-structured **feature** (operator `AI_MEMORY_LLM_HEADERS` + ai-memory User-Agent identification), fail-closed on reserved headers, thorough wire-level tests. No blocking findings. Two notes for merge: it changes two overridable defaults (a default `User-Agent: ai-memory/<version>` layered under operator headers, and a per-process `x-opencode-session` for opencode) — both additive and disclosed, but a default change warrants a conscious sign-off. Routing to **2.1.0** (features → minor, per our semver discipline), separate from the current bug-fix patch batch. Will confirm the two default changes and CI before merging with 2.1.0.

lucasliet:
> Audited — clean, well-structured **feature** (operator `AI_MEMORY_LLM_HEADERS` + ai-memory User-Agent identification), fail-closed on reserved headers, thorough wire-level tests. No blocking findings. Two notes for merge: it changes two overridable defaults (a default `User-Agent: ai-memory/<version>` layered under operator headers, and a per-process `x-opencode-session` for opencode) — both additive and disclosed, but a default change warrants a conscious sign-off. Routing to **2.1.0** (features → minor, per our semver discipline), separate from the current bug-fix patch batch. Will confirm the two default changes and CI before merging with 2.1.0.

Pushed ccdea09 after your audit — it doesn't touch either of the two defaults you flagged. It fixes a separate pre-existing bug in the same provider: AI_MEMORY_LLM_BASE_URL was silently dropped for provider=opencode, so Zen's catalogue was unreachable (the constant was named OPENCODE_ZEN_BASE_URL but held Go's URL). Go stays the default; OPENCODE_ZEN_BASE_URL keeps its value and export, deprecated in favour of OPENCODE_GO_BASE_URL. Marked since = "2.1.0" per your routing.

akitaonrails:
Heads-up: PR #610 (@lucazz) landed the focused OpenCode Go `x-opencode-session` + `User-Agent` fix in the **2.0.2** patch, because OpenCode Go begins rejecting header-less requests on 2026-09-06 (couldn't wait for 2.1.0). Your broader operator-headers feature still stands for 2.1.0, but it now overlaps #610 on `openai.rs`/`types.rs`/`opencode.rs` — it'll need a rebase onto current main and to reconcile the header-injection mechanism with the one #610 introduced (`client_headers` opt-in + `LlmOperationId`). Happy to help reconcile when you rebase.

lucasliet:
Thanks @akitaonrails  no issue with #610 going first; the 09/06 deadline had to win, and a focused patch was the right call for 2.0.2.

Rebased onto current main and reconciled. The two turned out to be layers rather than competing mechanisms:

    2.0.2's client headers stay the default, semantics untouched. Your per-operation LlmOperationId is better than what this branch carried (a per-process id): it stays stable across retries and the strict/tolerant fallback, which is what OpenCode's metrics actually key on. I dropped mine and its helpers.
    AI_MEMORY_LLM_HEADERS layers over those defaults. An operator entry for user-agent or x-opencode-session now wins; absent one, nothing changes.
    The call site takes the header name and agent string from OPENCODE_SESSION_HEADER and DEFAULT_USER_AGENT instead of repeating the literals.

The one place worth your eyes: OpenAiProvider::post now builds a single header map per request, operator entries, then set_default for the agent string and the operation-id header, and applies it once. Applying the two sources separately would append rather than replace, sending two user-agent values whenever an operator configures one; a duplicate is worse than either value alone. There's a wire test pinning exactly that (an_operator_session_header_beats_the_per_operation_default), both headers arrive with a single value.

Dropped tests, deliberately: the unit tests that read the stored header map to assert the session default pinned the pre-#610 mechanism, since the header is applied per request now. Your wire tests already cover the defaults including retry stability; I added one for what this branch adds. Coverage moved to the wire, not deleted.

Still uniquely in this PR, unaffected by #610: the operator config itself, the default User-Agent on every provider (yours covers opencode; the rest were still anonymous), and a separate pre-existing bug, AI_MEMORY_LLM_BASE_URL was silently dropped for provider=opencode, so Zen's catalogue was unreachable. OPENCODE_ZEN_BASE_URL has always held Go's URL; it's now deprecated in favour of OPENCODE_GO_BASE_URL, keeping its value and export.

The two default changes you wanted signed off are unchanged from your audit: the global User-Agent, and now only the overridability of the opencode session header rather than the header itself. Happy to split the base-URL fix into its own PR if you'd rather keep 2.1.0 to the headers feature.

akitaonrails:
Reconciled onto `release/2.1` as merge `2ef03ca1` (your commits preserved). It was CONFLICTING only because it predated 2.0.2's #608/#610 opencode-header work; merging against that base produced a single CHANGELOG conflict — the header code auto-merged, and I verified the result rather than trusting the auto-merge:

- **Security boundary re-audited on the merged tree:** the 9-entry reserved-header list, CRLF-rejecting `HeaderName`/`HeaderValue::try_from` (no header injection), case-insensitive fail-closed reserved check, sensitive-marked values, and names-only `Debug` all hold.
- **Gate on merged tree:** `cargo fmt --check` clean · `clippy --workspace --all-targets -D warnings` clean · full workspace tests **0 failed**. Your on-the-wire tests and #610's opencode tests pass together, which is exactly the interaction that mattered.

Nicely done — the `ExtraHeaders` typed boundary, operator-wins layering, and the Zen/Go base-URL untangling are all clean. Queued for **2.1.0**; the branch holds until release. Closing as resolved-and-ready-to-release rather than merged-to-main because it lands on the 2.1 feature train, not `main`.

One pre-existing item you flagged stays open by design: the Copilot provider still identifies as `GitHubCopilotChat` (its API expects that). That's tracked separately, not a blocker here.

PR #609, opened 2026-09-03
fix(install-hooks): stop reporting a ZCode config ZCode has rejected as up to date

### LINKED ISSUE #600: install-hooks --agent zcode: apply reports no-op on stale/broken config; print and apply disagree on --auth-token
### Environment
- Client: ai-memory 2.0.0, windows-x86_64
- Agent: ZCode (z.ai) engine 0.16.5; hooks block in `~/.zcode/cli/config.json` previously generated by ai-memory 1.39.0

### Problem 1 — stale/broken config reported as "already up to date"
`install-hooks --agent zcode --apply` printed:

```
✓ no-op C:\Users\Administrator\.zcode\cli\config.json (already up to date)
```

while the existing hooks block was in fact stale **and silently broken**: entries generated by 1.39.0 (no `--auth-token` in args) plus `"SessionEnd": []` and `"PreCompact": []` event keys. ZCode 0.16.5 strict-validates the events map and rejected the entire hooks block (`config.file.invalid: hooks.events: Unrecognized keys: "SessionEnd", "PreCompact"`, startup diagnostics `hookCount: 0`) — so capture had never run at all, yet the freshness check reported up-to-date.

2.0.0's generated config itself is fixed (only the six event keys ZCode accepts, thank you), but the freshness check apparently only compares hook command paths — it should diff event keys/args too, otherwise a future format change will again silently mask a dead install.

### Problem 2 — print and apply disagree on `--auth-token`
The printed rendering says:

```
# Auth: token embedded in each hook's args below
```

and indeed shows `--auth-token` in the args, but `--apply` writes the hooks **without** `--auth-token`. Authentication then depends on `AI_MEMORY_AUTH_TOKEN` in the hook process environment or the data-dir `auth-token` file. If the omission is intentional (presumably related to #552, token visible on the command line), the printed rendering/comment should be updated to match; otherwise apply should embed it like the print shows.

### Workaround used
Strip all ai-memory entries from `config.json` (keeping user-authored hooks), re-run `--apply` (now reports `✓ updated`), and rely on the data-dir `auth-token` file for authentication (verified working). Capture then confirmed end to end on 2.0.0.

costajohnt:
## What changed

`install-hooks --agent zcode --apply` used to print `✓ no-op … (already up to date)` over a hooks block ZCode had rejected outright, so capture was dead (`hookCount: 0`) while the tool reported the install as current. Three changes:

- Apply now withdraws ai-memory's own entries from event keys this version does not write, wherever they sit. It did this only for the six keys in `ZCODE_EVENTS` before, so anything else was never even looked at.
- It reports what it found on stderr. A key it withdrew from gets a warning (a withdrawal is positive evidence ai-memory once lived there); a key left unable to run anything gets a note. Both name the key and the config file, and both state the consequence: ZCode rejects the entire block over one key it does not recognize.
- **Changed output:** when such a key is present and the file is otherwise unchanged, the result line now reads `no-op … (unchanged; see the notes below)` instead of `already up to date`. Without this the fix would not have helped the reported case at all, because the notes go to stderr and the reassuring line goes to stdout. Stdout is flushed before the notes so the two stay in order when merged.

Nothing is deleted. Event keys are never removed and hooks ai-memory did not write are never removed, so a key holding someone else's hook is reported, not touched.

## Why

Fixes #600.

One correction to the issue's diagnosis, which changed the shape of the fix. The report attributes the stale `SessionEnd`/`PreCompact` keys to ai-memory 1.39.0, and there is no "freshness check" that compares command paths. `ZCODE_EVENTS` has held the same six keys since ZCode support landed (`3f0af7d`, and `git show v1.39.0` agrees), so ai-memory never wrote those keys into a ZCode config. That is why this deletes nothing: a key ai-memory did not author is not ai-memory's to remove, and the useful thing it can do is say precisely what is wrong and where.

`apply_atomic` returns `NoOp` on a byte-for-byte match, and `build_zcode_hooks_config` is byte-identical between v1.39.0 and today, so a config in the reported state re-renders identically and stays `NoOp` no matter how dead it is. That is why the result wording had to change too.

## Test plan

- [x] Manual test: built the binary and ran `--apply --config-file` twice against the reported shape (`{"hooks":{"enabled":true,"events":{"SessionEnd":[],"PreCompact":[]}}}`). First run reports `updated` plus a note for each dead key; second run, with the file already current, reports `no-op … (unchanged; see the notes below)` on stdout and names both keys on stderr. Before this change that second run printed `already up to date` and nothing else. Also verified with a third-party hook under a foreign key: the foreign hook, sibling config keys (`model`), and all six real event keys survive untouched.

Two new tests: one drives `--apply` over a config mixing a dead key, a key holding only our entry, and a key mixing our entry with a hook we did not write, and asserts the foreign hook survives while ours is withdrawn; one enumerates the degenerate shapes (`[]`, `[{}]`, `[{"matcher":"*"}]`, `[{"hooks":[]}]`, `[{"hooks":null}]`, a non-array) against the runs-nothing predicate. I checked both are load-bearing by neutering the ownership filter and by removing the foreign-key pass; each makes the first test fail.

## Commit attribution

## Release impact

## CHANGELOG (merge gate)

- [x] Any changed default is called out explicitly in "What changed" above (the `NoOp` result wording)

## Notes for reviewers

Two things I deliberately did not do, both of which are your call rather than mine:

1. **The second problem in the issue, `--auth-token`.** The report is accurate: the dry run passes `Some(token)`, so `--auth-token` appears in the printed args along with `# Auth: token embedded in each hook's args below`, while `--apply` passes `None` and writes no such flag (`install_hooks.rs:347-365`). Given #552 deliberately took the token off the command line, I assume the printed rendering is what should change rather than apply embedding it, but that is a decision about intent, so I left it out of this PR rather than guess. Happy to send it as a follow-up either way.

2. **Removing dead keys automatically.** It would be a smaller experience than a note telling the user to delete them by hand, and the reporter did exactly that by hand. I did not do it because it means deleting configuration ai-memory never wrote, which would also change what the merge promises. If you would rather apply just clean these up, say so and I will send it.

One thing worth flagging: for the six keys ai-memory does write, the pre-existing group prune drops any matcher group left with an empty `hooks` array, including a user-authored empty group. That predates this change and I did not touch it, but the function's doc comment previously said "never edits matcher groups it did not write", which did not quite hold. I narrowed that sentence to what the code actually does and named the exception rather than leave the doc overclaiming.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- `install-hooks --agent zcode --apply` no longer reports a hooks block
+  ZCode has thrown away as `no-op … (already up to date)` (#600). ZCode
+  validates `hooks.events` strictly and rejects the whole block, including
+  ai-memory's own entries, over a single key it does not recognize, so
+  capture never ran (`hookCount: 0`). Apply only ever merged the six event
+  keys it writes, so a key it does not write was neither inspected nor
+  reported, the file kept matching byte for byte, and the report claimed
+  the install was current. Apply now also withdraws ai-memory's own
+  entries from event keys it no longer writes, wherever they sit, and
+  reports what it found: a warning naming any key it withdrew from, and a
+  note for a key left unable to run anything, both naming the config file.
+  **Changed output:** an unchanged ZCode config with such a key now reports
+  `no-op … (unchanged; see the notes below)` instead of `already up to
+  date`, so a redirected stdout no longer carries the reassurance without
+  the cause. Event keys are never deleted and hooks ai-memory did not write
+  are never removed, so a key holding someone else's hook is reported, not
+  touched.
 - `as_of` time-travel queries and the entity retrieval stream now work
   on real stores. Both read the entity index, which was populated only
   from an LLM consolidator's `entities:` frontmatter — absent on the
--- a/crates/ai-memory-cli/src/commands/install_hooks.rs
+++ b/crates/ai-memory-cli/src/commands/install_hooks.rs
@@ -15,6 +15,7 @@
 //!   and produces no backup.
 
 use std::fs;
+use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result};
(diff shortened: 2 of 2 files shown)


costajohnt:
Closing this one. You landed the fix in 724c467 while I had this in flight, and for the reported case your approach is the better one: stripping every ai-memory hook from every key and then dropping emptied groups and empty keys handles it directly, and my PR does not improve on that. Thanks for the quick turnaround, and for updating the printed snippet to spell out the `--apply` token behavior. That was the half I deliberately left alone rather than guess at, so it is good to see where you landed on it.

One residual I checked against e20e7d2, in case it is useful. Two shapes still reproduce the original symptom, because the group and key retain arms keep anything that is not an empty `hooks` array:

```
$ echo '{"hooks":{"enabled":true,"events":{"SessionEnd":[{"matcher":"*"}]}}}' > c.json
$ ai-memory install-hooks --agent zcode --apply --config-file c.json   # run twice
✓ no-op c.json (already up to date)
# SessionEnd survives as [{"matcher": "*"}]
```

A non-array value behaves the same way (`"SessionEnd": 42` survives untouched). Neither can run a hook, and both are shapes ZCode will not accept, so capture stays dead behind an up-to-date report.

Worth saying plainly: this is a hand-edited-config edge, not anything a released ai-memory ever wrote, so it may well not be worth your time. If you do want it closed off, I am happy to send a small follow-up that treats a group with no runnable hooks, and a non-array slot, as empty in those same two retains. Just say the word and I will keep it to that.

Issue #260, opened 2026-07-26
SessionEnd opt-in LLM consolidation is silently cancelled by the drain timeout and never retried


aguirreSL:
**Version**
`ai-memory 1.19.0`, server and client (also reproduces on 1.18.0).

**What happened**
With `AI_MEMORY_CONSOLIDATE_ON_SESSION_END=true` and a working Anthropic provider, every
session page lands as the M3 heuristic fallback. The consolidation neither succeeds nor
fails: no "SessionEnd: LLM consolidation written (opt-in)", no "SessionEnd LLM
consolidation failed". Session size does not matter; a 30-minute, 72-observation session
fails the same way.

**What I expected**
The opt-in consolidation to run, or to fail loudly and be retried on redelivery.

**Steps to reproduce**
1. Docker server with an Anthropic provider and `AI_MEMORY_CONSOLIDATE_ON_SESSION_END=true`;
   native Claude Code hooks with the default drain timeout (3s).
2. Run any session, exit cleanly.
3. The page stays `_Synthesised by ai-memory (M3, no-LLM heuristic)._`, the server logs no
   consolidation attempt, and ~30s later the redelivered SessionEnd logs "ignoring
   SessionEnd for missing, mismatched, or already-ended session".

**Root cause (1.19.0 sources)**
Three pieces interact:

1. The batch that carries `session-end` is small. `stop` is also a boundary event, so the
   spool is nearly empty at session end, and `batch_request_timeout` (`hook_spool.rs`)
   caps the POST at `per_event_timeout * item_count`: 3-6s for a 1-2 item batch.
2. `handle_hook_batch` (`router.rs`) processes items inline in the response window. The
   SessionEnd path writes the heuristic page, persists `end_session`, inserts the
   auto-handoff, then awaits `consolidate_session(...)` in the same request future. When
   the client hangs up at the cap, the future is dropped and the await is cancelled
   between the two log branches, so nothing is logged. The store's recurring "writer
   reply dropped / caller cancelled" warnings are the same mechanism hitting ordinary
   ingests.
3. `end_session` is persisted before the cancelled await, so the redelivery resolves to
   `SessionEndDisposition::DropStale` and the LLM step is never retried.

The consolidation loses a race it cannot win: the only request allowed to run it has the
tightest window of the whole drain, and the dedup that makes redelivery safe also makes
the loss permanent.

**Evidence**
[...]

PR #615, opened 2026-09-03
fix(bootstrap): tolerate a chunk that returns no pages

### LINKED ISSUE #614: `bootstrap`: a chunk that returns no `pages` key aborts the whole multi-chunk run (`serde: missing field 'pages'`), discarding every page the earlier chunks produced
**Version**
`ai-memory 2.0.1`, server (Docker, `akitaonrails/ai-memory:latest`) and client (macOS aarch64 native binary). Provider: `anthropic` / `claude-haiku-4-5`.

**What happened**
`ai-memory bootstrap` on a docs-heavy repository fails after ~20 minutes with:

```
Error: server returned 502 Bad Gateway: {"error":"serde: missing field `pages`"}
```

Every page the successful chunks produced is lost. The run is not resumable, so the next attempt starts from chunk 1 and pays for all the LLM calls again.

**What I expected**
A chunk that has nothing new to add is a normal outcome — later chunks are told which paths earlier ones already wrote, so "the material is already covered" is a correct answer. I expected that chunk to contribute zero pages and the run to continue.

**Steps to reproduce**
1. Point `bootstrap` at a repository whose sources are mostly `docs/**/*.md` rather than git commits, large enough to split into several chunks. Mine: 70 sources collected / 36 kept, 0 git commits, 33 doc files, `llm_chunks=9`.
2. `ai-memory bootstrap`.
3. It fails partway through with the error above. Reproduced on 2 of 4 attempts (the other two failed for unrelated transient reasons — see "Other failure modes" below).

**Root cause (2.0.1 sources)**

`BootstrapBatch` in `crates/ai-memory-consolidate/src/bootstrap.rs:161-168` makes `pages` mandatory while `rationale` is defaulted:

```rust
pub struct BootstrapBatch {
    /// Pages to create.
    pub pages: Vec<BootstrapPage>,
    #[serde(default)]
    pub rationale: String,
}
```

The Anthropic provider requests structured output through a `tool_use` block with `input_schema` (`crates/ai-memory-llm/src/anthropic.rs:255-260`). Unlike OpenAI's `response_format` with `strict: true`, that schema is advisory — the model can and does answer `{"rationale": "..."}` with no `pages` key when it judges a chunk fully covered by earlier ones. serde then rejects the payload.

The blast radius comes from where that failure lands. In `process_sources` (same file), each chunk is awaited inside the loop with `?`:

```rust
for (idx, chunk) in chunks.iter().enumerate() {
    ...
    let batch: BootstrapBatch = complete_structured(&*self.llm, request).await?;  // :373
    ...
}
// pages are only written after the loop, at :390
```

Pages accumulate in `pages_by_path` and are written **after** the loop completes. So a rejected payload on chunk *n* propagates out of the function and discards chunks *1..n-1* as well. In my case chunk 9 of 9 threw away eight successful calls.

The rest of the pipeline already handles an empty batch correctly: `insert_bootstrap_page` is simply never called, and `merged_pages` ends up shorter. Only the deserialisation step objects.

**Fix**

Adding `#[serde(default)]` to `pages` makes an empty batch a normal outcome. PR attached.

**Other failure modes seen on the same repo (not part of this issue)**

Reporting these only as context for why the missing-`pages` bug is expensive rather than merely annoying — each cost a full ~20-minute run:

- `Connection reset by peer (os error 54)` — the `bootstrap` POST stays open for the whole multi-chunk run with no bytes flowing, and a reverse proxy in front of the server closes it. Fixed on my side with `transport http { read_timeout 0; write_timeout 0 }` in Caddy. Possibly worth a note in `docs/https-via-proxy.md`, since the shipped Caddy template in `docker/compose.tls.caddy.yml` does not mention it.
- `provider error 520` — a transient Anthropic-side error on a single chunk.

Both are external to ai-memory, but they share the same amplifier as the bug above: **any** single-chunk failure discards the entire run. `#[serde(default)]` fixes the case ai-memory controls; whether partial progress should survive a genuinely failed chunk (write incrementally per chunk, or persist progress for a `--resume`, in the spirit of the durable queue from #260) is a design call I did not want to presume in a PR. Happy to open a separate issue for that if it is worth pursuing.

pepaulo8:
## What changed

`BootstrapBatch::pages` is now `#[serde(default)]`.

Before: a chunk answering `{"rationale": "..."}` with no `pages` key failed deserialisation with `missing field 'pages'`, and because that error propagates out of the chunk loop before any page is written, the whole run was discarded — including every page the earlier chunks had produced.

After: that answer deserialises as an empty batch, the chunk contributes nothing, and the run continues to the next one.

No default changes: a batch that *does* carry `pages` behaves exactly as before.

## Why

Fixes #614.

Later chunks are handed the paths earlier ones already wrote, so "the remaining material is already covered" is a correct answer for a model to give — particularly on repositories whose sources are documentation rather than commits, where a handful of dense files can be fully described by the first few chunks. Anthropic's `tool_use` schema (`crates/ai-memory-llm/src/anthropic.rs:255-260`) is advisory rather than enforcing, unlike OpenAI's `strict: true`, so that answer reaches serde.

The rest of the pipeline already tolerates an empty batch — `insert_bootstrap_page` is simply never called and `merged_pages` ends up shorter. Only the deserialisation step objected.

## Test plan

- [ ] `cargo test --workspace` — **not run locally**, see "Notes for reviewers"
- [x] Manual test: reproduced on a docs-only repository (70 sources collected / 36 kept, 0 git commits, 33 doc files, `llm_chunks=9`) against `claude-haiku-4-5`. The unpatched server failed with `502 {"error":"serde: missing field 'pages'"}` on 2 of 4 attempts, each time discarding a ~20-minute run. Added `batch_without_pages_key_deserialises_as_empty` to cover the payload shape directly.

## Commit attribution

## Release impact

## CHANGELOG (merge gate)

- [x] Any changed **default** is called out explicitly in "What changed" above (none — behaviour for batches that carry `pages` is unchanged)

## Notes for reviewers

**On the missing `cargo test --workspace`:** I could not run the full suite locally. On aarch64 the build fails in `gemm-f16` (`error: instruction requires: fullfp16`) pulled in through `candle`, and the x86_64 path under emulation was too slow to finish. `fmt` and `clippy` both pass on aarch64. The change is one attribute plus one test that only exercises `serde_json::from_str`, so I am relying on CI here — please bounce this back if you would rather see a green local run first.

**On scope:** I deliberately kept this to the deserialisation fix. The larger issue — that *any* single-chunk failure discards the entire run, because pages are only written after the loop at `bootstrap.rs:390` — is a design question (write incrementally per chunk, or persist progress for a `--resume`, along the lines of the durable queue from #260). I did not want to presume a direction in a PR. Happy to open a separate issue or take a stab at it if you have a preference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- `bootstrap` no longer aborts the whole multi-chunk run when one chunk
+  returns no `pages` key (#614). Later chunks are told which paths
+  earlier ones wrote, so a model that judges the material already
+  covered legitimately answers with a rationale and no pages; because
+  Anthropic's `tool_use` schema does not enforce required fields the way
+  OpenAI's `strict: true` does, that answer reached serde and failed
+  deserialisation with `missing field 'pages'`. Since pages are only
+  written after every chunk completes, the error discarded the work of
+  all preceding chunks. `pages` now defaults to empty.
 - `install-hooks --agent antigravity-cli` on native Windows now emits
   an unquoted hook `command` (#611). The native-binary rendering wrapped
   the executable, `--data-dir`, and `--server-url` in double quotes, but
--- a/crates/ai-memory-consolidate/src/bootstrap.rs
+++ b/crates/ai-memory-consolidate/src/bootstrap.rs
@@ -160,6 +160,15 @@ pub struct BootstrapPage {
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct BootstrapBatch {
     /// Pages to create.
+    ///
+    /// Defaulted because a chunk legitimately has nothing to add: later
+    /// chunks receive the paths earlier ones already wrote, so a model
+    /// that judges the material covered answers with a rationale and no
+    /// `pages` key. Anthropic's `tool_use` schema does not enforce
+    /// required fields the way OpenAI's `strict: true` does, so that
+    /// answer reaches serde — and without a default it aborts the whole
+    /// multi-chunk run, discarding every page the earlier chunks paid for.
+    #[serde(default)]
     pub pages: Vec<BootstrapPage>,
     /// Brief one-paragraph note about what was processed, surfaced
     /// in the wiki/bootstrap.md manifest + the auto-commit message.
@@ -1554,6 +1563,16 @@ mod tests {
         assert_eq!(collected, 27_873);
         assert_eq!(sent, 1);
         assert_eq!(dropped, 27_872);
+    }
+
+    #[test]
+    fn batch_without_pages_key_deserialises_as_empty() {
+        let batch: BootstrapBatch =
+            serde_json::from_str(r#"{"rationale":"already covered by earlier chunks"}"#)
+                .expect("a chunk that adds no pages must not fail the whole run");
+
+        assert!(batch.pages.is_empty());
+        assert_eq!(batch.rationale, "already covered by earlier chunks");
     }
 
     #[test]


pepaulo8:
Update on the missing `cargo test --workspace` I flagged above: I ran the project's own `ci.yml` via `workflow_dispatch` on my fork, on this exact commit (4e4cdaa). All 15 jobs pass, including both test legs:

- `rustfmt`, `clippy`, `cargo-deny`, `cargo-audit`
- `test (ubuntu-latest)` and `test (macos-latest)`
- `changelog sections`, `gitleaks`, the four release builds, and the docker image smoke test

Run: https://github.com/pepaulo8/ai-memory/actions/runs/33794592989

So the aarch64 `gemm-f16` build failure I hit locally is specific to my machine, not the change. Nothing outstanding on the test plan from my side.

Issue #613, opened 2026-09-03
2.0 OKF migration seeds index.md into project directories with no project row; the watcher then fails on them on every pass, forever


samirhvbr:
After upgrading a production instance to 2.0, the wiki watcher logs the same reindex failure for the same paths on every reconciliation pass and never converges.

```
WARN ai_memory_wiki::watcher: reconcile reindex failed path=index.md
     error=not found: project 01a038ea-… does not belong to workspace 019f23aa-…
```

## Measured on the affected instance

| | dirs | pages (`.md`, excluding `index.md`/`_meta.md`) |
|---|---|---|
| project dirs with a row in `projects` | 96 | 2842 |
| **project dirs with no row** | **77** | **73** |

- ~600 of those warnings every 2 minutes, indefinitely
- ~60 MB/day of log
- Store for scale: 2926 pages, 2477 sessions, 306k observations, `wiki format: OKF v0.2 (migrated)`

The orphan directories are near-empty shells — 73 pages across 77 of them, almost all `log-<month>.md` from months ago. **No meaningful memory is stranded**, so this is not a data-loss report.

## Why it is worth fixing anyway

The cost is not disk, it is signal. That volume of identical warnings buries everything else in the log. On this host the ai-memory data directory sat on a 7.6 GB root filesystem that filled up — the OKF migration's own (correct, backup-gated) 447 MB safety archive was the last straw — and the server then logged `sqlite: disk I/O error` **for 22 hours** without anyone noticing, because those lines were a rounding error next to the reconcile spam.

A warning that repeats every two minutes and never changes state is not a warning. It is noise wearing a warning's label, and it hid a real outage.

## Where it comes from

`ensure_bundle_indexes` enumerates project directories from the filesystem, never from the database:

https://github.com/akitaonrails/ai-memory/blob/v2.0.0/crates/ai-memory-wiki/src/migrations/m2026_09_okf_conformance.rs#L247-L266

```rust
let Ok(projects) = std::fs::read_dir(ws.path()) else { continue; };
for proj in projects.flatten() {
    // …
    let index = proj.path().join("index.md");
    if index.exists() { continue; }
```

So every directory gets an `index.md`, including ones the store knows nothing about. The watcher then picks that file up, resolves its scope from the path, and fails — every pass, forever. Before 2.0 these directories were inert; the migration is what made them loud.

## Two places this could be fixed

I am deliberately not picking one, since the call is yours:

[...]

PR #616, opened 2026-09-03
fix(watcher): apply the orphan skip to directory events too

samirhvbr:
Follow-up to 249c616b (#613). Small and deliberately narrow — please read the "what this is not" section before weighing it.

## The gap

`reconcile` got the orphan guard: check the project scope once per directory, skip a rowless one wholesale at `debug`. A project directory reaches the indexer by a second route, and it kept the old behaviour:

```rust
if ft.is_dir() {
    let Some((ws, proj, proj_root)) = extract_project_dir_ids(wiki.root(), raw_path) else { continue; };
    reindex_project_dir(wiki, ws, proj, proj_root).await;   // no scope check
    continue;
}
```

`reindex_project_dir` walks the tree and lets every page fail scope resolution on its own — one `warn` each — which is the behaviour that pass was about.

## What this is not

**It is not a second incident, and the functional outcome does not change.** `reindex_page` makes the same `ensure_project_workspace` call, so no page from a rowless directory was ever indexed by this path either. What changes is one `debug` line instead of one warning per page, and not walking a tree whose every entry is going to fail.

It is also far quieter than what #613 reported: it needs a filesystem event on that specific directory rather than firing every 30 seconds, so it could never have produced the flood that buried the 22-hour outage on our host. I am sending it as consistency in the sibling function, not as a bug — if you would rather leave the event path alone, that is a reasonable call and no argument from me.

## On the return value

`reindex_project_dir` now returns whether it indexed. Same reasoning you gave for `ReconcileStats`: the skip has to be observable to be testable. The caller ignores it; if you prefer it folded into the caller as a counter, or a `ReconcileStats`-shaped type instead of a `bool`, say so and I will change it.

## Verification

- New test `directory_events_skip_project_dirs_with_no_store_row`, checked both ways: **fails** without the guard, passes with it.
- Worth mentioning because it nearly shipped wrong: my first version asserted the orphan page was absent from the store, and **passed with and without the guard** — the store is precisely what does not differ between the two. Asserting on the skip is what makes it a regression test rather than a decoration.
- `directory_event_reindexes_project_markdown` (the resolvable-directory case) still passes, so the legitimate path is untouched.
- `cargo test -p ai-memory-wiki` — 142 passed. `cargo fmt --check` and `cargo clippy --all-features` clean.

### DIFF
--- a/crates/ai-memory-wiki/src/watcher.rs
+++ b/crates/ai-memory-wiki/src/watcher.rs
@@ -244,21 +244,39 @@ async fn handle_event(wiki: &Wiki, event: notify_debouncer_full::DebouncedEvent)
     }
 }
 
+/// Returns `false` when the directory was skipped because the store has no
+/// row for it — the same orphan case `reconcile` counts as `skipped_orphans`,
+/// surfaced here so the skip is testable on this path too.
 async fn reindex_project_dir(
     wiki: &Wiki,
     ws: WorkspaceId,
     proj: ProjectId,
     proj_root: std::path::PathBuf,
-) {
+) -> bool {
+    // Same orphan guard `reconcile` applies (#613), for the other way a
+    // project directory reaches the indexer. A filesystem event on a rowless
+    // directory would otherwise walk it and warn once per page, which is the
+    // behaviour that pass was about — quieter here only because it needs an
+    // event rather than firing every 30s. Checking once per directory also
+    // saves walking a tree whose every page is going to fail scope resolution.
+    if let Err(e) = wiki.ensure_project_workspace(ws, proj).await {
+        debug!(
+            workspace = %ws,
+            project = %proj,
+            error = %e,
+            "skipping directory event for a project directory with no store row",
+        );
+        return false;
+    }
     let pages = match tokio::task::spawn_blocking(move || walk_markdown(&proj_root)).await {
         Ok(Ok(pages)) => pages,
         Ok(Err(e)) => {
             warn!(error = %e, "watcher directory walk failed");
-            return;
+            return true;
         }
         Err(e) => {
             warn!(error = %e, "watcher directory walk task failed");
-            return;
+            return true;
         }
     };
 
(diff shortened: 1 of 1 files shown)


abhisheksharma2411 (commented):
Checked all three of your claims against the tree rather than taking the description's word, and they hold.

**The guard parity is real.** `reconcile` has `ensure_project_workspace` and the event path did not; on this branch both do (`watcher.rs:262` and `:320`).

**"The functional outcome does not change" is accurate**, which is the part I most wanted to verify before treating this as a no-op — `Wiki::reindex_page` calls `ensure_project_workspace` at `wiki.rs:1294`, so no page under a rowless directory was ever indexed by this route either. This really is noise and wasted walking, exactly as you framed it, and I appreciate you saying so in the PR rather than letting it read as a second incident.

**The test is load-bearing.** Deleting just the new guard fails it:

```
test watcher::tests::directory_events_skip_project_dirs_with_no_store_row ... FAILED
  panicked at crates/ai-memory-wiki/src/watcher.rs:858
```

With the guard restored: `125 passed` in `ai-memory-wiki`, clippy clean on the crate.

Two small things, neither blocking.

**1. The `bool` is consumed only by the test.** The production call site still discards it:

```rust
// watcher.rs:219
reindex_project_dir(wiki, ws, proj, proj_root).await;
```

Your doc comment is upfront that it exists to make the skip testable, so this is declared rather than hidden. But the test already proves the same thing through the store — `search_pages("eventtoken")` comes back empty — so the return value may be carrying less weight than it looks. Worth asking whether it earns its place, or whether the store assertion alone says it.

**2. If it stays, the name is slightly off.** `true` is also returned when the walk itself fails:

```rust
Ok(Err(e)) => { warn!(...); return true; }
Err(e)     => { warn!(...); return true; }
```

Nothing was indexed in either case, so the value means "the directory had a store row", not "it indexed". Your doc comment says precisely that, but the test binds it to `let indexed = ...`, which reads as the other thing. `had_store_row` (or asserting on the skip some other way) would keep the reader from having to reconcile the two.

Neither changes the outcome, and I would take this as-is. Sending it because the sibling-consistency argument is the right one and the PR is easy to verify — which is not the norm.

PR #619, opened 2026-09-03
fix(llm): route Luna through Responses API

### LINKED ISSUE #618: Route GPT-5.6 Luna through OpenCode Go Responses API
## Business Value

OpenCode Go users can use `gpt-5.6-luna` for ai-memory exploration and structured workflows without every request failing with HTTP 500.

## Problem

`OpenCodeProvider` sends every model to `/chat/completions`. OpenCode Go publishes `gpt-5.6-luna` through `/responses`, so both plain and structured ai-memory calls fail with an internal server error. Other models such as `deepseek-v4-flash` still use Chat Completions.

## Proposed Fix

- Route `gpt-5.6-luna` to `/responses`.
- Use Responses request and response schemas for plain and strict structured completions.
- Preserve Chat Completions routing for the rest of the OpenCode Go catalogue.
- Preserve ai-memory's user-agent and logical-operation session header.

## Verification

Add provider-level regression tests for Luna plain completion, Luna structured completion, and a Chat Completions compatibility model.

lucazz:
## What changed

- Route `gpt-5.6-luna` through OpenCode Go's `/responses` endpoint.
- Add typed plain and strict structured Responses request/response handling.
- Keep other OpenCode Go models on `/chat/completions`.
- Preserve the ai-memory user-agent, logical-operation session header, timeout, and reasoning effort.
- Update the provider documentation and changelog.

## Why

- OpenCode Go does not serve Luna through Chat Completions, so ai-memory plain and structured calls returned HTTP 500.
- Closes #618.

## How to verify

- `cargo test -p ai-memory-llm`
- `git diff --check`
- `TAILWIND_SKIP=1 cargo test --workspace` reaches 806 passing tests but is blocked by four pre-existing `install_hooks` failures; an exact failing test reproduces from merge-base `3e6db6d`.
- Live `llm-test` was not run because the isolated worktree session had no `OPENCODE_API_KEY`.


akitaonrails:
Reconciled onto `release/2.1` as merge `fc592a78` (your commits preserved), queued for 2.1.0. It was based on `main`, which predates 2.0.2's #606 (OpenCode operator headers + the Go/Zen base-URL split, both on the 2.1 branch), so the merge conflicted in `opencode.rs`/`install.md`. I kept both:

- Your transport enum and the `/responses` routing for `gpt-5.6-luna` are intact.
- #606's `with_base_url`/`with_extra_headers` and the `OPENCODE_GO_BASE_URL` rename are re-applied on top, threading **both** transports. The one substantive addition beyond a textual merge: `OpenCodeResponsesProvider::post` now layers headers through `ExtraHeaders::set_default`+`apply` (mirroring `openai.rs`) instead of hardcoding them — so an `AI_MEMORY_LLM_HEADERS` operator override wins on the Luna Responses path too, not just Chat Completions. Without that, Luna would have silently ignored operator headers.

Gate on the merged tree: `clippy --workspace -D warnings` clean; full workspace tests **0 failed** — #606's header/base-url tests and your Luna Responses tests pass together. Nicely structured PR; the transport enum made the reconcile clean. Closing as resolved-and-ready-to-release rather than merged-to-main because it lands on the 2.1 feature train.

One thing I couldn't verify locally: OpenCode Go's live `/responses` wire format — I'm trusting your mock tests and the linked Go docs for that shape. If you've run it against the live endpoint, all the better.

PR #605, opened 2026-09-03
feat(install-hooks): wire Codex's SessionEnd hook (Codex CLI 0.145.0)

holocaster:
## What changed

`install-hooks --agent codex` (and `setup-agent --agent codex`) now wires
Codex's `SessionEnd` lifecycle hook in addition to the six events it wired
before. A finished Codex session now gets the same automatic end-of-session
summary and cross-agent handoff that Claude Code already gets, instead of
needing a manual `ai-memory finalize-session --agent codex`.

- `CODEX_EVENTS` gains `("SessionEnd", "session-end.sh")`. The
  `hooks/codex/session-end.{sh,ps1}` scripts already shipped in the bundle
  and already POST `event=session-end&agent=codex`; they were just never
  listed in the profile. The `.ps1` is picked up automatically by the
  Windows extension-swap in `script_for_platform`.
- `merge_codex_payload` no longer unconditionally strips a `SessionEnd`
  key on every apply. It relied on that removal because the event was
  considered Claude-Code-only; now it goes through `overlay_event_hooks`
  like every other event, so a user's own third-party `SessionEnd` hook is
  preserved and ai-memory's entry is added alongside it.
- No on-disk / schema / default-behaviour change beyond the extra hook
  entry. Re-running the installer adds one `SessionEnd` entry; Codex
  prompts once to trust it.

## Why

Closes #604.

Codex had no `SessionEnd` hook when the Codex integration was written, so
the code, tests, and `docs/support-matrix.md` all encoded "Codex has no
true session-end; use `finalize-session`". Codex shipped a first-class
`SessionEnd` event in **Codex CLI 0.145.0** (openai/codex#33895, "Add
SessionEnd hooks for thread teardown"; originally requested in
openai/codex#20374). `rust-v0.145.0` is the first release tag that contains
the commit; `rust-v0.144.0` does not.

## Test plan

- [x] `cargo fmt --all -- --check` passes
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `cargo test --workspace` passes
- [x] Manual: inspected the generated `~/.codex/hooks.json` from
      `install-hooks --agent codex` — `SessionEnd` present with the same
      nested `{matcher, hooks:[{type:command, command}]}` shape as the
      other events; `session-end.ps1` staged alongside `session-end.sh`.

## Commit attribution

- [x] I verified the name and email on every commit in this PR.

## Release impact

- [x] **Minor** — additive: wires an existing-but-unlisted lifecycle event
[...]

PR #622, opened 2026-09-04
feat(opencode2): add first-party OpenCode 2.0 beta support

enrell:
## What changed

First-party OpenCode 2.0 beta (`opencode2`, side-by-side with v1) support, following the Kiro v2/v3 precedent (new CLI/harness variants reusing the existing `AgentKind::OpenCode` — no new kind, no store migration):

- `install-mcp --client opencode2`: V2 `mcp.servers` remote entry (no `enabled` field, `oauth: false` so the beta skips OAuth discovery against the local endpoint). Same `opencode.json(c)` file as v1 — the beta explicitly supports mixed V1/V2 members under `mcp`.
- `install-hooks --agent opencode2`: generates `ai-memory-opencode2.ts` in the beta `{ id, setup }` plugin shape with `ctx.session.hook("prompt"|"context")`, `ctx.tool.hook("execute.before"|"execute.after")`, and `ctx.event.subscribe` lifecycle coverage (`session.created/idle/deleted`, plus `session.compaction.started`). Handoff injection moved from v1's removed `experimental.chat.system.transform` to the `context` hook (once-per-session guard; `{ type: "text", text }` items per the runtime schema). Hook registrations are disposed on unload. The capture prelude (queue + spooling, marker resolution, capture policy) is shared verbatim with v1 via anchored rewrite.
- `ai-memory run opencode2`: launches through the `opencode2` binary (`--session` placement, `--auto` yolo, beta utility-subcommand list, all verified on beta-18999) against the shared `opencode.db`.
- New read-only transcript adapter for the beta tables (`session_v2` / `session_message`): user text, assistant text, completed tool calls/results; encrypted reasoning becomes a named loss.
- `setup-agent`, `uninstall` (per-file ownership markers + nested `mcp.servers` removal), `scripts/install-hooks.sh`, and the managed-workstream acceptance script gain `opencode2` cases.
- Docs (`install`, `support-matrix`, `managed-workstreams`, `mcp-install`, `usage`, `use-cases`, `windows`, `managed-harness-contributions`, `AGENTS.md`, README) and a `CHANGELOG.md` `[Unreleased]` `Added` entry.

## Why

The beta installs side by side as `opencode2` with an incompatible MCP schema, plugin API, and session-store schema. Without this, ai-memory captures nothing on 2.0.

## What this is not

- Not a new agent kind: v1/v2 sessions share attribution deliberately, and bare `ai-memory run` auto-selection stays on v1 (listing both would duplicate every candidate).
- Not a claim of full managed-delta acknowledgement: the beta's shared background service cannot see a managed run's environment, so the ledger-delta ack (`context_delivered`) does not fire for opencode2 legs. Failure mode is redelivery, never loss; cross-harness continuity into opencode2 arrives through the ordinary handoff loop. Documented in `managed-workstreams.md`.
- The two plugin files share one auto-loaded dir; a host may warn about its sibling's file. Benign; `uninstall` keys each file to its own markers.

## How to verify

- `cargo fmt --all -- --check` and `git diff --check` clean.
- `cargo test -p ai-memory-core -p ai-memory-cli -p ai-memory-workstream`: all green (incl. 10 new opencode2 tests: wire aliases, MCP shape/upsert, plugin binding + drift-failure path, harness contract, v2 transcript fixtures, CLI alias parsing).
- Rendered plugin passes `tsc --noEmit` against `@opencode-ai/plugin@beta`; `install-mcp --apply` is idempotent with v1+v2 coexisting; `uninstall` dry-run detects both.
- Live against beta-18999: hook capture persisted (`open-code` session + observations), handoff loop proven end-to-end (fresh session output a sentinel present nowhere in its own prompt), `finalize-session --agent opencode2` works.
- `scripts/managed-workstream-acceptance.sh` with `AI_MEMORY_ACCEPTANCE_HARNESSES="opencode opencode2"`: **passed**, returned to the same native session (proves managed launch/capture/export/cross-binary resume).
- `scripts/check-changelog-sections.sh`, `check-changelog-frozen.sh`, `check-native-packaging.sh`: pass. No dependency changes (`cargo deny`/`audit` unaffected). No secrets added (gitleaks-relevant literals: none).

## Notes for reviewers

- **Full-workspace clippy**: green in every touched file; the only failures are pre-existing new-toolchain lints (`result_large_err`, `chunks_exact_to_as_chunks`) in untouched `ai-memory-web` / `ai-memory-consolidate` — this environment forces Rust 1.98 over the pinned 1.95 (CI uses stable, and `main` is green there). Separately verified `cargo check --workspace --all-targets` passes on a locally installed 1.95 toolchain.
- **Two beta caveats are documented in `install.md`**: back up `~/.local/share/opencode` first (the beta has migrated the shared database in place before — upstream #42260), and move the beta background service off port 49374 when ai-memory serves there (both default to it; the service crash-loops otherwise — observed live).
- **Acceptance script change**: pins the fixture opencode2 service at a free port for the same collision reason.
- Commit attribution: please confirm `enrell <enrellsa10@proton.me>` maps to the author's GitHub identity.

## Release impact

- [x] **Minor** — new agent-harness surface (`opencode2` client/agent/run targets), additive only
- [x] I added a `CHANGELOG.md` `[Unreleased]` entry (reference number to fill on open)


akitaonrails:
Reconciled onto `release/2.1` as merge `30ef4812` (your commits preserved), queued for 2.1.0. It targeted `main`, but new-harness support is a minor feature, so it lands on the 2.1 train per our semver policy.

The static review was clean — additive throughout: new CLI enum arms, wire aliases folded into the existing `AgentKind::OpenCode` (no new variant, no store migration), a read-only, fully-parameterized SQLite transcript adapter for the beta's `session_v2`/`session_message` tables, and an `ai-memory-opencode2.ts` plugin derived from v1 with a loud anchor-drift guard. No dependency/workflow/binary changes; loopback-only URLs; argv-only harness launch.

One reconcile adjustment: I sequenced this after #625 (generated-TS auth fix, merged to main in this batch), so the opencode2 plugin inherits its runtime `resolveToken()` — the plugin test assertion moved from `Bearer ${TOKEN}` to `Bearer ${token}` accordingly. CHANGELOG, mcp-install, and support-matrix rows reconciled with the Codex-SessionEnd (#605) and capture-assistant (#627) wording already on the branch.

Gate on the merged tree: fmt + clippy `-D warnings` clean; full workspace tests green. Residual I couldn't verify locally: your live beta-18999 claims (hook capture, handoff loop, managed resume/import) are trusted from your report, not independently reproduced. Nicely-built feature — the v1-template-rewrite-with-drift-guard approach kept it clean. Closing as resolved-and-ready-to-release on the 2.1 branch.

enrell:
Live verification of the merged feature (follow-up to the residual noted in the reconcile comment — the beta claims that were trusted but not independently reproduced).

Tested on this machine with the merged tree, `release/2.1` @ `bcdfd44e` (includes the #625 sequencing fix), against `opencode2` v0.0.0-beta-19059, launched via `ai-memory-2.1 run opencode2`:

- **Managed launch** — the harness arm is recognized end-to-end (`run --help` lists `opencode2: OpenCode 2.0 beta`, side-by-side with v1).
- **Plugin** — `ai-memory-opencode2.ts` loaded by the beta with the runtime `resolveToken()` → `Bearer ${token}` auth path (the #625-inheriting form); no anchor-drift warnings.
- **Hook capture** — observations incremented live during the session (1,644 → 1,646 mid-conversation), so the sanitizer → spool → writer-actor chain is recording real beta prompts/tool events.
- **MCP read round-trip** — `memory_status`, `memory_briefing`, `memory_query` all answered from inside the beta session; FTS5 search returned the expected hit with `<mark>` highlighting.
- **MCP write round-trip** — `memory_write_page` → `memory_read_page` → `memory_delete_page` all green (page persisted with derived title + frontmatter, then removed cleanly). Atomic write + same-transaction index commit confirmed by the returned checkpoints.
- **Default model config** — v1 and v2 share `~/.config/opencode/opencode.json`; `opencode2 debug config` parses the shared default model correctly.

All gates re-verified on the installed build: service healthy, `fmt`/`clippy`/workspace tests green at the merge. The only cosmetic note: the binary still reports `2.0.2` — expected, since the version bump lands with the 2.1 release tag.

Thanks for the clean reconcile — the v1-template-with-drift-guard approach held up exactly as designed under a live beta.

Issue #627, opened 2026-09-04
Codex hooks cannot capture assistant responses with the documented double opt-in


caam1406:
## Problem

The Codex integration accepts MCP and lifecycle hooks, but ai-memory 2.0.2 rejects the documented assistant-response capture flag for the Codex agent. This prevents a complete sanitized record of prompts, tool calls, assistant response, and handoff.

## Environment

- ai-memory: v2.0.2
- release commit: 7e787c919c5617beb531dafdd136c4163728b05d
- Docker image: docker.io/akitaonrails/ai-memory:2.0.2
- Docker digest: sha256:14eef51d7c5d560066491699a2ceeee37a68f968a380ffb54ef578b888dff96e
- Host: Windows native Codex with Docker Desktop Linux engine
- Server: http://127.0.0.1:49375 (loopback only)

## Reproduction

1. Start ai-memory 2.0.2 with capture_assistant=true and zero external LLM/embedding providers.
2. Run: ai-memory install-hooks --agent codex --capture-assistant --apply --server-url http://127.0.0.1:49375
3. The installer rejects --capture-assistant for --agent codex.
4. Install hooks without that flag, run a session with a prompt and tool call, finalize-session, and inspect MCP/UI.

## Expected

The documented double opt-in should provide a supported Codex path for a sanitized, bounded assistant final-turn observation, or documentation must explicitly state that capture is Claude-Code-only.

## Observed

- MCP works after Codex restart.
- Hooks capture sanitized prompts, tool lifecycle events, and session metadata.
- Smoke counts: pages_latest=1, pages_all=1, sessions=1, observations=5.
- Synthetic bearer, API key, cookie, and email fixtures were absent from database, wiki, spool, and logs.
- Loopback binding is enforced and invalid Host returns HTTP 403.
- Codex rejects --capture-assistant, so no assistant response is captured.
- UI shows montandoideias/emulador but no navigable assistant-response or handoff page.

## Impact

Projects requiring complete sanitized lifecycle evidence cannot meet that policy with the Codex profile. MCP, prompt capture, tool capture, persistence, and sanitization otherwise work.

## Privacy

All test values were fictitious. No credentials, tokens, personal data, ROMs, BIOS paths, firmware, dumps, saves, or raw private transcripts are included here.

## References

- https://github.com/akitaonrails/ai-memory/blob/main/docs/install.md
- https://github.com/akitaonrails/ai-memory/blob/main/docs/support-matrix.md
[...]

Issue #196, opened 2026-07-18
Design explicit opt-in assistant/Stop capture


akitaonrails:
## Context

Issue #190 added safe metadata-only PreToolUse context while preserving bounded PostToolUse excerpts. Capturing last_assistant_message from Stop remains intentionally disabled because assistant text is not path-attributable and can quote code, secrets, or private content.

## Scope

Design a separate explicit opt-in for assistant/Stop capture with:

- disabled-by-default behavior and clear upgrade semantics,
- typed agent-specific payload schemas,
- sanitization before spool/network/storage,
- strict UTF-8-safe size caps,
- privacy/capability documentation,
- reviewer and persistence sentinel tests,
- no claim that path exclusions make assistant text safe.

Do not enable raw assistant-message capture by default or infer unsupported agent fields.

Follow-up to #190.

akitaonrails:
Post-v1.15 issue audit: this enhancement correctly remains open and unimplemented. Current behavior still keeps Stop/assistant-message capture disabled (`docs/marker-file.md`), and `unsupported_and_stop_tool_payloads_never_render_content` in `crates/ai-memory-hooks/src/payload.rs` verifies sentinel assistant content cannot reach persisted hook content. No merged PR claims to implement this opt-in design; close this issue only after the scoped privacy/configuration work and regression coverage land.

akitaonrails:
Post-v1.17.2 clarification after live testing and PR #221:

- The optional assistant/Stop capture proposed here is still **unimplemented and intentionally disabled**.
- Native `ai-memory hook` commands strip Claude Code's raw `last_assistant_message` before local spooling or network transport; legacy spooled entries are stripped while draining.
- The server also strips the field immediately on `/hook` and `/hook/batch` receipt before persistence.
- Docker-wrapper and explicit `AI_MEMORY_HOOK_PLATFORM=posix` installs remain script fallback. Running `install-hooks` through the Docker wrapper refreshes those scripts but cannot convert them to native host commands, because a helper-container binary path is not usable by the host harness.
- To eliminate the residual client-to-server local-wire exposure for Claude Stop payloads, install a native ai-memory client on the agent host and use that native executable to reinstall the Claude Code hooks. Retaining the script fallback still does not persist the field because the server strips it on receipt.

[...]

PR #631, opened 2026-09-04
docs(support-matrix): state that Codex has no assistant-response capture

samirhvbr:
Closes the documentation half of #627.

## Why the reporter read this as a gap

The behaviour they hit is deliberate and tested — `capture_assistant_allowed` admits Claude Code on a native hook platform only (#196), and `install-hooks` bails rather than installing an opt-in that cannot take effect. `docs/install.md:555` already states it.

What is missing is where they looked. **The support matrix documents negative capabilities as carefully as positive ones** — the Codex row already carries *"No automatic true session-end hook"*, Pool's says *"handoff injection does not"*, Grok's says *"no hook handoff injection"*. Assistant capture is the one capability that appears only in the positive, in the Claude Code row. Someone comparing harnesses opens the Codex row, finds nothing, and concludes it is available.

## Scope

**Deliberately one line.** Every other unsupported harness has the same silence, so the consistent fix touches most rows of a table that is clearly hand-written with per-harness voice — I would rather not guess your wording across all of them in a drive-by PR. Say the word and I will extend it to the rest, or take whatever phrasing you prefer for this one.

No code change, no test impact.


akitaonrails:
Thanks — you independently spotted the same doc gap as #627, and your framing (the matrix reader looked at the row and found nothing) is exactly right. It's already fixed: the Codex support-matrix row now states assistant-final-turn capture is Claude-Code + native-platform only and is refused for other agents, landed on `main` in `4181e0b4` (via #627) and forward-merged to `release/2.1`. Your wording is a touch richer on the *why* (refuses rather than installs a no-op), but the merged line covers the same ground and I don't want to re-open a one-line docs change across both branches for a marginal reword. Closing as already-addressed — appreciated the careful diagnosis (and the #196 cross-reference).

Issue #617, opened 2026-09-03
bootstrap: a failed chunk discards every preceding chunk — partial progress or --resume


pepaulo8:
**Follow-up to #614** (as offered there, and per your 2.1 suggestion).

## Problem

`#[serde(default)]` on `pages` fixed the case ai-memory controls — a chunk that legitimately has nothing to add. It does not help when a chunk fails for a reason ai-memory cannot prevent: a provider 5xx, a rate limit, a dropped connection. Those still discard the entire run.

The reason is where the write happens. In `process_sources` (`crates/ai-memory-consolidate/src/bootstrap.rs`), every chunk is awaited with `?` inside the loop, pages accumulate in `pages_by_path`, and nothing reaches disk until `apply_batch` runs after the loop completes (~:456). So chunk *n* failing throws away chunks *1..n-1* as well.

The cost scales with the thing that makes chunking necessary in the first place. On a 9-chunk run at roughly 2-3 minutes per chunk, a failure on the last chunk costs eight successful LLM calls and ~20 minutes, and the retry starts from chunk 1 and pays for all of it again. During my own bootstrap of a docs-heavy repository I hit three distinct transient failures across four attempts (`provider error 520`, `Connection reset by peer`, and the missing-`pages` bug from #614) — the run never completed. Larger repositories are more exposed, because more chunks means more chances that at least one fails.

There is currently no partial state to recover from either: `bootstrap.md` is written in the same `apply_batch` as the pages, so a failed run leaves no manifest and no record of which chunks succeeded.

## Possible directions

You named two in #614; writing out what each implies, without assuming a preference:

**A. Write incrementally, per chunk.** After each successful chunk, persist that chunk's pages instead of buffering them. A failed run leaves the completed chunks' pages in the wiki.
- Smallest change; no new state to model.
- Changes bootstrap from one atomic write to N writes, so a failed run leaves a project partially bootstrapped with no manifest. `--force` re-runs would need to be well-defined against pre-existing pages (the `insert_bootstrap_page` first-write-wins rule is currently per-run, in memory).
- The per-chunk prompt already receives the paths earlier chunks wrote, so the dedup input survives naturally.

**B. Persist progress and add `--resume`.** Record completed chunks (and their pages) as durable state, in the spirit of #260's `session_consolidation_jobs` queue; `--resume` picks up at the first incomplete chunk.
- Keeps the wiki write atomic — nothing lands until the run truly completes.
[...]

PR #634, opened 2026-09-04
feat(bootstrap): --resume reuses the chunks an interrupted run completed

### LINKED ISSUE #621: bootstrap --resume: durable per-chunk progress to survive a crash/hard failure
Split from #617 (the transient-failure half is fixed there by bounded chunk retry, shipped in 2.0.3). This tracks the remaining, larger piece: durable resumability so a **crash, server restart, or a chunk that fails past the retry budget** does not discard the completed chunks' expensive LLM work.

## Design (grounded in the code)

**Fingerprint-guarded resume.** Chunks are positional-index only, and `collect_sources` re-reads git/docs at run time, so a naive index-resume is unsafe. Key the durable progress by a **fingerprint of the pruned source set + chunk budget**: identical inputs ⇒ deterministic chunk plan ⇒ index-skip is valid; changed inputs ⇒ fingerprint mismatch ⇒ fresh run.

**Storage.** A new `V60 bootstrap_chunk_progress(fingerprint, chunk_index, pages_json, rationale, created_at, PK(fingerprint, chunk_index))` table, mirroring #260's `session_consolidation_jobs` lease/state pattern (that table is session-specific — FK to `sessions`, PK `(session_id, generation)` — so bootstrap needs its own). Store methods: `record_bootstrap_chunk`, `load_bootstrap_progress`, `clear_bootstrap_progress`, plus a stale-prune.

**Flow.** Persist each chunk's pages right after it succeeds (`bootstrap.rs` ~L382, after the retry helper). On `--resume`: seed `pages_by_path` from the store and skip completed indices; clear on the final `apply_batch`. The wiki write stays **atomic** — nothing lands until the run completes.

**Wiring.** A `--resume` flag on `BootstrapArgs` threaded through the `/admin/bootstrap` POST body into `BootstrapConfig`; `Bootstrap` gains a store-writer handle (it currently holds only `wiki`/`reader`/`llm`).

## Open design question (carried from #617)
Is a partially-bootstrapped project (pages present, no manifest) an acceptable intermediate state, or must the manifest-gated `--force` guard hold as an invariant? The atomic-write design above preserves the invariant (nothing lands until completion), which is the intent.

@pepaulo8 offered to implement this and has a reproduction repo that reliably produces 9 chunks — very welcome. ~6-7 file feature (migration, store methods, bootstrap, CLI, admin HTTP, tests).

pepaulo8:
## What changed

`ai-memory bootstrap` now records each chunk's pages as it completes, and a new `--resume` flag reuses them instead of paying for those LLM calls again.

Before: a multi-chunk run held every page in `pages_by_path` until one `apply_batch` after the loop, so a crash, a server restart, or a chunk that failed past its retry budget discarded every chunk already paid for. On a nine-chunk run that is up to eight LLM calls and ~20 minutes, and the retry restarted at chunk 1.

After: each chunk's output is written to a new `bootstrap_chunk_progress` table as it lands, and cleared once the run writes its pages. **The wiki write stays atomic** — nothing reaches the wiki until the run completes, so the manifest-gated `--force` guard holds as an invariant. Without `--resume` the observable behaviour is unchanged.

No default changes. Progress is recorded on every run (a run cannot know it is about to fail), but it is only *read* under `--resume`.

## Why

Closes #621. Follows #617, where the bounded transient retry fixed the common case; this covers the rarer crash / restart / past-retry-budget one.

**Fingerprint-guarded, per your design.** Chunks are addressed by position and `collect_sources` re-reads git and `docs/` at run time, so a naive index-resume is unsafe. Progress is keyed by a SHA-256 of the pruned source set plus the chunk budget — the planner's exact inputs, hashed before `plan_bootstrap_chunks` consumes `kept` at `bootstrap.rs:401`. `plan_bootstrap_chunks` is pure in those two, so an equal fingerprint guarantees an identical chunk plan; a new commit, an edited doc, or a different `--chunk-input-tokens` changes the hash and starts a fresh run.

**Three corrections to the issue's design, from the code:**

- The migration is **`V59`**, not `V60` — `V58__close_retired_entity_windows.sql` is the highest, and `api_credentials.rs:414` pins it.
- The per-chunk persistence point is **`bootstrap.rs:429`** (after `insert_bootstrap_page`), not ~L382.
- `apply_batch` is **`Wiki::apply_batch`** (`wiki.rs:1531`), called from `bootstrap.rs:498` — that is where the clear lands.

**One simplification**, flagged in the issue and carried through: the table has no lease, state, or attempt columns. `session_consolidation_jobs` needs those because workers race for jobs; `/admin/bootstrap` already serialises runs behind `bootstrap_lock` (`admin.rs:1759`). Stored state is just `(fingerprint, chunk_index) → pages_json, rationale`. Say the word and I'll widen it to mirror the queue shape.

**Only the contiguous prefix is reused.** This one is not obvious and I got it wrong first: `insert_bootstrap_page` is first-write-wins in chunk order, so if chunk 5 failed while 6-9 were already recorded, seeding 6-9 before re-running 5 would let a later chunk keep a path chunk 5 owns — a resumed run would produce different pages than a clean one. `reusable_prefix` stops at the first gap, and at a row whose JSON no longer parses.

Failures around the optimisation are logged and swallowed, never fatal: a store error while recording or reading progress just means the chunk is paid for again, which is what would have happened anyway.

## Test plan

- [x] `cargo clippy -p ai-memory-store -p ai-memory-consolidate --all-targets -- -D warnings` passes (both crates this touches; it caught an `explicit_counter_loop` I've fixed)
- [x] `cargo test -p ai-memory-store -p ai-memory-consolidate --lib` — **537 passed, 0 failed** (350 + 187), all 13 new tests green. Not the full `--workspace`; see "Notes for reviewers"
- [x] `cargo test -p ai-memory-store --lib` — **350 passed**, including the 5 new `bootstrap_progress` tests and the bumped `schema_top_version_is_pinned`
- [x] 13 new tests: 5 store (ordering, fingerprint isolation, replace-on-rerecord, clear-all-fingerprints, exclusive prune cutoff), 7 fingerprint (stability, budget/content/ordering/kind sensitivity, length-prefix ambiguity), 6 `reusable_prefix` (complete run, stops at gap, missing first chunk, unreadable JSON, payload carried, empty)

## Commit attribution

## Release impact

- [x] **Minor** — additive: a new `--resume` flag, a new POST field, a new table

## CHANGELOG (merge gate)

- [x] I added a `CHANGELOG.md` `[Unreleased]` entry under `### Added`
- [x] No default changed — without `--resume`, behaviour is unchanged

## Notes for reviewers

**On the test scope.** Both crates this touches run green: 350 in `ai-memory-store`, 187 in `ai-memory-consolidate`. I ran the second set on an x86_64 host rather than my own machine — on aarch64 the test binary cannot link, because `gemm-f16` fails with `error: instruction requires: fullfp16` (pulled in through `candle` via `ai-memory-llm`'s default `local-embeddings` feature). That was the gap in #615; this time I went and found a machine where it links. I have not run the full `--workspace`, so the crates this PR does not touch are still CI's word.

**Validated end-to-end on the reproduction repo**, against a real Anthropic provider (`claude-haiku-4-5`), in a throwaway container built from this branch:

| behaviour | observed |
|---|---|
| progress survives a hard kill | `docker kill` (SIGKILL) during chunk 8/9 left chunks 1-7 recorded, ~226 KB of pages |
| the wiki write stays atomic | after that crash: no pages, no manifest — the `--force` guard still holds |
| a changed source set is refused | resuming from a differently-cloned checkout (69 vs 70 sources, 8 vs 9 chunks) logged `--resume found no reusable progress for these sources; starting from the first chunk` and paid for the run again, rather than resuming into a shifted plan |
| fingerprints stay isolated | two fingerprints' rows coexisted in the table without either adopting the other's chunks |
| **resume skips paid chunks** | after a kill during chunk 4/8: `--resume reusing completed chunks reused=3 total=8`, and the first LLM call of the run was `chunk=4` |
| **progress is cleared on success** | the completed run emptied the table — including the orphaned rows from the earlier abandoned fingerprint |
| the run actually completes | 58 pages + `bootstrap.md` written |

Worth noting: that repository had **never** completed a bootstrap across four attempts during my pilot (the #614 bug, a `520`, a dropped connection). It completed on the resume.

**One thing I'd like your read on:** progress is recorded unconditionally, so every bootstrap now writes N small rows even when it will never resume. The alternative is recording only under `--resume`, but that flag is chosen *before* anyone knows the run will fail, which makes it useless exactly when it matters. Clearing on success plus the age sweep should keep the table empty in the common case.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   snapshot.
 
 ### Added
+- `ai-memory bootstrap --resume` reuses the chunks an interrupted run already
+  completed instead of paying for their LLM calls again (#621). A multi-chunk
+  bootstrap holds every page in memory until one post-loop write, so a crash,
+  a server restart, or a chunk that fails past its retry budget discarded
+  every chunk paid for so far — on a nine-chunk run, up to eight LLM calls and
+  ~20 minutes, and the retry started from the first chunk. Each chunk's pages
+  are now recorded as they complete and cleared once the run writes, so the
+  wiki write stays atomic. Progress is keyed by a fingerprint of the pruned
+  source set plus the chunk budget: chunks are addressed by position and
+  sources are re-read from git and `docs/` at run time, so a new commit or a
+  different `--chunk-input-tokens` changes the hash and starts a fresh run
+  rather than resuming into a plan that has shifted underneath.
 - `ai-memory status` and the pre-migration backup log now report the data
   directory's **filesystem free space** (#629). The OKF-migration backup gate
   proves the archive is writable but said nothing about disk headroom — an
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -1144,6 +1144,12 @@ pub struct BootstrapArgs {
     /// the same project (the manifest is `wiki/bootstrap.md`).
     #[arg(long)]
     pub force: bool,
+    /// Reuse the chunks an interrupted run already completed instead of
+    /// paying for their LLM calls again. Progress is only reused when the
+    /// sources and chunk budget are unchanged; otherwise the run starts
+    /// over on its own.
+    #[arg(long)]
+    pub resume: bool,
 }
 
 /// Arguments for `setup-agent`.
--- a/crates/ai-memory-cli/src/commands/bootstrap.rs
+++ b/crates/ai-memory-cli/src/commands/bootstrap.rs
@@ -139,6 +139,7 @@ pub async fn run(config: &Config, args: BootstrapArgs) -> Result<()> {
         "chunk_input_tokens": args.chunk_input_tokens,
         "dry_run": args.dry_run,
         "force": args.force,
+        "resume": args.resume,
     });
     let outcome: BootstrapOutcome = post_json(&ep, "/admin/bootstrap", &body).await?;
 
--- a/crates/ai-memory-consolidate/src/bootstrap.rs
+++ b/crates/ai-memory-consolidate/src/bootstrap.rs
@@ -40,6 +40,7 @@ use ai_memory_wiki::{Wiki, WritePageRequest};
 use jiff::Timestamp;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 use thiserror::Error;
 use tracing::{debug, info, warn};
 
@@ -277,6 +278,11 @@ pub struct BootstrapConfig {
     pub dry_run: bool,
     /// Allow re-running even when `wiki/bootstrap.md` already exists.
     pub force: bool,
+    /// Reuse the chunks a previous interrupted run already completed, instead
+    /// of paying for their LLM calls again. Only progress recorded under the
+    /// same source/budget fingerprint is eligible; anything else is ignored and
+    /// the run starts from the first chunk.
+    pub resume: bool,
 }
 
 /// Bootstrap driver. Holds the LLM provider + wiki handle + reader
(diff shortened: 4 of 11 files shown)


pepaulo8:
Closing — you landed this in `08486813` before I finished, and re-litigating a merged design would just cost you a review. No complaints; the two implementations converged on the same shape (same SHA-256 fingerprint over the pruned source set + chunk budget, same atomic wiki write, same tolerance for corrupt rows), which is a good sign the design in #621 was the right one.

One thing from mine is worth carrying over, and I'd rather hand it to you than let it sit in a closed PR. I'll open it as a separate issue with a test, but the short version:

**`release/2.1` adopts every recorded chunk, not just the contiguous prefix from chunk 1.** At `bootstrap.rs:464-485` the loop seeds `pages_by_path` from each row in turn and inserts the index into `completed`; a row that fails to parse hits `continue` and the ones after it are still adopted.

That is safe only while the recorded chunks are a prefix. They aren't, in the case this feature exists for: a run that dies at chunk 5 with chunks 6-9 already recorded (transient failure past the retry budget, or a `docker kill`) leaves a gap. On resume, chunks 6-9 seed *before* chunk 5 is re-run — and since `insert_bootstrap_page` (`:1400`) is first-write-wins by insertion order, chunk 6 keeps any path chunk 5 owns. The run completes, writes, and clears progress. No error, no warning beyond the usual duplicate-path `warn!` — but the resumed wiki differs from what a clean run would have produced, which is the one property a resume has to preserve.

I hit this in my own implementation first and only caught it writing the test for it. The fix is small: stop at the first gap (and at the first unparseable row, for the same reason) rather than `continue`.

Happy to send that as a focused patch against `release/2.1` if useful — it's ~15 lines plus tests, and it does not touch anything else you shipped.

pepaulo8:
For the record on the overlap, since it was my miss and not a race: I commented on #621 saying I was taking it, then worked heads-down and opened the PR without re-checking the issue — which had already been closed by `08486813` about 45 minutes earlier. Had I refreshed it before pushing I would have closed the branch instead of asking you to look at a duplicate. Sorry for the noise on your queue.

The finding above stands on its own regardless, and I'd still like to send it as a small patch against `release/2.1` if you want it.

PR #637, opened 2026-09-04
fix(wrapper): support Podman without Docker

### LINKED ISSUE #636: Support for podman
**Problem this would solve**
For those who do not run docker, podman is the main alternative. Adding a support to podman would be ideal.

**Proposed solution**
add support to `podman` during installation

**Alternatives considered**
None

**Would this break existing installs?**
I don't see any break changes here

**Additional context**

jpramos123:
## Summary
- auto-select Podman when Docker is unavailable while preserving `AI_MEMORY_DOCKER` overrides
- use a fully qualified Docker Hub image name for non-interactive Podman resolution
- preserve the selected engine in upgrade recovery commands and document the Podman workflow

Closes #636

## Testing
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`
- `cargo deny check`
- `bash -n bin/ai-memory`
- `bash -n tests/wrapper_upgrade.sh`
- `tests/wrapper_upgrade.sh`
- `AI_MEMORY_NO_VERSION_CHECK=1 AI_MEMORY_NO_TTY=1 bin/ai-memory --version` (real rootless Podman; `ai-memory 2.0.2`)
- `git diff --check upstream/main...HEAD`
- repository global pre-commit gate

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `dest_free_bytes`, so the same number is visible right before a migration.
 
 ### Fixed
+- The `bin/ai-memory` container wrapper now automatically uses Podman when
+  Docker is unavailable, uses a fully qualified Docker Hub image name to avoid
+  Podman's non-interactive short-name resolution failure, and preserves the
+  selected engine in standalone-container recovery scripts emitted by
+  `ai-memory upgrade` instead of hard-coding `docker` (#636).
 - Corrected `docs/MIGRATION-2.0.md`, which wrongly promised the automatic
   pre-migration archive could roll a store back to 1.x (#633). The archive is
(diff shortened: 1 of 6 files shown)


akitaonrails:
Reconciled onto `release/2.1` as merge `614a8bda` (your commit preserved), queued for 2.1.0. Podman support is a new capability, so it rides the 2.1 feature train rather than a 2.0.x patch.

Clean change: the engine auto-selection (`AI_MEMORY_DOCKER` override → `docker` → `podman`, with a docker fallback that preserves the existing command-not-found error) and the fully-qualified `docker.io/…` image default (required for Podman's non-interactive short-name resolution, tolerated by Docker) are exactly right, and `%q`-quoting the engine in the reconstructed `ai-memory upgrade` recovery script is a nice touch. Packaging tests (including your new Podman coverage) pass and the wrapper is shell-syntax-clean. Closing as resolved-and-ready-to-release on the 2.1 branch. Thanks!

Issue #372, opened 2026-08-05
memory_write_page writes to wrong project when another project has an unclosed session (ended_at = NULL)


Huang1024:
## Description

When `memory_write_page` is called without an explicit `project` parameter, the page can be written to the wrong project if another project has a session with `ended_at = NULL` (an unclosed/zombie session).

## Steps to reproduce

1. Open Project A in Claude Code — this creates a session with normal lifecycle hooks
2. Open Project B in Claude Code — this creates another session
3. Close Project B's Claude Code session WITHOUT properly ending it (e.g., terminal killed, process crash) — the session row keeps `ended_at = NULL`
4. In Project A, work normally and call `memory_write_page` (without explicit `project` param) to persist knowledge discussed in Project A
5. The page is written to **Project B** instead of Project A

## What I observed

- Two `memory_write_page` calls made from two different project sessions (Project A, Project C) both landed in Project B
- Project B had exactly one session with `ended_at = NULL` and `ended_observation_count = 0`
- The other projects' sessions had proper `ended_at` timestamps
- Confirmed via `audit_log`: the `create_page` entries show `project_id` pointing to Project B
- The session content in Project A/C confirms the topics were discussed there, not in Project B

## Expected behavior

`memory_write_page` (without explicit `project`) should resolve to the project of the **current active session**, not to a project with a stale unclosed session.

## Environment

- ai-memory: latest
- Claude Code: latest
- OS: macOS arm64

## Database evidence

The sessions table shows:
- Project B session: `ended_at = NULL`, `ended_observation_count = 0`
- Project A and C sessions: proper `ended_at` values

Pages created during Project A/C sessions were attributed to Project B's `project_id` in both `pages` and `audit_log` tables.

## Possible cause

The project resolution logic for `memory_write_page` (the "recent hook activity" path) may be giving higher priority to sessions with `ended_at = NULL`, or failing to filter them out, causing a stale session from another project to "win" the project resolution.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

samirhvbr:
Reproduced on `main` (v1.24.0, `146c8a1`) against a live server. The trigger is narrower than "an unclosed session wins", and it explains why two different projects' writes both landed in B.

**Mechanism: any hook event republishes the process-wide active-project slot, with no recency or liveness guard.**

[...]

PR #638, opened 2026-09-04
Fix static MCP project routing guidance

arthurgimenes:
## Summary
- distinguish session-aware clients from static MCP clients in the canonical routing snippet and MCP handshake
- require static clients to pass exact `workspace` + `project` values on project-scoped calls while preserving global-mode exceptions
- align all project-scoped managed Agent Skills and tool schemas, with regression coverage and updated docs/changelog

## Tests
- `git diff --check`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-core routing`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-mcp snippet`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-mcp memory_install_self_routing_response_includes_managed_skills_and_targets`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-mcp project_scoped_tool_schemas_expose_the_static_client_contract`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-mcp routing_prompt_surfaces_share_the_client_aware_scope_contract`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-mcp prompts_warn_static_mcp_clients_need_explicit_scope`
- `TAILWIND_SKIP=1 cargo test -p ai-memory-cli --test routing_instructions --test routing_skills`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,19 +4,23 @@
 This project uses [ai-memory](https://github.com/akitaonrails/ai-memory)
 for cross-session continuity.
 
-**Default to the current project - always.** Every ai-memory tool
-auto-scopes to the project resolved from your session's working
-directory. **Do NOT pass `project`, `workspace`, or `cwd` arguments unless
-the user explicitly references a *different* project by name** (e.g. "what
-did we decide in the `other-app` project?"). Phrases like "this project",
-"here", "we", "our work", and "where did we leave off" all mean the
-*current* project, so call tools with no scoping args.
+**Choose project scope from the MCP client's identity support.**
 
-This default assumes the MCP client can identify the current agent
-session. Static MCP clients in parallel sessions for the same user cannot
-forward the real agent session id automatically; pass explicit
-`workspace` + `project` / `scopes`, or use a session-aware bridge that
-forwards the lifecycle-hook session id on MCP calls.
+- **Session-aware MCP clients** that forward the real lifecycle-hook session id
+  on every request should use automatic current-project routing. Omit `workspace`,
+  `project`, and `cwd` for the current repository; pass explicit scope only when
+  the user names a different project.
+- **Static MCP clients** (including clients with lifecycle hooks but no bridge
+  connecting that hook session id to MCP requests) must pass `workspace` and
+  `project` together on every project-scoped call, including requests about "this
+  project", "here", or "our work". Read the exact names from the nearest
+  `.ai-memory.toml` when it declares both. If it does not, obtain the names from
+  the operator or server configuration; never guess them from a directory name
+  and never rely on the server's last active project.
+
+This rule applies only to project-scoped calls. For cross-project retrieval,
+`global=true` must omit `workspace`, `project`, and `scopes`. For a standing
+preference written with `scope: "global"`, omit `workspace` and `project`.
 
 **Lifecycle hooks already capture sanitized, bounded prompt and tool-lifecycle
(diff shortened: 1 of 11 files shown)


abhisheksharma2411 (commented):
The distinction this draws is the right one, and the AGENTS.md rewrite reads much better than what it replaces — "default to the current project, always" plus a caveat paragraph was two rules where the reader needed one decision. Making it a fork on client capability is clearer, and pulling the `global=true` / `scope: "global"` exceptions up next to it closes the obvious follow-up question.

Ran the two gates your list doesn't mention, since CONTRIBUTING asks for all four:

- `cargo clippy -p ai-memory-core -p ai-memory-mcp --all-targets -- -D warnings` — clean
- `cargo test -p ai-memory-core routing` — 14 passed

(I skipped the full `cargo test --workspace` for a boring local reason: my disk filled. The change is prose in `SNIPPET_BODY`, the `SKILL.md` files and the MCP tool descriptions — `server.rs` is +221 lines of tool-description text, not logic — so I don't think a workspace run holds surprises. Worth someone running it anyway before merge.)

The tests are where I'd push back, and both problems point the same way — they're strict about wording that doesn't matter and blind to the contradiction they exist to catch.

## The positive assertions pin line wrapping

```rust
assert!(SNIPPET_BODY.contains("nearest\n  `.ai-memory.toml`"));
```

That's asserting a newline and a two-space indent. I reflowed one paragraph — **same words, same meaning**, just a different break point:

```
-  project", "here", or "our work". Read the exact names from the nearest
-  `.ai-memory.toml` when it declares both. If it does not, obtain the names from
+  project", "here", or "our work". Read the exact names from
+  the nearest `.ai-memory.toml` when it declares both. If it does not, obtain
```

```
test snippet_distinguishes_session_aware_and_static_scope_routing ... FAILED
```

The split between `contains("must pass \`workspace\` and")` and `contains("\`project\` together on every project-scoped call")` looks like it's already working around this — that phrase is one sentence, cut in two because the wrap falls in the middle. Normalising whitespace before matching would let you assert the sentence as a sentence:

```rust
let body: String = SNIPPET_BODY.split_whitespace().collect::<Vec<_>>().join(" ");
assert!(body.contains("must pass `workspace` and `project` together on every project-scoped call"));
```

This matters more than usual here because the thing under test is prose. Someone will reword it, the test will fail for a reason unrelated to the contract, and the cheapest fix will look like deleting the assertion.

## The negative assertions only catch three exact phrasings

`project_scoped_skills_share_the_static_client_contract` asserts the absence of three literal strings. I appended this to `ai-memory-retrieval/SKILL.md`:

> Do not supply `workspace` or `project` for the current repository; the server already knows which project you are in.

which is precisely the guidance the list exists to forbid, in fresh words:

```
test project_scoped_skills_share_the_static_client_contract ... ok
test result: ok. 14 passed; 0 failed
```

I don't think there's a good string-matching answer to "no contradictory guidance anywhere", and I'd rather say so than suggest a longer blocklist — the next phrasing escapes that too. The positive assertions are the load-bearing half; the negative list is a tripwire for the three phrasings you just removed, which is a fine thing to have as long as it isn't read as a guard against the general case. A one-line comment saying that would stop the next person trusting it further than it goes.

Neither is a correctness problem and I wouldn't hold the PR for either — the guidance change itself is good, and the CHANGELOG entry under `### Fixed` with the `(#372)` reference is right where it should be.

PR #672, opened 2026-09-07
feat(retrieval): opt-in session-recall routing and L0 abstract vector stream

evanmaranzano:
## What

Two independent, **default-off** ranking signals for `memory_query`, both borrowed from OpenViking's context-database design and both visible per hit in `memory_query(explain=true)`:

**1. `[retrieval] query_intent` — session-recall routing (`session_recall_bonus`, default 0.25)**

Session pages carry a bounded authority penalty (kind `session` −0.15, tier `episodic` −0.08 → ×0.77) because they are rarely the right answer to a fact query. But a query that is *about* a past session ("上次排查 Caddy 的会话完成了吗", "what did we decide last time") needs one. A zero-LLM lexical router (CJK substring + Latin word-boundary markers) recognises that phrasing and hands session pages the penalty back plus `session_recall_bonus`, still clamped inside the authority bounds every other page lives in. Unrouted queries keep their exact baseline ordering. Report: `intent: "session_recall"`, `intent_boost`.

**2. `[retrieval] abstract_vectors` — L0 abstract-embedding stream**

New migration `V61__page_abstract_embeddings` (schema pin bumped 60 → 61): a page whose frontmatter carries `abstract:` gets that one-line summary embedded by the same backfill pass as its body (`run_embedding_backfill`, new `abstracts_embedded` counter), and `hybrid_search` gains a fifth RRF stream over it. Rationale: a one-line L0 summary embeds far more sharply than a multi-thousand-character body, so the vector side gets a precise second vote per page. Contributes nothing while the table is empty. Report: `abstract_rank`, `abstract_cosine`, `rrf.abstract`.

## Why (measured, not vibes)

138-query golden set (frozen before evaluation: 26 calibrated keyword queries + LLM-generated NL/keyword/session-recall/temporal-update queries, expected pages validated against the store) over a production two-year wiki, four-stream baseline (FTS5 + entity + vector + graph, mis-tei Qwen3-Embedding-8B), every run on a restored store snapshot:

| config | hit@1 | MRR | NDCG@10 | recall@10 |
|---|---|---|---|---|
| defaults off (= today) | 0.609 | 0.738 | 0.782 | 0.920 |
| abstract only | 0.681 | 0.780 | 0.815 | 0.935 |
| **both (shipped defaults)** | **0.746** | **0.845** | **0.879** | **0.986** |

Paired bootstrap (defaults vs both), 2000 resamples: every metric's 95% CI is above zero — hit@1 Δ+0.138 [+0.072, +0.203] wins/losses 22/3; NDCG@10 Δ+0.097 [+0.051, +0.141] 36/7.

Per-category hit@1 (both enabled): fact-keyword 0.682→0.773, fact-NL 0.591→0.705, session-recall **0.143→0.714**, temporal-update 0.600→0.700, v1-calibrated 0.769→0.808. Unrouted categories keep their exact baseline ordering (Δ exactly 0.000) because the router simply does not fire on them.

Defaults-off is byte-identical to `main` on the same snapshot — verified, not assumed.

**Honest trade-off.** On a 24-query adversarial set (fact queries artificially prefixed with "之前/上次", expected page unchanged) the combined feature costs hit@1 0.542→0.417, NDCG@10 0.768→0.662, while recall@10 stays 0.917 — the right page drops a few ranks rather than disappearing. Lexical routing cannot separate "上次说的X" (fact) from "上次那次会话" (recall); that needs an LLM gate. This is why both signals ship **off by default**: operators with session-recall traffic turn them on knowing this number. Lowering `session_recall_bonus` to 0.15 halves the adversarial cost at a small golden-set price (NDCG@10 0.871).

Also measured and **not** proposed: OpenViking-style hotness weighting (`sigmoid(ln(1+access_count)) × exp(−ln2·age/7d)`), which was monotonically negative on the same golden set (NDCG@10 0.782 → 0.768/0.739/0.713 across α ∈ {0.3,0.3,0.6} × half-life ∈ {7,30,30}), and a recency boost under a detected "current state" intent (−1 hit@1 on its own target category). Neither survives an honest A/B, so neither is in this PR.

## Tests

- `retrieval_tuning::tests` (store, unit): marker detection (CJK/Latin/word-boundary/会话-as-topic), default inertness.
- `suite::retrieval_tuning_streams` (store, integration): routing flips a session page above an evenly-ranked semantic page only for routed queries; the abstract stream flips ordering only when enabled, degrades to nothing for pages without abstract rows; `explain` carries `intent`/`intent_boost`.
- Existing suites pass; `schema_top_version_is_pinned` bumped to 61 as part of adding V61. One pre-existing flaky timing test (`auto_improve::tests::eval_error_timeout_and_invalid_json_fail_closed`) fails under parallel load on this Windows machine on unmodified `main` too — unrelated.

## Docs

`docs/ARCHITECTURE.md` configuration block gains the `[retrieval]` section; CHANGELOG entry under Unreleased/Added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `[retrieval]` section with two opt-in ranking signals, both off by default
+  so unconfigured stores rank exactly as before. `query_intent` routes
+  lexically session-recall queries ("上次 / 之前那次…的会话 / last time /
+  yesterday …") past the default session-page authority penalty
+  (×0.77 combined kind/tier), with `session_recall_bonus` (default 0.25)
+  sizing the lift; `abstract_vectors` adds a fifth RRF stream over the new
+  `page_abstract_embeddings` table (V61) — the L0 layer: a page's
+  frontmatter `abstract:` line is embedded by the same backfill as its
+  body, and a one-line summary embeds far more sharply than a
+  multi-thousand-character body. Measured on a 138-query golden set over a
+  production two-year wiki (FTS5 + entity + vector + graph, mis-tei
+  Qwen3-Embedding-8B): hit@1 0.609 → 0.746 (+22%), NDCG@10 0.782 → 0.879
+  (+12%) with both enabled; every unrouted category keeps its exact
+  baseline ordering, and both signals are visible per hit in
+  `memory_query(explain=true)` (`intent`, `intent_boost`,
+  `abstract_rank`, `rrf.abstract`).
+
 ### Fixed
 - OKF-conformed event ledgers are skipped by the indexer again, so a migrated
   store stops growing without bound. The reserved-file check treated any
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,13 +15,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (×0.77 combined kind/tier), with `session_recall_bonus` (default 0.25)
   sizing the lift; `abstract_vectors` adds a fifth RRF stream over the new
   `page_abstract_embeddings` table (V61) — the L0 layer: a page's
-  frontmatter `abstract:` line is embedded by the same backfill as its
-  body, and a one-line summary embeds far more sharply than a
-  multi-thousand-character body. Measured on a 138-query golden set over a
+  frontmatter `abstract:` line is embedded on its own, at write time when an
+  embedder is attached and by the same backfill as the body otherwise
+  (a one-line summary embeds far more sharply than a
+  multi-thousand-character body). Measured on a 138-query golden set over a
   production two-year wiki (FTS5 + entity + vector + graph, mis-tei
   Qwen3-Embedding-8B): hit@1 0.609 → 0.746 (+22%), NDCG@10 0.782 → 0.879
   (+12%) with both enabled; every unrouted category keeps its exact
-  baseline ordering, and both signals are visible per hit in
+  baseline ordering. Cross-checked on an independently built 99-query
+  golden set against the same store: hit@1 0.556 → 0.626, NDCG@10
+  0.712 → 0.779 (+9.4%). Both signals are visible per hit in
   `memory_query(explain=true)` (`intent`, `intent_boost`,
   `abstract_rank`, `rrf.abstract`).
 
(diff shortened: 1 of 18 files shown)


evanmaranzano:
## Cross-check on the original 99-query golden set

Re-ran the A/B against the independently built company-machine golden set (the 102-query `golden_queries_v2.jsonl`: 99 ranked queries after its own scoring rules exclude the three empty-expected abstain probes; the five `_global:*` expectations queried against the `_global` scope), same snapshot-restore protocol, same build:

| | hit@1 | MRR | NDCG@10 | recall@10 |
|---|---|---|---|---|
| defaults off | 0.556 | 0.670 | 0.712 | 0.920 |
| both on | **0.626** | **0.742** | **0.779** | **0.942** |

(+12.6% hit@1, +9.4% NDCG@10 — consistent with the primary eval's +22.6% / +12.4%.)

Per-category hit@1 movement: aimem-ops 0.111 → 0.278, server-project 0.474 → 0.632, typology-multipage 0.333 → 0.667, gotcha 0.909 → 0.955, typology-update 1.00 → 1.00, session-recall 0.500 → 0.500.

One honest per-category regression, attribution included: the 5 `hard-control` probes (deliberately ambiguous recalls where a session log and a curated page compete) drop hit@1 0.800 → 0.400 / NDCG@10 0.646 → 0.593. None of those five queries carry session-recall markers, so the router is not involved — the abstract stream elevates curated pages (e.g. `local-memory/project_jszx15_newapi_gateway.md` above a session log for "15服务器 Caddy 站点配置检查"). Whether a curated page outranking a session log there is a regression or a fix is genuinely arguable; operators who side with the session logs can leave `abstract_vectors` off.

This push also adds the two integration tests that pin the abstract embedding paths (`suite::abstract_backfill`, `write_page_embeds_frontmatter_abstract_and_cleans_stale_rows`), and gives `write_page` write-time abstract embedding — a page rewritten without its frontmatter `abstract:` now has its latest abstract row removed, mirroring the body row's replace-on-write semantics.

akitaonrails:
Merged onto `release/2.2` (`d42ab97a`) for 2.2.0 — retargeted from `main` (features ride the 2.2 train now). Clean, additive, and opt-in: both signals default off so unconfigured stores rank exactly as before; V61 is additive and linear (pin bumped to 61); scope isolation preserved on every new query. I added the missing `(#672)` CHANGELOG ref. GitHub can't auto-close since it landed on a non-base branch. Excellent evidence (and honest negative-result reporting) — thanks @evanmaranzano.

PR #673, opened 2026-09-08
fix(wrapper): match multi-arch image digest and preserve volume mount modes

GrYllO:
## Summary
- **Multi-arch image digest resolution**: In `bin/ai-memory`, `remote_digest()` previously ran `grep -oE 'sha256:[a-f0-9]{64}' | head -n 1` on `manifest inspect` output. Because Docker Hub publishes `ai-memory` as a multi-arch manifest list with `linux/arm64` listed first, the extracted remote digest was always `arm64`. On `x86_64` hosts and under Podman (where `.RepoDigests 0` is the manifest list index digest), comparing local digests against the remote digest always failed, producing a false-positive `ai-memory: a newer image is available on Docker Hub` warning on every version check. `remote_digest()` now targets the host platform architecture (`amd64` vs `arm64`) and checks against all locally known repo digests.
- **Preserve volume mount modes**: `emit_docker_run_script` now includes `{{if .Mode}}:{{.Mode}}{{end}}`, preserving `:Z` / `:z` / `:ro` flags on inspected container mounts so reconstructed containers on SELinux / rootless Podman systems retain required permissions on the `/data` volume.
- **Filter transient runtime env vars**: In `emit_docker_run_script`, transient container runtime variables (`HOSTNAME`, `container=podman`) injected by engines are filtered out of the reconstructed run flags.

## Testing
- `bash -n bin/ai-memory`
- `bash -n tests/wrapper_upgrade.sh`
- `./tests/wrapper_upgrade.sh` (all tests pass, including standalone `:Z` mode preservation, env var filtering, and automated pty test cases for matching and outdated digests across amd64 and arm64)
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   summary or handoff was produced. The handlers are now `async` and await a
   bounded (2s) drain of the pending request(s) before returning, mirroring
   the joinable dispose-drain already used by the OpenCode integration (#676).
+- The `bin/ai-memory` container wrapper now matches multi-architecture image
+  manifests against the host's platform architecture, eliminating a
+  false-positive "a newer image is available on Docker Hub" warning on x86_64
+  and Podman. Additionally, `emit_docker_run_script` now preserves volume mount
+  modes (such as `:Z` on SELinux/Podman environments) and filters transient
+  runtime environment variables (`HOSTNAME`, `container=podman`).
 
 ## [2.1.1] - 2026-09-07
 
--- a/bin/ai-memory
+++ b/bin/ai-memory
@@ -60,16 +60,47 @@ WRAPPER_SHA256_URL="${AI_MEMORY_WRAPPER_SHA256_URL:-${WRAPPER_URL}.sha256}"
 
 # ---- version-check helpers (best-effort; never block the wrapper) --------
 
-local_digest() {
-  "${DOCKER}" image inspect --format='{{index .RepoDigests 0}}' "${IMAGE}" 2>/dev/null \
-    | sed 's/.*@//' || true
+local_digests() {
+  "${DOCKER}" image inspect --format='{{range .RepoDigests}}{{println .}}{{end}}{{println .Digest}}' "${IMAGE}" 2>/dev/null \
+    | sed 's/.*@//' | grep -v '^$' || true
 }
 
 remote_digest() {
+  local host_arch
+  case "$(uname -m 2>/dev/null || true)" in
+    x86_64 | amd64) host_arch="amd64" ;;
+    aarch64 | arm64) host_arch="arm64" ;;
+    *) host_arch="" ;;
+  esac
+
   # docker manifest inspect was experimental pre-20.10; on modern
-  # docker it's stable. Silent on failure (offline, podman, etc.).
-  "${DOCKER}" manifest inspect "${IMAGE}" 2>/dev/null \
-    | grep -oE 'sha256:[a-f0-9]{64}' | head -n 1 || true
+  # docker and podman it's stable. Silent on failure (offline, etc.).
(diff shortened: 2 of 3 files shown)


akitaonrails:
Thanks @GrYllO — both bugs you identified are real, the shell is clean (no injection, no privilege/scope change, quoting is sound, engine-selection and the fully-qualified image from #636/#637 are untouched), and the env-var filtering (`HOSTNAME`/`container=`) is correct and properly tested. But the headline digest fix needs rework before it can land, because it breaks on **Docker — the default engine**:

**1. `.Digest` doesn't exist on Docker → the version check is silently disabled.** `local_digests()` uses `--format='…{{println .Digest}}'`. Docker has no top-level `.Digest` field; the template errors and Docker discards the **whole** output, so `local_digests()` returns empty, `maybe_warn_outdated` hits its `[ -n "${local_d}" ] || return 0` guard, and **no upgrade notification ever fires on Docker** — true ones included. It works on Podman (which does expose `.Digest`). So the false positive is 'fixed' on Docker only by silencing all upgrade checks there.

**2. Manifest-list vs per-arch digest still won't match on Docker.** Even past the crash, Docker's classic image store puts the **manifest-list** digest in `.RepoDigests`, while your new `remote_digest` returns the **per-arch child** digest — they can't be equal, so the false positive persists on that store. The fix is effectively Podman-correct, Docker-fragile.

**3. The mount-mode fix (`{{if .Mode}}`) is correct but untested.** `tests/wrapper_upgrade.sh`'s fake engine returns a pre-formatted `-v …:Z`, bypassing the real template — so the `:Z`/`:ro` assertion passes even against the *base* wrapper. The digest tests likewise return canned values regardless of the `--format` string, so the Docker `.Digest` crash is invisible to CI.

Suggested rework:
- Split the inspect so a missing `.Digest` on Docker doesn't nuke `.RepoDigests` (two separate `image inspect` calls, tolerate the empty one).
- Reconcile Docker's manifest-list `RepoDigests` against the per-arch remote digest (accept the list digest too, or gate per-arch matching to engines/stores that expose a per-arch local digest) so the false positive is actually fixed on Docker, not silenced.
- Add a test that drives the real `{{if .Mode}}` template and the real `image inspect` format (a fake engine that honors `--format`, or an integration case), so the mount-mode fix and the `.Digest` behavior are covered.

The mount-mode and env-filter parts are good and worth keeping. This targets `main` for 2.1.2 once the digest logic is Docker-correct. Happy to re-review.

GrYllO:
Thanks for the thorough review and catch, @akitaonrails! My local development environment had `/usr/bin/docker` aliased to Podman under CLI emulation, which completely masked Docker's template parser behavior and image store structure during testing.

I've pushed an update addressing all three points:

1. **Split Local Inspect**: `local_digests()` now runs two independent inspect calls: `local_repo_digests` (`--format='{{range .RepoDigests}}{{println .}}{{end}}'`) and `local_instance_digest` (`--format='{{println .Digest}}'`). If Docker errors on `.Digest`, it is tolerated silently without discarding `.RepoDigests`.
2. **Reconciled Manifest-List vs. Per-Arch Digests**:
   - `remote_list_digest` queries `${DOCKER} buildx imagetools inspect "${IMAGE}"` to retrieve the manifest list / index digest, matching Docker's `.RepoDigests`.
   - `remote_arch_digest` retrieves the architecture-matched child digest via `${DOCKER} manifest inspect`, matching Podman's `.Digest`.
   - Gated per-arch matching: if running on classic Docker without buildx (no remote list digest and no local instance digest), the check is safely bypassed to avoid apples-to-oranges false positives.
3. **Faithful Test Coverage**:
   - The fake docker in `tests/wrapper_upgrade.sh` now inspects the mount template for `if .Mode` before returning `:Z`.
   - `run_version_check_case` now tests real `--format` strings. When simulating Docker, it rejects `.Digest` with Docker's exact template parsing error and returns only the manifest-list digest in `.RepoDigests`.
   - The test suite covers both Docker (with buildx & classic) and Podman across `amd64` and `arm64` for both matching and outdated states.

PR #679, opened 2026-09-08
fix(cli): name the request method and path in server error messages

jaysonsantos:
## What changed

- A non-2xx response from the configured server printed only
  `server returned 404 Not Found: <body>`. The message named no endpoint. A
  failed CLI subcommand gave no clue which request failed.
- `ServerResponseError` now carries the request method and the URL path. The
  error prints `GET /admin/open-sessions: server returned 404 Not Found:
  <body>`.
- The error keeps the path only. A token in the URL userinfo or query string
  never reaches a log line. A test asserts this.
- The six subcommand helpers repeated the same status check. One
  `require_success` helper replaces them.

No default changes. The only observable change is the text of the error.

## Why

A CLI subcommand runs several requests. `finalize_session` and `run` each call
three or more endpoints. When one returned a non-2xx status, the message named
the status and the body but not the request. A user had to guess the endpoint,
or read the source.

The path is enough to identify the request. The full URL is not safe to print,
because `ServerEndpoint` can hold a bearer token, and a caller can put
credentials in the URL userinfo or the query string. `require_success` reads
the path from `resp.url()`, so no call site can pass a raw URL by mistake.

`run.rs` and `finalize_session.rs` downcast to `ServerResponseError` to read
`status()` and `body()`. Both keep the same behaviour.

## Test plan

- [x] `cargo tf` (or `cargo test --workspace --all-targets`) passes
- [x] Manual test: added
      `http_client::tests::response_errors_include_method_and_credential_free_path`.
      It builds the error from
      `https://username:password@example.test/workstream/runs?access_token=secret`
      and asserts the message is
      `POST /workstream/runs: server returned 404 Not Found: missing`. It also
      asserts the message holds none of `username`, `password`,
      `access_token`, `secret`, and that the error still downcasts to
      `ServerResponseError` with the original status and body.

## Commit attribution

## Release impact

      new agent harness or provider

## CHANGELOG (merge gate)

## Notes for reviewers

**Question: can `flake.lock` be committed?** `flake.nix` is tracked, but
`flake.lock` is not, and `.gitignore` does not list it. An untracked lock file
means each contributor resolves a different toolchain from the same
`flake.nix`. I kept `flake.lock` out of this PR, because the choice belongs to
the maintainers and not to this fix. Tell me the answer and I open a separate
PR:

- Commit it, so `nix develop` gives every contributor and CI the same
  toolchain.
- Add it to `.gitignore`, so the current state is deliberate and `git status`
  stays clean.

Two design points to check:

1. `require_success` takes the method as an argument. `reqwest::Response` does
   not carry the request method, so the call site supplies it. A wrong method
   at a call site produces a wrong label, and no test catches it.
2. The error holds `url.path()`. It drops the query string. A route that
   depends on a query parameter is harder to identify from the message alone.
   I chose this because the query string can hold a token.

https://claude.ai/code/session_01Uuoo7H5ZdDTZBAsGkE5VJS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Fixed
+- CLI server errors now name the request. A non-2xx response from the
+  configured server printed only `server returned 404 Not Found: <body>`,
+  so a failure gave no clue which endpoint answered. The message now leads
+  with the request method and path: `GET /admin/open-sessions: server
+  returned 404 Not Found: <body>`. The error keeps the path only, so a
+  token in the URL userinfo or query string never reaches a log line.
 - Wiki auto-commits stage what the wiki wrote instead of walking the
   whole tree, keep the repository open between commits, and no longer
   drop the commit when another session is writing a file at the same
--- a/crates/ai-memory-cli/src/http_client.rs
+++ b/crates/ai-memory-cli/src/http_client.rs
@@ -26,6 +26,8 @@ use ai_memory_web::normalize_prefix;
 /// Non-success response returned by the configured ai-memory server.
 #[derive(Debug)]
 pub(crate) struct ServerResponseError {
+    method: reqwest::Method,
+    path: String,
     status: reqwest::StatusCode,
     body: String,
 }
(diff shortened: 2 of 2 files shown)


jaysonsantos:
### What the error actually looks like

Captured by running the real binary against a stub server that answers every request with `404 {"error":"not found"}`, on this branch and on its parent (`195be4e5`).

**Before**

```
$ ai-memory status
Error: server returned 404 Not Found: {"error":"not found"}

$ ai-memory search foo
Error: server returned 404 Not Found: {"error":"not found"}

$ ai-memory compact --confirm
Error: server returned 404 Not Found: {"error":"not found"}
```

Three different subcommands, three different endpoints, byte-identical output. Nothing in the message tells you which request failed.

**After**

```
$ ai-memory status
Error: GET /admin/status: server returned 404 Not Found: {"error":"not found"}

$ ai-memory search foo
Error: GET /admin/search: server returned 404 Not Found: {"error":"not found"}

$ ai-memory compact --confirm
Error: POST /admin/compact: server returned 404 Not Found: {"error":"not found"}
```

### Where it matters most

Some callers already wrap the failure in an `anyhow` context layer. That layer named the operation but never the request, so the endpoint was still missing:

```
# before
$ ai-memory write-page --path notes/x.md --title X --body hi
Error: writing page via server

Caused by:
    server returned 404 Not Found: {"error":"not found"}

# after
$ ai-memory write-page --path notes/x.md --title X --body hi
Error: writing page via server

Caused by:
    POST /admin/write-page: server returned 404 Not Found: {"error":"not found"}
```

This is also why the request went into `ServerResponseError` rather than into another `.with_context(...)`. A context layer would have stacked a third line on an error that already has two, and two call sites (`commands/resume.rs:69`, `commands/serve.rs:673`) render the error with `error.to_string()`, which prints one layer only. Keeping method and path on the struct means those sites get the endpoint too.

### Base-path mounts

The path is taken from `resp.url()`, so a server mounted under a prefix reports the prefix it actually requested:

```
$ AI_MEMORY_SERVER_URL=http://127.0.0.1:49999/wiki ai-memory status
Error: GET /wiki/admin/status: server returned 404 Not Found: {"error":"not found"}
```

### Downcasting still works

`finalize-session` treats a 404 from `/admin/open-sessions` as "no open sessions" by downcasting to `ServerResponseError` and reading `status()`. Against the same 404 stub it still takes that branch and prints its normal message, not an error:

```
$ ai-memory finalize-session --agent claude
No open claude-code sessions matched default/ai-memory
```

### What is deliberately not shown

The query string is dropped. `ai-memory search foo` sends `?q=foo`, and the message shows `/admin/search` without it. That loses a little detail, but the query string is where a token would sit if anyone ever put one there, and `POST /workstream/runs` style paths are specific enough to find the call site. The unit test pins this by building the error from `https://username:password@example.test/workstream/runs?access_token=secret` and asserting none of `username`, `password`, `access_token`, or `secret` appear in the output.

https://claude.ai/code/session_01Uuoo7H5ZdDTZBAsGkE5VJS

PR #686, opened 2026-09-09
feat(llm): send OpenRouter app-attribution headers by default

jaysonsantos:
## Summary
- The `openai-compat` provider now sends OpenRouter's app-attribution headers (`HTTP-Referer`, `X-Title`) by default when its base URL points at `openrouter.ai`, so ai-memory's usage shows up on OpenRouter's app leaderboard.
- An explicit `AI_MEMORY_LLM_HEADERS` entry for either header still wins, following the same precedence as the default `User-Agent`.
- A non-OpenRouter compat endpoint (Ollama, vLLM, LM Studio) never receives these headers.

## Test plan
- [x] `cargo clippy -p ai-memory-llm --all-targets -- -D warnings`
- [x] `cargo test -p ai-memory-llm` (229 passed)
- [x] New unit tests in `crates/ai-memory-llm/src/factory.rs` covering header layering and operator override
- [x] New integration tests in `crates/ai-memory-llm/tests/suite/extra_headers_on_the_wire.rs` proving the headers reach the wire for an OpenRouter base URL and are absent otherwise

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- The `openai-compat` provider now sends OpenRouter's app-attribution
+  headers (`HTTP-Referer`, `X-Title`) by default when its base URL points
+  at `openrouter.ai`, so ai-memory's usage shows up on OpenRouter's app
+  leaderboard. An explicit `AI_MEMORY_LLM_HEADERS` entry for either header
+  still wins, and a non-OpenRouter compat endpoint (Ollama, vLLM, LM
+  Studio) never receives them.
+
 ### Changed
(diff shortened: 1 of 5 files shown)


akitaonrails:
Merged onto `release/2.2` for the 2.2.0 release (your commit `a5f25704`, via merge `029eaad6`) — OpenRouter app-attribution headers are a minor/additive change, so they ride the 2.2 feature train rather than a 2.1.x patch. Verified locally: rustfmt, clippy `-D warnings`, and the full workspace test suite are green on the merged branch, and your wire-level integration test confirms the headers reach an OpenRouter base URL while non-OpenRouter endpoints and an explicit `AI_MEMORY_LLM_HEADERS` override both behave as documented.

GitHub can't auto-mark this PR merged because it targets `main` and the commit landed on `release/2.2`; closing it manually. Thanks!

PR #687, opened 2026-09-09
run: accept claude*-prefixed harness names

maikyguanaes:
## Summary
- `ai-memory run` now accepts **any harness name starting with `claude`** (`claude-corp`, `claude-personal`, `claudex`, ...) — all resolving to the same `ManagedHarness::Claude` / `AgentKind::ClaudeCode`. No new harness variant, agent kind, or session-store migration.
- Unknown, non-`claude*` values are still rejected exactly as before (`invalid value ... expected one of: ...`).
- Documents that `PATH` resolution bypasses shell-only aliases (`Command::new` does not go through an interactive shell), so a `claude` defined only via `.bashrc`/`.zshrc` `alias` won't be picked up — a `PATH`-visible shim/wrapper script or `--executable PATH` is needed instead.

## Use case: more than one Claude account

If you switch between Corporate and Personal Claude accounts, name each account's `PATH` wrapper script `claude-<account>` (earlier on `PATH` than the bare `claude`) and pair it with a matching `--executable`:

```bash
# ~/bin/claude-corp and ~/bin/claude-personal each exec the real `claude`
# binary with that account's config/credentials directory.
ai-memory run claude-corp --executable claude-corp --model opus
ai-memory run claude-personal --executable claude-personal
```

Only the `--executable` value determines which binary actually runs; the positional harness name just has to start with `claude` so it isn't rejected as an unknown value, and different spellings share nothing else (same agent kind, same session store, same transcript import) — pick whatever name reads clearly to you in shell history.

## Test plan
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — no issues found
- [x] `cargo test --workspace --all-targets` — 3150 passed, 11 ignored, 0 failed (includes `claude_wildcard_names_parse_to_the_claude_harness`, `non_claude_unknown_harness_is_still_rejected`)

All run locally (rustup + pinned 1.95 toolchain).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- `ai-memory run` now accepts any `claude*`-prefixed harness name
+  (`claude-corp`, `claude-personal`, ...), all resolving to the same
+  `ManagedHarness::Claude`/`AgentKind::ClaudeCode` — no new session store,
+  migration, or agent kind. This is for callers juggling more than one
+  Claude account (e.g. Corporate and Personal): name each account's `PATH`
+  wrapper script `claude-<account>` and pair it with `--executable
+  claude-<account>` (bare names already resolve through `PATH`) so
+  `ai-memory run claude-corp --executable claude-corp` and `ai-memory run
+  claude-personal --executable claude-personal` each launch the right
+  binary while reading clearly in shell history (#1).
+
 ### Changed
 - The managed routing snippet now states that Claude Code loads `CLAUDE.md` and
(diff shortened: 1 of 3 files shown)


akitaonrails:
Merged onto `release/2.2` for the 2.2.0 release (your commit landed via merge `56f036d7`). Accepting new harness-name spellings is additive, so it rides the 2.2 feature train rather than a 2.1.x patch. One small maintainer touch on merge: the changelog entry's `(#1)` reference was corrected to `(#687)`. Verified green on the merged branch (rustfmt, clippy `-D warnings`, full workspace test — including your `claude_wildcard_names_parse_to_the_claude_harness` / `non_claude_unknown_harness_is_still_rejected` cases).

GitHub can't auto-mark this merged because the PR targets `main` and the commit landed on `release/2.2`; closing manually. Thanks!

Issue #564, opened 2026-09-01
Ambiguous unscoped writes resolve to the server default instead of erroring


akitaonrails:
Split from #562 rather than bundled, so a release that already changes a default does not carry two behaviour changes at once.

## Current behaviour

`ScopeResolver::resolve_write_args` falls back to the active-project pointer, then to the server's configured default:

```rust
active.unwrap_or((self.default_workspace_id, self.default_project_id))
```

So an unscoped `memory_write_page` whose pointer misses is written to the default project rather than refused.

## Why this is much less dangerous than it was

Before #562 the pointer was a process-wide last-write-wins slot, so a miss could route a write into *another operator's* project. With `per_actor` keying, a genuine mismatch now fails closed to the server default — a junk drawer, not someone else's work. That was the sharp edge and it is gone.

## Why it is still worth closing

The default project is still not where the caller meant to write. The page is real, it is attributed, it is searchable, and nobody looks for it there. A hard error naming the two fixes — pass explicit `workspace`/`project`, or install lifecycle hooks so the pointer is populated — turns a silent misfile into an immediate, actionable failure.

## Care needed

- Reads should keep their current fallback. A read answering from the default project is a wrong answer the caller can see; a write is a misfile they cannot.
- Anonymous/legacy callers with no coordinate at all resolve through the shared slot by design and must keep working — the error should fire only for a caller that *has* a coordinate and still fails to resolve.
- `ai-memory bootstrap` and other seeding paths write unscoped on purpose; they need explicit scope before this lands, or an exemption.

PR #688, opened 2026-09-09
fix(scope): seed the active-project fallback from disk on startup (#678)

### LINKED ISSUE #678: Reads silently return zero counts when the daemon restarts mid-session (ActiveProject lost, no error)
**Version**

`ai-memory 2.1.1`

**What happened**

After `systemctl --user restart ai-memory.service` during a live Claude Code session — the restart the package's own `post_upgrade()` tells you to do — `memory_status` with no arguments returned an empty project:

```json
{"counts":{"pages_latest":0,"pages_all":0,"sessions":0,"observations":0}}
```

No error, no warning in the server log. The data was fully intact. Reading the SQLite directly at the same moment:

```
observations 2670   sessions 19   pages 9   projects 4

denfis             2289
victorandrade.dev   373
judeu                 8
scratch               0
```

And passing the scope explicitly to the very same tool returned the right numbers:

```json
{"counts":{"pages_latest":3,"pages_all":4,"sessions":10,"observations":373}}
```

**Why (as far as I can tell)**

Reads resolve their scope through the hook-published `ActiveProject`, which lives in server memory. From `crates/ai-memory-mcp/src/server.rs`:

```rust
/// Falls back to the baked default only when no `ActiveProject`
/// has been published yet (early startup / no hooks).
```

A restart drops that state. The read then falls back to the baked default scope, and `status_counts_for_project` on that empty scope returns zeros through the **success** path. The caller cannot distinguish "scope could not be resolved" from "this project genuinely has no memory yet".

**What I expected**

Either the `ActiveProject` is recovered from the database on start — the hook history that would rebuild it is already persisted — or the read tells the caller its scope could not be resolved, so the agent can retry with an explicit `workspace`/`project`.

**Impact**

This is a bad failure shape specifically for a memory server. The agent asks "what do we have on this project?", is told "nothing", and reports to the user that there is no memory — while thousands of observations sit in the DB. It is silent on both sides: no error for the model to react to, nothing in the log for the user to find. The window opens on the normal upgrade path, since upgrading the package is exactly when the service restarts.

`memory_status` is just where I noticed it; the same `effective_ids_for_read_args_with_actor` path backs `memory_briefing` and the other reads.

**Steps to reproduce**

1. Run a session with the lifecycle hooks active so observations are being captured for the current project.
2. `systemctl --user restart ai-memory.service` while that session is still open.
3. Call `memory_status` with no arguments → all zeros.
4. Call it again with explicit `workspace` + `project` → correct counts.

**Environment**

- OS: Arch Linux (kernel 7.1.9)
- Docker? no
- Agent: Claude Code
- Transport: http (user systemd service, `127.0.0.1:49374`)

**Relevant logs**

Nothing is emitted for the fallback itself. The restart was otherwise clean — migrations V59 and V60 applied without error, and `PRAGMA integrity_check` on the database returns `ok`:

```
INFO refinery_core::traits::sync: applying migration: V59__purged_scopes_tombstones
INFO refinery_core::traits::sync: applying migration: V60__bootstrap_chunk_progress
INFO ai_memory_cli: ai-memory starting version="2.1.1" ...
```

rafaelkenedy:
Fixes #678.

## The failure

The active-project pointer lives only in process memory, so restarting the daemon mid-session — the restart `post_upgrade()` itself tells you to run — drops it. Every keyed read then misses, an unscoped read resolves through the baked default scope, and `memory_status` answers zero counts for a project holding thousands of observations. It does so through the **success** path with nothing in the log, so neither the agent nor the operator can tell "scope unresolved" from "project genuinely empty". `memory_briefing` and every other read on `effective_ids_for_read_args_with_actor` share the path.

## The approach

This is the remedy scoped in @akitaonrails' triage comment on the issue, not the fail-closed read: a keyed miss is also the normal shape of the pre-publish window (hooks are fire-and-forget, so a session's first read can arrive before its `SessionStart` write lands) and of TTL/cap eviction on a busy server, and the multi-session stress guards document that an evicted keyed read degrades rather than errors. Erroring there would trade a silent wrong answer for a broad spurious-error regression.

So the degraded answer is made a real one instead. `serve` seeds a **read-side fallback slot** at startup from the most recently active project already in SQLite — the newest observation names the scope the operator was in, since every lifecycle event lands as one stamped with its routed scope.

Four bounds keep it narrow:

- **Reads only.** The seed lives in its own slot, so an unscoped write resolves exactly where it did before: `resolve_write_args` stays on `lookup_for`, which still answers `Unset`/`Mismatch` for a caller the pointer cannot place. This is @abhisheksharma2411's finding 1 below — reproduced with their probe, fixed, and pinned by a test.
- **Keyed entries are never reconstructed**, so a keyed hit still wins and per-actor isolation is untouched.
- **Bounded by the pointer's own per-key TTL** (`[auto_scope] session_ttl_secs`, default 1h). Activity older than that would have aged out of a live pointer, so a server that was down overnight starts with no fallback at all rather than answering for a project nobody is in any more.
- **The first foreground hook event supersedes it** for every caller, exactly as it overwrites any other shared-slot value. Admin invalidations (`purge-project`, `move-project`, workspace removal) reach the seed too, so a scope that no longer exists cannot keep answering from it.

Seeding runs on the HTTP transport only, where the hook ingress that owns the pointer also lives; stdio keeps its baked default unchanged. The read is non-fatal — a server that cannot answer it starts exactly as it does today.

## Changes

| file | what |
|---|---|
| `crates/ai-memory-store/src/reader.rs` | new `most_recently_active_scope(since_us)` — one row off `idx_observations_created_at` |
| `crates/ai-memory-core/src/active_project.rs` | the `seeded` slot, `seed_read_fallback()` / `seeded()`, `get_for_read()`, and the `per_key_ttl()` accessor (the normalized TTL, so the seed cutoff cannot drift from eviction) |
| `crates/ai-memory-store/src/scope.rs` | the read path resolves through `get_for_read`; the write path is untouched |
| `crates/ai-memory-cli/src/commands/serve.rs` | `seed_active_project_fallback()`, called at the top of the HTTP arm |
| `docs/auto-scope.md` | new "Surviving a restart" section, including why reads do not fail closed |
| `CHANGELOG.md` | `### Fixed` entry (#678) |

## Tests

- `crates/ai-memory-store/tests/suite/most_recently_active_scope.rs` (new, direct-SQL fixture so timestamps are controlled): resolves the newest observation's scope across projects, honours the recency cutoff in both directions, `None` on a store with no activity.
- `serve.rs`: the #678 regression itself — a caller whose keyed entry died with the process resolves through `ScopeResolver::resolve_read_args` to the real project instead of the baked `scratch`, while that same caller's unscoped **write** still resolves to the default; plus a guard that a pointer a hook already published is never seeded over (its fixture now records activity in another project, so the test fails when the guard is removed — finding 2).
- `ai-memory-core`: the reviewer's probe as a test — a seeded fallback answers `get_for_read` for a stranger while `lookup_for` still says `Unset` and `get_for` still says `None`; a published pointer outranks the seed and turns a stranger back into `Mismatch`; every invalidation path reaches the seed.
- `ai-memory-store/src/scope.rs`: read-vs-write resolution under a seed for both a session-carrying and a coordinate-less caller, plus the named-project cross-workspace case.

## Invariant check (#16, multi-session/multi-user)

Scope resolution changes have to be argued against the shared-project invariant, so explicitly: the per-key map and `ActiveProjectMode` semantics are unchanged, `lookup_for` is unchanged, and `Mismatch` still fails closed for writes (#564). The only new state is a read-side slot that no write path consults, so the set of scopes an unscoped write can reach is exactly what it was. `crates/ai-memory-store/tests/suite/multi_session.rs` and the `active_project` pointer tests pass unchanged.

## Gates

`cargo fmt --all -- --check`, `git diff --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --all-targets` all clean on Windows, with one unrelated flake: `ai-memory-consolidate::auto_improve::tests::eval_error_timeout_and_invalid_json_fail_closed` (a 1-second process timeout) fails under parallel load and passes in isolation; this branch does not touch that crate.

https://claude.ai/code/session_015xtQHDFw2S2Ch1ZNGoE62m

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   now uses the import instead of a prose pointer (#680).
 
 ### Fixed
+- Reads no longer report an empty project after the daemon restarts
+  mid-session. The active-project pointer lives in process memory, so a
+  restart — the one the packages' own post-upgrade note tells you to run —
+  dropped it, and `memory_status`, `memory_briefing`, and every other
+  unscoped read then resolved through the baked default scope and answered
+  zero counts through the success path, with nothing in the log to
+  distinguish "scope unresolved" from "project genuinely empty". `serve` now
+  seeds the shared fallback slot at startup from the most recently active
+  project recorded in the database, bounded by the same TTL as a per-key
+  entry, so a keyed miss right after a restart degrades to real data. Keyed
+  per-actor entries are never reconstructed and the first hook event
+  publishes over the seed, so pre-publish and eviction reads keep degrading
+  exactly as they do today (#678).
 - The from-source AUR `PKGBUILD` now builds and tests on constrained AUR
   builders. Release LTO was disabled (`options=('!debug' '!lto')`) so the
   final link no longer gets OOM-killed on low-memory build hosts, and the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,11 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   unscoped read then resolved through the baked default scope and answered
   zero counts through the success path, with nothing in the log to
   distinguish "scope unresolved" from "project genuinely empty". `serve` now
-  seeds the shared fallback slot at startup from the most recently active
-  project recorded in the database, bounded by the same TTL as a per-key
-  entry, so a keyed miss right after a restart degrades to real data. Keyed
-  per-actor entries are never reconstructed and the first hook event
-  publishes over the seed, so pre-publish and eviction reads keep degrading
+  seeds a read-side fallback at startup from the most recently active project
+  recorded in the database, bounded by the same TTL as a per-key entry, so a
+  keyed miss right after a restart degrades to real data. The seed serves
+  reads only: keyed per-actor entries are never reconstructed, an unscoped
+  write resolves exactly where it did before, and the first hook event
+  supersedes the seed — so pre-publish and eviction reads keep degrading
   exactly as they do today (#678).
 - The from-source AUR `PKGBUILD` now builds and tests on constrained AUR
   builders. Release LTO was disabled (`options=('!debug' '!lto')`) so the
--- a/crates/ai-memory-cli/src/commands/serve.rs
+++ b/crates/ai-memory-cli/src/commands/serve.rs
@@ -2139,7 +2139,7 @@ fn host_allowed(host: &str, allowed_hosts: &[String]) -> bool {
     })
 }
 
-/// Seed the single-slot active-project fallback from the most recently active
+/// Seed the read-side active-project fallback from the most recently active
 /// project already on disk.
 ///
 /// The pointer lives only in process memory, so restarting the daemon
@@ -2152,11 +2152,13 @@ fn host_allowed(host: &str, allowed_hosts: &[String]) -> bool {
 /// fire-and-forget) and of TTL/cap eviction, both of which must keep
 /// degrading gracefully.
 ///
-/// So make the degraded answer a real one. Only the shared slot is seeded, so
-/// a keyed hit still wins and the first hook event publishes straight over it.
-/// The recency bound is the pointer's own per-key TTL: activity older than
-/// that would have aged out of a live pointer anyway. Non-fatal — a server
-/// that cannot read this starts exactly as it does today.
+/// So make the degraded answer a real one. The seed lands in a slot only READS
+/// consult: a keyed hit still wins, an unscoped write from an unrecognized
+/// caller still fails closed rather than being attributed to a reconstructed
+/// project, and the first hook event publishes straight over it. The recency
+/// bound is the pointer's own per-key TTL: activity older than that would have
+/// aged out of a live pointer anyway. Non-fatal — a server that cannot read
+/// this starts exactly as it does today.
 async fn seed_active_project_fallback(reader: &ReaderPool, active_project: &ActiveProject) {
     if active_project.get().is_some() {
         return;
@@ -2167,7 +2169,7 @@ async fn seed_active_project_fallback(reader: &ReaderPool, active_project: &Acti
         .saturating_sub(ttl_us);
     match reader.most_recently_active_scope(since).await {
         Ok(Some((workspace_id, project_id))) => {
-            active_project.set(workspace_id, project_id);
+            active_project.seed_read_fallback(workspace_id, project_id);
             let workspace = reader
                 .workspace_name_by_id(workspace_id)
                 .await
@@ -3047,9 +3049,14 @@ mod tests {
         let active_project = ActiveProject::new();
         seed_active_project_fallback(&store.reader, &active_project).await;
         assert_eq!(
-            active_project.get(),
+            active_project.seeded(),
             Some((workspace_id, worked_in)),
-            "the shared slot must name the project the last activity landed in"
+            "the seed must name the project the last activity landed in"
+        );
+        assert_eq!(
+            active_project.get(),
+            None,
+            "a reconstruction is not a publish: the shared slot stays empty"
         );
 
         // The live session's keyed entry died with the process, so its read
@@ -3068,6 +3075,22 @@ mod tests {
             (workspace_id, worked_in),
             "an unscoped read after a restart must not silently answer for the baked default"
         );
+
+        // ...while an unscoped WRITE from a caller the pointer cannot place is
+        // untouched by the seed: it resolves exactly where it did before, so a
+        // page is never attributed to a project reconstructed from a session
+        // that is not this caller's.
+        let written = ai_memory_store::ScopeResolver::new(&store.reader, workspace_id, scratch)
+            .with_writer(&store.writer)
+            .with_active_project(&active_project)
+            .resolve_write_args(None, None, &actor)
+            .await
+            .unwrap();
+        assert_eq!(
+            written.as_tuple(),
+            (workspace_id, scratch),
+            "the seed must not retarget unscoped writes"
+        );
     }
 
     #[tokio::test]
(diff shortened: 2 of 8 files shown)


abhisheksharma2411 (commented):
Reviewed this closely because it lands in the scope resolver I touched in #582, and the write-side there fails closed for a reason. The approach is right and matches @akitaonrails' triage exactly — "seed the single-slot fallback from the most-recently-active project" — and the three bounds you chose are the correct three. Two findings, one of which I think matters.

**What I verified rather than read:**

- `idx_observations_created_at ON observations(created_at DESC)` exists (`V08__performance_indexes.sql:23`), so "one row off the index" is accurate.
- 425 tests pass in `ai-memory-store` including your three new `most_recently_active_scope` cases and the whole `multi_session` suite; 32 `active_project` tests pass.
- `write_single` does not touch `ever_keyed` — only `set_for` does (`:459`) — so your "keyed entries are never reconstructed, a keyed hit still wins" claim holds by construction.

## 1. In `PerActor` mode the seeded slot answers for *any* actor, including a different operator

`PerActor` is the default (`#[default]` on the variant), and `ever_keyed` is per-process, so it is `false` for the whole window between startup and the first hook publish. In that window `lookup_for` falls through to `from_single(read_single())` — which the seed has now made non-empty.

Probe, on your branch:

```rust
let ap = per_actor();
ap.set(ws_alice, proj_alice);          // exactly what seed_active_project_fallback does
ap.lookup_for(&key_actor("bob", "s-bob-never-published"))
```

```
Resolved(WorkspaceId(01a087de-…0a71), ProjectId(01a087de-…8177))   <- alice's scope
```

Not `Mismatch`, not `Unset`. So during that window Bob's unscoped **read** answers from Alice's project — and more importantly his unscoped **write** is *attributed* to it, because `resolve_write_args` consumes `Resolved` directly and only errors on `Mismatch` (`scope.rs:446-451`). Before this change that same write got `Unset` and landed in the configured default.

The reason I'd flag it rather than wave it through is that `lookup_for`'s own comment describes this exact hazard:

> answering from the shared slot would route the request into whichever project published last, possibly another operator's

The seed creates precisely that condition, at the one moment `ever_keyed` cannot tell a hookless install (where reading the shared slot is correct and `a_hookless_install_still_reads_the_shared_slot` pins it) from a hook-fed install that just restarted (where it is not).

It is bounded — one window per restart, ending at the first hook event — and on a single-operator install it is a pure improvement. On a shared server it is a cross-operator write attribution, which is the class #564 and #582 exist to prevent.

The shape that would keep your fix and close this: track the seeded value as *seeded* rather than *published* — a separate slot, or a flag beside the single slot — and have `lookup_for` offer it to reads while continuing to return `Mismatch`/`Unset` for the unscoped-write path. #678 is a read complaint ("`memory_status` returned zero counts"); nothing in it asks for writes to change target. That also lets you keep the fix in `PerActor` instead of restricting it to `Single`, which would gut it given the default.

Your call entirely — you may reasonably judge the window too narrow to be worth the extra state, and I'd rather flag it than quietly assume.

## 2. `seeding_never_overwrites_a_pointer_a_hook_already_published` passes with the guard deleted

```
remove `if active_project.get().is_some() { return; }`
  -> test seeding_never_overwrites_a_pointer_a_hook_already_published ... ok
     2 passed; 0 failed
```

The fixture creates the `published` project and sets the pointer, but inserts **no observations** — so `most_recently_active_scope` returns `None` and the seed no-ops whether or not the guard is there. It passes for the wrong reason.

One addition fixes it: activity in a *different* project, so the seed has something to overwrite the pointer with. I tried it both ways to be sure rather than suggest it blind:

```
with the extra observation + guard present  -> ok
with the extra observation + guard removed  -> FAILED
    left:  ProjectId(…8925)   (other)
    right: ProjectId(…a56d)   (published)
```

Happy to push that as a suggestion if it's easier than retyping the fixture.

## Nit

The PR body cites `crates/ai-memory-store/tests/multi_session.rs`; it's `tests/suite/multi_session.rs`. It does exist and does pass — just so nobody else goes looking in the wrong place.

The invariant section in your description is the part I'd most want other contributors to copy — arguing the change against the shared-project invariant explicitly, before anyone asks, is why finding 1 above took me twenty minutes instead of an afternoon.

rafaelkenedy:
Thank you — finding 1 is real, and I'd missed it. Reproduced it as your probe before changing anything:

```
left:  Resolved(WorkspaceId(01a088b5-…3750), ProjectId(01a088b5-…4778))
right: Unset
```

`ever_keyed` being per-process is the part I reasoned past: I checked that a keyed *hit* still wins and stopped there, without walking the keyed-*miss* fallthrough for an actor other than the one the seed came from. And you're right that the write is the sharp end — a read answering from the wrong project is visible to the caller, an attributed page in someone else's project is not.

Fixed in d38452e by taking your suggested shape: the seed gets its own slot and only reads consult it.

- `ActiveProject::seed_read_fallback` fills a `seeded` slot; `set` / `set_for` and the per-key map are untouched, so `lookup_for` is byte-for-byte what it was.
- New `get_for_read` = `get_for`, plus the seed **only** for `Unset`. `Mismatch` still resolves to nothing: hooks are live there and the coordinate matched none of them, so the caller's own default remains the honest answer.
- `ScopeResolver::resolve_current_or_project` uses it. `resolve_write_args` stays on `get_for` / `lookup_for`, so #564 and #582 behave exactly as before — including the explicit-project branch, which keeps inferring its workspace from published state only.
- `clear`, `clear_project`, `clear_workspace`, `retarget_project_workspace` and the `contains_*` probes reach the seed as well. That one wasn't in your review, but it follows from the same "is this a publish or a reconstruction?" question: a purge that leaves the seed pointing at a dead project would reintroduce the bug through the other door.

Your probe is now `a_seeded_fallback_serves_reads_without_placing_a_foreign_actor`, asserting all four outcomes at once (`lookup_for` → `Unset`, `get_for` → `None`, `get_for_read` → the seed, single slot → empty), plus `a_published_pointer_outranks_the_seed` for the transition back to `Mismatch` once anything is keyed. `scope.rs` got the read-vs-write pair for both a session-carrying and a coordinate-less caller, and the serve test now asserts the write target is unchanged alongside the read.

Finding 2 is correct too, and thank you for actually deleting the guard rather than eyeballing it — the fixture proved nothing. It now records activity in a different project first, and I verified it both ways:

```
guard present -> ok
guard removed -> FAILED
    left:  Some((…c0af, …c9355))   (seeded from `elsewhere`)
    right: None
```

Nit fixed in the description (`tests/suite/multi_session.rs`), and the invariant section now says what actually holds after this change: the only new state is a read-side slot no write path consults, so the set of scopes an unscoped write can reach is exactly what it was.

Gates re-run on the full change: `fmt`, `clippy --workspace --all-targets -D warnings`, and `cargo test --workspace --all-targets` green, except the pre-existing `ai-memory-consolidate::auto_improve::tests::eval_error_timeout_and_invalid_json_fail_closed` flake (1-second process timeout; passes in isolation, and this branch does not touch that crate).

Issue #653, opened 2026-09-06
purge-session leaves the session's wiki page on disk, so a reindex brings it back


gb:
## The problem

`ai-memory purge-session <id>` (and `POST /admin/purge-session`) is meant to erase one session completely. Today it deletes the session's rows from SQLite, but it never touches the wiki file for that session, `sessions/<id>.md`. The file stays on disk with its full content.

That file is not just a leftover. The wiki is the source of truth and the file watcher rebuilds the index from what it finds on disk. On the next reconcile pass, or on `reindex`, the "purged" page is picked up again and inserted back into the database. From the user's point of view the session was deleted, then it came back.

## How to see it

1. End a session so it has a summary page at `sessions/<id>.md`.
2. Run `ai-memory purge-session <id> --confirm`.
3. Look at `<data_dir>/wiki/<workspace>/<project>/sessions/<id>.md`. It is still there.
4. Run `ai-memory reindex`. The page is searchable again.

## Why it matters

- People use `purge-session` to honour "forget this conversation" requests. Right now that promise is only half kept.
- The CLI output says the wiki pages were removed, which is not true. The list it prints comes from the database, not from the disk.
- The sibling commands already do this right. `purge-project` and `delete-workspace` remove their files after the database transaction commits, report what was and was not removed, and tell webhook observers when file cleanup failed. `purge-session` is the odd one out.

## What a fix should do

- After the SQL purge commits, remove the returned page paths under that project's own directory, so a page with the same name in another project is not touched.
- Report the real result in the response: which files were removed, and which could not be.
- If a file could not be removed, keep the database purge and say so, instead of rolling back or pretending it worked.

Related: the deletion epic #387 mentions this gap, and PR #248 tried an earlier form of the fix but was never merged.

PR #696, opened 2026-09-10
fix(wiki): serialize session purge with page mutations

samirhvbr:
## What changed

`purge-session` now acquires the wiki's exclusive mutation guard before submitting the SQL purge and holds it through page-file cleanup. The admin handler delegates both steps to `Wiki::purge_session`. HTTP fields, authorization, admission and partial-failure reporting are unchanged.

## Why

Follow-up to #653 / #654. Previously the database purge completed before file removal acquired the guard. A watcher already holding a shared guard could reinsert the still-existing page between those steps; cleanup would then remove only the file. A concurrent page write or wiki move could also interleave with cleanup.

## Test plan

- [x] `git diff --check`
- [x] `cargo test --workspace --all-targets`
- [x] `cargo deny --all-features check`
- Manual validation: no live-server test. A deterministic regression holds the watcher's shared guard, polls the purge, fences the writer queue, and verifies SQL has not deleted the session. It then finishes reindexing, releases the guard and verifies purge removes both the file and searchable page. Moving lock acquisition after SQL makes this test fail.
- A second regression verifies a wrong-project store error preserves the page file. Existing admin tests cover cleanup failures and sibling-project isolation.

## Commit attribution

- [x] Verified author name and GitHub noreply email.

## Release impact

- [x] **Patch** — bug fix, no new operator surface
- [ ] **Minor**
- [ ] **Major (breaking)**

## CHANGELOG (merge gate)

- [x] Added a `[Unreleased]` / `Fixed` entry and lifecycle documentation.
- [x] No defaults or response shape changed.

## Notes for reviewers

The shared-project invariant is enforced by using the same guard as page writes, reindexing and wiki moves. Admission stays outside the lock. SQL failure removes no files; filesystem failure still leaves the SQL purge committed and reports `files_failed`. This is process-local coordination, not crash-atomic rollback. The exclusive guard also covers optional SQL compaction, so a compacting purge can hold up wiki mutations longer.

Maintainer action: please add the `windows` label for the locking change. The author account lacks permission to apply it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   now uses the import instead of a prose pointer (#680).
 
 ### Fixed
-- `purge-session` held the wiki mutation guard across database deletion and file
-  cleanup, preventing an in-flight watcher reindex from restoring deleted rows
-  or a concurrent page write from losing its file during cleanup (#653).
+- `purge-session` took the wiki mutation guard only for the file cleanup, after
+  the database deletion had already committed. A watcher reindex could reinsert
+  the still-present page in that gap and keep only the row, and a concurrent page
+  write could lose its file to the cleanup. The purge now holds the guard across
+  both steps, so reindexes, page writes and wiki moves finish before it starts
+  and wait until it is done (#696, follow-up to #653).
 - The from-source AUR `PKGBUILD` now builds and tests on constrained AUR
   builders. Release LTO was disabled (`options=('!debug' '!lto')`) so the
   final link no longer gets OOM-killed on low-memory build hosts, and the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   now uses the import instead of a prose pointer (#680).
 
 ### Fixed
+- `purge-session` held the wiki mutation guard across database deletion and file
+  cleanup, preventing an in-flight watcher reindex from restoring deleted rows
+  or a concurrent page write from losing its file during cleanup (#653).
 - The from-source AUR `PKGBUILD` now builds and tests on constrained AUR
   builders. Release LTO was disabled (`options=('!debug' '!lto')`) so the
   final link no longer gets OOM-killed on low-memory build hosts, and the
--- a/crates/ai-memory-mcp/src/admin.rs
+++ b/crates/ai-memory-mcp/src/admin.rs
@@ -3673,7 +3673,7 @@ async fn handle_purge_session(
     {
         Ok(s) => s,
         // Absent from this scope (or already purged) is a 404, not a fault.
-        Err(e @ StoreError::NotFound(_)) => {
+        Err(WikiError::Store(e @ StoreError::NotFound(_))) => {
             return (
                 StatusCode::NOT_FOUND,
                 Json(serde_json::json!({ "error": e.to_string() })),
--- a/crates/ai-memory-mcp/src/admin.rs
+++ b/crates/ai-memory-mcp/src/admin.rs
@@ -3598,38 +3598,6 @@ pub struct PurgeSessionReport {
     pub checkpoint: Option<String>,
 }
 
-async fn remove_purged_session_storage(
-    state: &AdminState,
-    workspace_id: WorkspaceId,
-    project_id: ProjectId,
-    removed_paths: &[PagePath],
-) -> (Vec<PagePath>, Vec<PagePath>) {
-    let mut files_deleted = Vec::with_capacity(removed_paths.len());
-    let mut files_failed = Vec::new();
-
-    for path in removed_paths {
-        match state
-            .wiki
-            .remove_page_file(workspace_id, project_id, path)
-            .await
-        {
-            Ok(true) => files_deleted.push(path.clone()),
-            Ok(false) => {}
-            Err(error) => {
-                warn!(
-                    operation = "purge-session",
-                    path = path.as_str(),
-                    error = %error,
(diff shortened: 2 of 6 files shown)


samirhvbr:
Self-review of this branch. Pushed 8017a8bd with the fixups; two things I did **not** change are below, one of which limits what the changelog entry can honestly claim.

## Pushed

- `Wiki::purge_session` returned `StoreResult` — the only fallible method on `Wiki` that did not return `WikiResult`. It returns `WikiError::Store` now, and the admin handler matches the wrapped variant the way the `move_project_workspace` arm above it already does.
- `PurgeSessionOutcome` derives `Debug, Clone` like `MoveSessionOutcome` beside it (Copilot flagged this too). The concurrency regression prints the outcome instead of a bool.
- Second regression for the other half of the claim: a page write cannot install its file while a purge owns the guard.
- `files_deleted` pre-sized from `removed_paths` again; log fields one per line; `WriterHandle::purge_session` doc-warns that server callers must go through `Wiki::purge_session`.
- Changelog: the entry stated the *fixed* behaviour in past tense, which reads as the bug when scanning `### Fixed`. Also reused `(#653)`, which the 2.1.x entry already carries for the earlier fix. Now states bug-then-fix and references this PR.

One note on the new test, because the first version I wrote was wrong: a single poll does **not** prove the write was excluded. `write_page` parks on the writer queue before it touches disk, so with the guard removed entirely the single-poll version still passed. It gives the write real time now, and I confirmed the mutant dies (remove the shared guard from `write_page` → fails).

## Not fixed: a purge with `files_failed` is undone by the next reconcile tick

The guard closes the in-process window. It cannot close the one where the file *outlives* the purge — and that is a normal, already-tested outcome, not just a crash:

- `purge_session_reports_file_cleanup_failure_after_db_commit` deliberately produces it and asserts we report `files_failed`;
- so, by inspection, does a client that disconnects or times out after `WriterHandle::purge_session` enqueued the command: the writer dispatch logs the dropped receiver and still runs the purge, while the dropped handler future releases the guard before the cleanup loop starts. The repro below uses the first route, not this one.

In that state the rows are gone and `sessions/<id>.md` is still on disk. Nothing on the reindex path consults the tombstone: `purged_sessions` is written in the purge transaction (`ops.rs:3564`) and read only by the ingest guard (`session_is_purged`, `ops.rs:1223`). `reindex_all` has the scope equivalent — `scope_is_purged` at `wiki.rs:1598` (this branch's head) — but a session purge does not touch the project row, so `reconcile`'s orphan guard (`watcher.rs:347`) passes and the page is reindexed on the next tick, 30s by default.

I reproduced it in the wiki crate's own harness, asserting the preconditions so it cannot pass vacuously (SQL committed, file present, `files_failed` reported, page unsearchable, `skipped_orphans == 0`), then one real `reconcile(&wiki)` call:

```
purged_session_page_is_resurrected_by_the_next_reconcile_tick ... FAILED
  a purged session's page must not be resurrected from a leftover file:
  [PageHit { path: "sessions/01a08aff-....md", snippet: "<mark>zimbabwe</mark> pineapple" }]
```

To make the unlink fail without destroying the file, the test sets the parent directory read-only — `EACCES`, markdown intact. Replacing the file with a directory (what the admin test does) is not usable here because the next step has to read the markdown back.

This predates the PR and fixing it here would change its shape, so I left it out. It does mean the entry's "a watcher reindex can no longer restore deleted rows" is true only while cleanup succeeds. Happy to open it as an issue, or send the fix — mirroring `scope_is_purged` with a session gate on the reindex path — whichever you prefer. Say the word and I'll write it rather than have you pick it up; I have the repro ready either way.

## Documented rather than restructured

The guard is taken *before* the purge is submitted to the writer actor, so it also covers the wait on whatever that single queue is already draining, and — with `compact: true` — the `VACUUM` after the commit. The PR body mentioned the `VACUUM`; it did not mention the queue wait. I added both to `docs/lifecycle-ops.md` instead of restructuring, since taking the `VACUUM` out (purge with `Compaction::Skip` under the guard, `WriterHandle::compact()` after releasing it) changes where `compacted` comes from and felt like your call. Say if you want that shape and I'll push it.

Same for the git checkpoints: `checkpoint_or_500` / `checkpoint_or_warn` call `commit_all`, which is sync and cannot take the async guard, so they bracket the guarded region from outside rather than snapshotting it. Pre-existing, now written down beside the guarantee so the doc does not overstate it.

## Unrelated heads-up

`cargo test --workspace --all-targets` fails 4 tests here that are not from this branch — `cursor_apply_is_idempotent`, `antigravity_apply_is_idempotent`, `kimi_code_apply_is_idempotent`, `kimi_code_apply_preserves_providers_and_third_party_hooks`, all `assertion left == right failed: second apply must be a no-op`. I confirmed them on `aed9b8be` with this branch's changes fully reverted, and CI is green on the same SHA, so it is environment-dependent rather than a branch regression. The rest of the workspace passes — 2,174 tests with `--exclude ai-memory-cli`, and the cli crate itself is 862 passing beside those 4. Flagging in case it is news; I have not dug into why.

samirhvbr:
Both open items are out of scope for this PR, so nothing here should hold up a merge. The resurrection case is now #701, filed against `main` with a repro that runs at `55fcce35` — it is pre-existing, and this PR narrows the window rather than causing it. The `VACUUM` point is documented in `docs/lifecycle-ops.md` rather than restructured: taking it out of the guarded region changes where the reported `compacted` flag comes from, so I left that as your call and will send it as a follow-up if you want that shape.

Retracting the test-failure note from my earlier comment — it does not reproduce. The same commit `8017a8bd` gives 866 passed / 0 failed in a fresh checkout under `$HOME`, and those four `install_hooks` failures appear only in the sandbox worktree I had been building in under `/tmp`. That was my environment, not your code, and I should have tried a second checkout before reporting it. I have not looked into why the path matters.

Issue #708, opened 2026-09-10
Multi-user servers have no per-project authorization


luisfnicolau:
**Problem this would solve**

2.1.1 authenticates users but does not authorize them against projects. Any
account can read every project on the server.

Reproduced on `195be4e5`, with two ordinary users (`role=user`, neither `root`):

```
alice  →  memory_write_page  default/alice-client-work/secrets/rates.md
          body: "ALICE CONFIDENTIAL: day rate is 900 EUR."

bob    →  memory_read_page   ✓ full body returned
       →  memory_query       ✓ "ALICE CONFIDENTIAL: day rate is 900 EUR."
```

`bob` was created seconds earlier and shares nothing with `alice` beyond the
server. Confirmed in the schema: `users`, `web_sessions` and `api_credentials`
carry no reference to a project or repository, and there is no grant, ACL or
membership table.

This is fine — arguably correct — for the documented single-operator and
homelab cases. It becomes a problem the moment a server holds work for more
than one team. Ours holds projects from three unrelated GitHub
organisations, and "everyone with an account can read all of it" is not a
position we can be in.

**Proposed solution**

A grant table and one authorization decision, deny-by-default:

```sql
CREATE TABLE memory_grant (
    id                 BLOB PRIMARY KEY NOT NULL,
    user_id            BLOB NOT NULL REFERENCES users(id)    ON DELETE CASCADE,
    repository_id      BLOB NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
    granted_by_user_id BLOB NOT NULL REFERENCES users(id),
    granted_at         INTEGER NOT NULL,
    revoked_at         INTEGER,
    revoked_by_user_id BLOB REFERENCES users(id)
);

CREATE UNIQUE INDEX idx_memory_grant_active
    ON memory_grant(user_id, repository_id) WHERE revoked_at IS NULL;
```

`root` keeps full access. A `user` sees a project only with an active grant.
The check lives at the route boundary rather than in each handler, so no route
can forget it, and it is one place to audit.

Two details that matter more than they look:

*Revocation is a timestamp, not a delete.* A grant records that someone was
given access, by whom, and when — the same evidence the purge design already
protects when it keeps the row and destroys only the content. Deleting the row
erases the answer to "who could see this, and for how long", which is the one
question a security review actually asks.

*Absence must not read as emptiness.* A denied read has to say "you do not have
access to this repository", never return an empty result. Silent emptiness is
[...]

PR #709, opened 2026-09-10
test(autoscope): budget process spawn, not the product

luisfnicolau:
`autoscope_env` fails intermittently on macOS. Three consecutive runs of the
unchanged suite, on an idle machine, failed **1, then 5, then 7** of its 10
tests — each finishing in almost exactly 8.0s, against `STARTUP_TIMEOUT` of 8s.

```
run 1: FAILED. 9 passed; 1 failed;  finished in 8.01s
run 2: FAILED. 5 passed; 5 failed;  finished in 8.02s
run 3: FAILED. 3 passed; 7 failed;  finished in 8.03s
```

## Why

These tests spawn a server and wait for its startup log line. The server's own
startup work is well under a second; what the budget actually bounds is ten
copies of an unoptimised debug binary being spawned, linked and paged in at once
by the harness. That cost tracks binary size and machine load rather than
anything about the product, so the failure count measures how busy the machine
is rather than whether the code works.

## After

Same suite, same machine, four consecutive runs:

```
run 1: ok. 10 passed;  finished in 16.29s
run 2: ok. 10 passed;  finished in 28.51s
run 3: ok. 10 passed;  finished in  8.15s
run 4: ok. 10 passed;  finished in  5.06s
```

The spread is the point: identical work varying five-fold, and the 8.15s run
landing a fifth of a second the wrong side of the old budget. A real startup
regression still fails at 45s — it just has to be a large one, which seems the
right trade for a test whose failure would otherwise be read as noise.

## Also in this change

The three fail-fast checks had the same defect and are now one named constant
rather than three inline `Duration::from_secs(5)`. They bound process spawn plus
config parse and read "still running" as a failure to reject a bad config, so
under load a slow spawn is indistinguishable from a misconfiguration being
silently accepted — which is the bug those tests exist to catch. They did not
fail in my runs; a downstream fork carrying a larger binary saw them fail for
this reason, which is what prompted looking at them.

Both constants now carry the reason they are large, so the next person to see
45s does not tighten it back.

## Notes

Tests only — no product code touched.

Likely invisible in CI today: `ci.yml` runs Linux by default, and the macOS leg
joins on a `full-ci` label or manual dispatch. A loaded shared runner is exactly
where this bites, so it may be costing occasional red runs on the pre-release
macOS leg.

Happy to split the fail-fast half out if you would rather take only the
demonstrated failure.

### DIFF
--- a/crates/ai-memory-cli/tests/suite/autoscope_env.rs
+++ b/crates/ai-memory-cli/tests/suite/autoscope_env.rs
@@ -152,7 +152,33 @@ fn spawn_and_wait_for_exit(envs: &[(&str, &str)], timeout: Duration) -> (bool, S
 }
 
 const STARTUP_NEEDLE: &str = "active-project isolation mode";
-const STARTUP_TIMEOUT: Duration = Duration::from_secs(8);
+
+/// Wall-clock budget for the server to reach its startup log line.
+///
+/// Deliberately generous, because it measures the wrong thing on purpose. The
+/// server's own startup work is well under a second; what this actually bounds
+/// is ten copies of an unoptimised debug binary being spawned, linked and paged
+/// in at once by the test harness. That cost tracks binary size and machine
+/// load rather than anything about the product, so a tight budget turns a busy
+/// machine into a test failure.
+///
+/// A real startup regression still fails here. It just has to be a large one,
+/// which is the correct trade for a test whose failure would otherwise be read
+/// as noise. Concretely, the measured spread while writing this was 5.06s on a
+/// quiet machine and 28.5s under parallel load, so 45s trips on roughly a 9x
+/// regression against a quiet run and a 1.6x one against the worst observed.
+/// That is the inverse bound: anything subtler than that belongs in a
+/// benchmark, not here.
+const STARTUP_TIMEOUT: Duration = Duration::from_secs(45);
+
+/// How long a bad config gets to be rejected before the harness calls it a hang.
+///
+/// Same measurement problem as [`STARTUP_TIMEOUT`] and the same reason for the
+/// size: this bounds process spawn plus config parse, and "still running" is
+/// read as a failure to fail fast. Too tight, and a slow spawn under parallel
+/// load is indistinguishable from a config error that was silently accepted —
+/// which is the bug these tests exist to catch.
+const FAILFAST_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// The shipped default keys the active-project pointer by whatever coordinate
 /// the caller has, so two harnesses in one project — or two operators on one
@@ -245,7 +271,7 @@ fn invalid_mode_value_fails_fast() {
     // want to catch in CI long before it reaches prod.
     let (exited_ok, all) = spawn_and_wait_for_exit(
         &[("AI_MEMORY_AUTO_SCOPE__MODE", "per_universe")],
-        Duration::from_secs(5),
+        FAILFAST_TIMEOUT,
     );
     assert!(
         !exited_ok,
@@ -317,10 +343,8 @@ fn empty_mode_string_fails_fast() {
     // `""` is not a valid serde enum variant. Just like
     // `per_universe`, this must abort startup — never quietly default
     // to single.
-    let (exited_ok, all) = spawn_and_wait_for_exit(
-        &[("AI_MEMORY_AUTO_SCOPE__MODE", "")],
-        Duration::from_secs(5),
-    );
+    let (exited_ok, all) =
+        spawn_and_wait_for_exit(&[("AI_MEMORY_AUTO_SCOPE__MODE", "")], FAILFAST_TIMEOUT);
     assert!(
         !exited_ok,
         "empty `mode` must NOT result in a successful startup.\nstderr:\n{all}"
@@ -335,7 +359,7 @@ fn pascalcase_mode_fails_fast() {
     // both forms (which would mask a future case-sensitivity bug).
     let (exited_ok, all) = spawn_and_wait_for_exit(
         &[("AI_MEMORY_AUTO_SCOPE__MODE", "PerSession")],
-        Duration::from_secs(5),
+        FAILFAST_TIMEOUT,
     );
     assert!(
         !exited_ok,


akitaonrails:
Thanks for this — but I can't merge it as submitted, and the reason is scope, not the idea.

The title is `test(autoscope): budget process spawn`, but the commit is a **feature**: it adds `V61__project_identity.sql` (new `identity` / `identity_source` columns + backfill), a new `ai-memory-core::repository_identity` module, the migration-pin bump, and the D1 identity-source ranking — with the autoscope test change riding alongside. That's the ARD-08 project-identity work, and its own migration comment says it "lands before authorization" — i.e. it's the foundation for #708, not a test tidy-up.

Three blocking issues:

1. **Wrong branch.** A new migration + schema surface is additive → it belongs on `release/2.2` (the 2.2.0 minor train), not `main` (the 2.1.x patch line). Landing a migration on main would turn the next main release into a minor and, worse, diverge the two lines' schema.
2. **Migration number collision.** `release/2.2` already has a `V61` (`project_abstract_embeddings`, #672) and a `V62` (#682). A second, different `V61` here would corrupt the migration sequence the moment the branches meet. On `release/2.2` this needs to be **V63**.
3. **Coupled to an unreviewed design.** Project identity is the access-control substrate for #708 ("multi-user servers have no per-project authorization"). Resolving a working dir to a `repository_id` decides who can read what, so the identity resolution and the authorization model need to be designed and reviewed together, not landed piecemeal under a test title.

Please split this: (a) if there's a genuine autoscope test-budget fix, send it on its own against `main`; (b) resubmit the project-identity feature against `release/2.2`, renumbered to **V63**, with an accurate `feat(identity):` title, and let's tie it to the #708 authorization design so the resolution rung ranking and the grant model land coherently. Happy to review the design first if you want to sketch it on #708.

abhisheksharma2411 (commented):
The timeout diagnosis is the best-evidenced flake report I've read on this repo — three runs failing 1, then 5, then 7 of 10, each landing at 8.0x seconds against an 8s budget, is not a hypothesis, it's a measurement. And the after-numbers make the case better than a pass would: 16.3s / 28.5s / 8.15s / 5.06s for identical work says the budget was measuring the machine. The 8.15s run a fifth of a second the wrong side of the old limit is the whole argument in one line.

Naming `FAILFAST_TIMEOUT` instead of three inline `Duration::from_secs(5)` is right for the same reason, and I appreciated the honesty that those three didn't fail locally — a downstream fork's larger binary prompting it is exactly the provenance a reviewer needs to weigh it.

## The branch carries more than the description says

> Notes: Tests only — no product code touched.

That isn't what's on the branch. Against its own merge-base — which **is** current `main`, 0 commits behind, so this isn't stale-base drift:

```
crates/ai-memory-cli/tests/suite/autoscope_env.rs           |  34 +-
crates/ai-memory-core/src/lib.rs                            |   2 +
crates/ai-memory-core/src/repository_identity.rs            | 644 +++++++++++++
crates/ai-memory-store/migrations/V61__project_identity.sql |  48 ++
crates/ai-memory-store/src/api_credentials.rs               |   2 +-
```

All of it in the single commit `b2fb97c4 test(autoscope): budget process spawn, not the product`. Neither new file exists on `main`, so they're genuinely introduced here.

And they are not incidental. `V61__project_identity.sql` adds a project `identity` column, and its own header says why it matters:

> Resolving by folder name means an `api/` checkout from one organisation resolves to the same row — and so the same grant — as an unrelated `api/` from another. **That is an access-control hole, not untidiness.**

`repository_identity.rs` implements the FR-06 identity chain behind it, and `lib.rs` exports `MARKER_FILENAME` / `MARKER_FILENAMES`.

That's a schema migration and a 644-line core module arriving under a `test(...)` type with "no product code touched" in the notes. Whoever reviews this on the strength of the description reviews a 34-line timeout change and merges an unreviewed migration — which is a worse outcome for the identity work than for this one, since it deserves its own review against ARD-08 rather than being carried along.

Almost certainly a branch-hygiene accident rather than intent — the two bodies of work read as different sittings. **Splitting them would let this land today**: the timeout change is self-contained, well-evidenced, and I'd take it as-is.

## One small thing on the constants, once it's split

Both now carry the reason they're large, which is the right instinct — "so the next person to see 45s does not tighten it back" is the comment that earns its place. Worth stating the *inverse* bound too: at 45s a real startup regression has to be roughly 5× before it trips. You say as much in the description ("it just has to be a large one"), and that sentence is more useful in the source than in the PR body, since the PR body isn't what the next person reads.

luisfnicolau:
@akitaonrails @abhisheksharma2411  you are right, and the branch hygiene failure is mine. I've split it.

**This PR is now test-only**: one file, `crates/ai-memory-cli/tests/suite/autoscope_env.rs`,
+31 −7. I rebuilt the branch from current `main` (it was 48 commits behind) and
applied only the timeout change, so all the project identity code are no
longer here. The 10 autoscope tests pass against current `main`.

@abhisheksharma2411 — the inverse bound is now in the source rather than the PR
body, with the numbers behind it:

> A real startup regression still fails here. It just has to be a large one,
> which is the correct trade for a test whose failure would otherwise be read
> as noise. Concretely, the measured spread while writing this was 5.06s on a
> quiet machine and 28.5s under parallel load, so 45s trips on roughly a 9x
> regression against a quiet run and a 1.6x one against the worst observed.
> That is the inverse bound: anything subtler than that belongs in a
> benchmark, not here.

You were also right that this was an accident rather than intent — two
sittings, one commit, and I wrote the description from what I meant to send
instead of checking what the commit actually contained.

@akitaonrails — I accept all three points, including the third. Project identity
decides which row a working directory resolves to, so it decides who can read
what; reviewing it apart from the authorization model it exists to support would
be reviewing half a design.
Design sketch posted on #708; I'll resubmit identity against whichever branch you
name there, renumbered. 
Note `V63__page_evidence.sql` has since landed, so
identity needs `V64` rather than `V63`.

akitaonrails:
Quick update since the release landscape moved and my earlier guidance is now stale: **2.2.0 shipped** (main is the 2.2.x line), new features go to the **`release/2.3`** branch, and the schema is at **V63** — so the project-identity migration here needs to be **V64 on `release/2.3`**, not V61/V62 (both taken). The core points from my earlier review still stand: this PR is a `feat` (new migration + `repository_identity` module + the ARD-08 identity work), not the `test(autoscope)` its title claims, and it should be split — (a) any genuine autoscope test-budget change on its own against `main`, and (b) the project-identity feature against `release/2.3` as V64, tied to the #708 authorization design (identity is that feature's access-control substrate). Happy to review the design on #708 first. Leaving this open for the resubmission.

abhisheksharma2411 (commented):
Re-reviewed. The split is clean and the branch is what the description now says.

```
merge-base:  3 commits behind origin/main (fetched, not a stale local ref)
PR commits:  1
files:       crates/ai-memory-cli/tests/suite/autoscope_env.rs | 31 +, 7 -
```

`repository_identity.rs`, `V61__project_identity.sql`, the `lib.rs` exports and the `api_credentials.rs` touch are all gone. Suite green here:

```
test result: ok. 10 passed; 0 failed; finished in 1.87s
```

## The inverse bound reads exactly right where it now lives

> the measured spread while writing this was 5.06s on a quiet machine and 28.5s under parallel load, so 45s trips on roughly a 9x regression against a quiet run and a 1.6x one against the worst observed

That is the sentence a future reader needs, and it answers the question the number provokes ("45s, really?") before they go looking for the PR. Naming the 1.6x explicitly is the honest half — it says plainly that under worst-observed load this budget is not far from the noise floor, rather than implying 45s is comfortably clear of everything.

One data point to add to your spread, since it happens to bracket it from the other side: my run above took **1.87s for all ten**, on a machine where the binary was already built and warm. Same tests, same count, an order of magnitude under your quiet-machine figure — which is your thesis rather than a contradiction of it. What varies is spawn, link and page-in, and a warm page cache collapses it. The old 8s budget sat inside that variance; 45s sits outside it.

## On the identity half

Noting @akitaonrails' guidance above so it isn't lost when you resubmit: schema is at **V63**, so it wants to be **V64**, and it targets **`release/2.3`** rather than `main`, tied to the #708 authorization design. The migration comment you'd written — that resolving by folder name collapses two organisations' `api/` checkouts into one grant — is the strongest argument in that change, and it belongs in front of the #708 reviewers rather than trailing a timeout fix.

Nothing further from me on this one.

luisfnicolau:
Thanks for the updated guidance — that settles the target, and it matches what I
found re-checking today: `release/2.3` is 55 commits ahead of my old base and the
schema is at V63, so identity is V64 and the `memory_grant` migration behind #708
would be V65.

On the split: it's already pushed here, from before your update landed. Current
head is `fa53b2b1` — one commit, one file,
`crates/ai-memory-cli/tests/suite/autoscope_env.rs`, +31/−7. The migration, the
`repository_identity` module, the `lib.rs` exports and the `api_credentials.rs`
pin bump are all gone from this branch; @abhisheksharma2411's re-review above
confirms it independently.

So this PR is now. Would you like it merged as-is, or closed in favour of a
fresh one against `main`? Happy either way — I only want to avoid it sitting open
as a resubmission that already happened.

akitaonrails:
Merged as-is into `main` (85a3a53b) and synced into `release/2.3` — the split is exactly right: one file, test-only, feature parts gone. The inverse-bound reasoning holds (these tests spawn ~10 debug binaries at once, so the budget measures harness load, not product startup; a large real regression still trips it). Test-only, so no changelog. Thanks for splitting it cleanly. The project-identity V64 + `memory_grant` work now belongs with the #708 authorization design, which I've accepted for the 2.3 line — happy to review that design when you open it.

Issue #659, opened 2026-09-06
Feature request: first-class Muse Code support (secure MCP + managed harness via MSP)


rjsoux:
# Feature request: first-class Muse Code support, including secure MCP and a managed harness via MSP

Hi Akita,

Before the technical request, I would like to express my sincere admiration for who you are and, especially, for the consistent care you show toward helping other people learn, think independently, and evolve. I wish you continued success in your businesses and, above all, a fulfilling and fortunate life.

Thank you for creating and continuing to evolve ai-memory. It has become an important part of my real development workflow, not just an experiment.

## Problem this would solve

Would you consider adding first-class support for **Muse Code**, ideally in two layers?

1. A secure, officially installed MCP integration (`install-mcp --client muse`) with the standard ai-memory routing skills/instructions.
2. If Muse's native contract is sufficient, a managed harness (`ai-memory run muse`) that provides the same portable-workstream continuity already available for Claude Code, Codex, and the other managed agents.

I currently manage **dozens of software projects** across four coding-agent environments:

- Claude Code;
- Codex;
- Cursor Agent; and
- Muse Code.

My environment spans an Apple Silicon Mac and a Linux development VPS, with multiple users and explicit workspaces/projects. A single central ai-memory 2.0.3 server provides cross-project and cross-agent memory. Claude Code, Codex, and Cursor are already working successfully with ai-memory through their officially supported integration paths.

Muse Code has recently become particularly valuable for my less demanding development sessions because the **Muse Spark 1.3 Contributor** model offers an unusually strong quality-to-cost ratio. In the model catalog available in my installation on September 6, 2026, `meta/muse-spark-1.3-contributor` was listed at USD 0.10/M input tokens, USD 0.20/M output tokens, and USD 0.002/M cached tokens, with a 1,007,997-token context window. Compared with the standard Muse Spark 1.3 entry in that same catalog, this was approximately 12.5 times cheaper for input, 21.25 times cheaper for output, and 75 times cheaper for cached tokens. I understand that Contributor mode may allow content, including inter-session messages, to be used for product improvement; I deliberately choose it only where that tradeoff is appropriate.

Because of that combination of quality, context size, and cost, Muse is likely to become a regular fourth environment in my workflow. The missing piece is reliable ai-memory continuity.

## Current result and limitation

I have a **working but partial** integration with Muse Code 1.0.3 (`1.0.3-R2198.1`):

- Muse connects to ai-memory's Streamable HTTP MCP endpoint.
[...]

PR #711, opened 2026-09-10
feat(install-mcp): add Muse Code as an MCP-only client

felipe-NR:
## What changed

New `install-mcp --client muse` registers ai-memory with Meta's Muse Code. It
merges a native streamable-HTTP entry into the snake_case `mcp_servers` map of
`~/.config/muse/settings.json`, preserving sibling servers:

```json
{
  "schema_version": 1,
  "mcp_servers": {
    "ai-memory": {
      "transport": "streamable_http",
      "url": "http://127.0.0.1:49374/mcp",
      "headers": { "Authorization": "Bearer <token>" },
      "enabled": true,
      "mode": "optional"
    }
  }
}
```

No behaviour changes for any existing client. No changed defaults.

## Why

Closes the MCP-only first increment from #659. Every field traces to Meta's
published docs rather than to an installed build, per the verification you asked
for in that thread:

- **`headers` on `streamable_http` is documented**
  ([extending#mcp](https://dev.meta.ai/docs/muse-code/extending#mcp)), so bearer
  auth works through the official contract. The credential-forwarding stdio
  bridge floated in the issue is unnecessary, and so is the reporter's
  server-wide `--allow-insecure-no-auth` workaround.
- **`"schema_version": 1` is mandatory**
  ([configuration#settings-file](https://dev.meta.ai/docs/muse-code/configuration#settings-file)):
  a settings file that omits it fails *every* `muse` command at startup with
  `malformed settings file`. The writer adds it when absent and never rewrites
  an existing value, so a future schema 2 is not silently downgraded. This is
  the sharpest edge in the integration — a careless writer breaks the user's
  agent, not just their memory server.
- **`"mode": "optional"` is written explicitly.** Muse defaults `mode` to
  `required`, and a required server that fails to start aborts the whole run.
  ai-memory augments a session rather than gating it, so an unreachable server
  should cost recall, not the coding session. This one is a judgement call
  rather than a documented requirement — happy to flip it if you'd rather
  inherit Muse's default.
- **No `framing` key**: a non-default `framing` on `streamable_http` fails
  Muse's own validation.
- **snake_case `mcp_servers`** is why this needed a new `JsonMcpLocation`
  variant instead of reusing `RootMcpServers`; the camelCase spelling most
  clients use is ignored by Muse.

**Scope: MCP-only.** Muse documents a lifecycle hook surface, but the output
contract of its `SessionStart` event is not specified, so capture and managed
workstreams are not claimed and no `AgentKind` variant or migration is added —
matching Swival, Zed, and vscode-copilot. The managed-harness half of #659 stays
design-doc-first, as you asked.

**Skills need no new code.** Muse loads user skills from `~/.agents/skills`
([extending#skills](https://dev.meta.ai/docs/muse-code/extending#skills)), which
`install-skills` already writes. Documented in the support matrix rather than
built.

## Test plan

- [x] `cargo test --workspace --all-targets` passes — 878 passed, 0 failed
      (nextest not installed on this box, so the CI-equivalent command was used)
- [x] `cargo deny check` — advisories ok, bans ok, licenses ok, sources ok
- [x] Manual test: `--apply` against a populated `settings.json` with a sibling
      `my-tools` stdio server and a `telemetry` block — the sibling server,
      `telemetry`, and `schema_version` all survive; a re-run reports `no-op`.
      Then `uninstall --only mcp --apply` removes only the ai-memory entry and
      leaves the rest of the file intact and valid.

Four focused unit tests cover the entry shape, `schema_version` added when
absent, an existing `schema_version` never rewritten, and the rendered snippet
carrying the key for hand-merging.

## Commit attribution

## Release impact

- [x] **Minor** — additive: new `install-mcp` client, new agent harness entry.

## CHANGELOG (merge gate)

- [x] I added a `CHANGELOG.md` `[Unreleased]` entry under `### Added` (#659).
- [x] No default changed. The `mode: "optional"` choice applies only to the
      entry ai-memory writes for this new client.

Docs updated in the same commit: new `docs/mcp-install.md` section, a
`docs/support-matrix.md` row, and the README support matrix.

## Notes for reviewers

**One thing worth your scrutiny beyond the diff.** The `uninstall` sweep
iterates a hand-maintained array of `McpClient` values that the compiler does
not check (`uninstall.rs`, the `for client in [...]` list). A new variant
compiles clean everywhere else while silently leaving its entry orphaned in the
user's config on uninstall. I hit it here and added `Muse` to that array, but
the trap is standing for whoever adds the next client — the same shape as the
`AgentKind::ALL` note in AGENTS.md about the Zero integration. Happy to send a
separate PR making it exhaustive if you want it closed properly rather than
patched per-client.

**On MSP.** I could not corroborate the `muse serve` / MSP surface from any
published Meta document; details are in a comment on #659. Nothing in this PR
depends on it.


akitaonrails:
Merged onto `release/2.2` for the 2.2.0 release (commit `817cde17`). Adding an MCP client is additive, so it rides the 2.2 feature train rather than the 2.1.x patch line. This is the MCP-only first increment from #659; verified green on the merged branch (rustfmt, clippy `-D warnings`, full workspace test — including your four `muse_*` cases). GitHub can't auto-mark this merged because the PR targets `main` and the commit landed on `release/2.2`; closing manually. Thanks — clean, well-documented work, and the schema_version handling is exactly the sharp edge to have gotten right.

Issue #701, opened 2026-09-10
purge-session: a page file left behind by a failed cleanup is reindexed by the next reconcile tick


samirhvbr:
**Version**

Output of `ai-memory --version`: verified against source at `55fcce35` (`main`, workspace version `2.1.1`).

This is **pre-existing on `main` and not caused by #696** — the repro below runs on `55fcce35`, which is the merge base of that PR. #696 narrows the window; it does not close this one.

**What happened**

`purge-session` reports partial failure when it cannot remove the session's page file, and returns `files_failed` with the path. The database rows are gone at that point, but `sessions/<id>.md` is still on disk. Within one reconcile tick — `RECONCILE_INTERVAL` is 30s (`crates/ai-memory-wiki/src/watcher.rs:35`) — the watcher indexes that leftover file again and the purged session's body is searchable once more.

Nothing on the reindex path consults the session tombstone. `purged_sessions` is written inside the purge transaction (`crates/ai-memory-store/src/ops.rs:3564`) and read only by the ingest guard (`session_is_purged`, `ops.rs:1223`), whose comment states it is "the one place the guard has to live" — true for ingest, but the wiki reindex path is a second way rows come back.

The scope-level equivalent exists and works: `reindex_all` skips a tombstoned scope via `scope_is_purged` (`crates/ai-memory-wiki/src/wiki.rs:1563`). There is no session-level counterpart. A session purge does not touch the project row, so `reconcile`'s orphan guard (`watcher.rs:347`) passes and the directory is walked normally.

This is not only the crash case. It is reached by:

- a cleanup failure — the state `purge_session_reports_file_cleanup_failure_after_db_commit` already produces on purpose and asserts we report as `files_failed`;
- by inspection (not exercised below), a client that disconnects or times out after `WriterHandle::purge_session` has enqueued the command: the writer still runs the purge and logs the dropped receiver, while the handler that would have removed the file is gone.

**What I expected**

A purge is a promise that the content is gone. After `purge-session` returns — including when it returns `files_failed` — no later reindex should make that session's page searchable again. Either the reindex path declines to index a page whose session is tombstoned, or the leftover file is otherwise made inert, the way a tombstoned scope's directory already is.

**Steps to reproduce**

Against a running server:

1. Run `ai-memory serve` with the wiki watcher active, and let a session end so it consolidates to `<data_dir>/wiki/<ws>/<proj>/sessions/<id>.md`. Confirm the page is searchable.
[...]

PR #717, opened 2026-09-11
feat(llm): add Codex account-aligned provider

### LINKED ISSUE #716: Add a Codex LLM provider that follows the Codex CLI account
**Problem this would solve**

ai-memory currently supports OpenAI API keys and its own `openai-oauth`
credential lifecycle, but it cannot reuse the account selected by an existing
Codex CLI login as an independent LLM provider.

This becomes particularly important for users who work with more than one
Codex account or subscription.

When Codex and ai-memory maintain independent authentication state, switching
accounts in the Codex environment does not necessarily switch the account used
by ai-memory. Background consolidation or other LLM operations may continue
consuming the limits of a different account until that account is exhausted.
The user may only discover the mismatch when ai-memory suddenly has no
available LLM.

The desired behavior is for ai-memory to follow the Codex credential selected
for its environment, including the applicable `CODEX_HOME`, and reload that
credential before every operation. Account selection then remains aligned with
the Codex CLI instead of becoming a second, easily forgotten ai-memory setting.

**Proposed solution**

Add an independent `codex` LLM provider with the following behavior:

- Support `AI_MEMORY_LLM_PROVIDER=codex`.
- Use `gpt-5.6-luna` as its default model.
- Read only `tokens.access_token` and `tokens.account_id` from the Codex-owned
  `auth.json`.
- Resolve that file from:
  1. `$CODEX_HOME/auth.json` when `CODEX_HOME` is non-empty.
  2. The platform home directory under `.codex/auth.json`.
- Reload the selected auth file before every LLM operation so an account change
  made through Codex is observed without maintaining a second ai-memory login.
- Never materialize `refresh_token` or `id_token`.
- Never copy, rewrite, or otherwise take ownership of the Codex credential.
- Delegate renewal to `codex app-server --stdio`.
- Allow overriding the executable with `AI_MEMORY_CODEX_EXECUTABLE`.
- Support text and JSON Schema structured Responses calls.
- Preserve the existing `openai-oauth` login, storage, and refresh behavior.

On the first HTTP 401, the provider should reload the auth file and retry if an
external process has already rotated the token. Otherwise it should serialize
recovery through a provider-local mutex, ask `codex app-server --stdio` to
refresh the Codex-owned credential, require the access token to change, and
retry the Responses request once.

The app-server protocol and process execution should be bounded:

- 256 KiB per JSONL line.
- 1 MiB cumulative stdout.
- 64 KiB stderr capture without reproducing it verbatim.
- Recovery timeout capped at 30 seconds.
- Process cleanup on success, failure, timeout, or cancellation.
- No retry loop after a second 401.

Initial credential-store support would be limited to file-backed Codex auth.
Keyring and ephemeral stores should fail with an actionable message.

**Design precedent**

This provider is inspired by the authentication model used by
[ai-usagebar](https://github.com/akitaonrails/ai-usagebar), maintained by the
same developer as ai-memory.

ai-usagebar reuses authentication maintained by the official Codex CLI and
supports multiple Codex accounts through account-specific auth paths and
`CODEX_HOME`. That avoids requiring users to paste or maintain another
credential merely to observe their Codex usage.

The proposed ai-memory provider follows the same account-alignment principle,
but uses a narrower credential boundary appropriate for LLM requests:

- ai-memory materializes only `access_token` and `account_id`;
- it does not parse or own the refresh token;
- renewal is delegated to `codex app-server`;
- it reloads the selected auth file before each operation.

**Alternatives considered**

1. **Continue using `openai-oauth`.** This works, but creates a separate
   credential lifecycle. With multiple accounts, it is easy for Codex and
   ai-memory to remain authenticated as different users and consume different
   limits without making that mismatch obvious.
2. **Require an explicit account selector inside ai-memory.** This would create
   another account setting that users need to keep synchronized with
   `CODEX_HOME`.
3. **Read and refresh the Codex refresh token directly.** This would couple
   ai-memory to private credential details and duplicate Codex's authentication
   responsibilities.
4. **Call `codex exec` for each completion.** This would introduce agent
   execution behavior instead of using ai-memory's typed LLM provider boundary.
5. **Copy Codex credentials into ai-memory's data directory.** This would
   duplicate sensitive state and introduce synchronization and ownership
   problems.

**Would this break existing installs?**

No. The provider, environment variable, executable override, and CLI option are
additive. Existing defaults and the `openai-oauth` provider remain unchanged.

This is minor-version functionality because it adds a new LLM provider and
configuration surface.

**Additional context**

A local implementation has been validated on Windows using both text and
structured calls with `gpt-5.6-luna` and reasoning effort `medium`.

The Codex auth file was hashed before and after both smoke tests and remained
byte-for-byte unchanged. `codex login status` also remained authenticated.

Recovery tests use a credential-free native Rust fake executable and cover
token rotation, unchanged tokens, concurrent 401 responses, JSON-RPC
validation, output limits, timeouts, premature exits, and process cleanup.

Docker support is intentionally outside the initial scope because the Codex
executable and its credential file must be available in the same environment
as ai-memory.

gabrielscharb:
## What changed

- Added an independent `codex` LLM provider with `gpt-5.6-luna` as its default model.
- Added `AI_MEMORY_LLM_PROVIDER=codex` and the optional `AI_MEMORY_CODEX_EXECUTABLE` override.
- Resolved Codex authentication from non-empty `CODEX_HOME` or the platform `.codex/auth.json`.
- Reloaded the selected Codex auth file before every operation, allowing account changes made through Codex to be observed without a separate ai-memory account selector.
- Materialized only `access_token` and `account_id`; the provider never reads refresh or ID tokens into its credential types and never writes the auth file.
- Delegated expired-token recovery to a bounded `codex app-server --stdio` JSON-RPC exchange.
- Added one-retry 401 handling with external-rotation detection and serialized concurrent recovery.
- Extracted the shared Responses HTTP transport used by `codex` and `openai-oauth` without changing the latter's login, storage, or refresh semantics.
- Preserved the current operator-configured header and fallback-chain surfaces while rebasing onto v2.1.2.
- Added text and structured support to `ai-memory llm-test --provider codex`.
- Updated provider, installation, architecture, eval, changelog, and testing documentation.

## Why

Users working with multiple Codex accounts should not need to maintain a second, independent account selection for ai-memory.

With separate authentication state, a user can switch the Codex CLI to one account while ai-memory silently continues consuming another account's limits. The mismatch may only become visible after the latter account is exhausted and ai-memory loses access to its LLM.

This provider follows the auth file selected by the same `CODEX_HOME`, reloads it before every operation, and leaves login and renewal ownership with Codex. It aligns ai-memory with the user's Codex environment without copying credentials or introducing another account selector.

The approach was inspired by [ai-usagebar](https://github.com/akitaonrails/ai-usagebar), which is maintained by the same developer and already demonstrates the value of reusing official Codex CLI authentication for users with multiple accounts. This implementation uses a narrower boundary: only the access token and account ID are materialized, while refresh remains delegated to Codex.

Closes #716.

## Test plan

- [x] `TAILWIND_SKIP=1 cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `TAILWIND_SKIP=1 cargo test --workspace` passes
- [x] Manual test: text and JSON Schema structured smokes with `gpt-5.6-luna` and reasoning effort `medium` both passed
- [x] `cargo deny --all-features check` passes
- [x] Existing `openai-oauth` regression suite passes
- [x] Standalone companion importer suite passes

The complete workspace test reached two pre-existing `ai-memory-core::routing_skills` failures caused by CRLF bytes in this Windows checkout:

- `managed_skill_payloads_use_lf_line_endings`
- `skill_frontmatter_is_valid_and_matches_metadata`

Neither the embedded skill assets nor those assertions are changed by this PR. A complementary workspace run excluding `ai-memory-core`, followed by all remaining `ai-memory-core` tests with only those two assertions skipped, passed.

After rebasing onto v2.1.2, the focused Codex suite has 10 passing tests, the existing `openai-oauth` suite has 18 passing tests, and the Codex-related CLI/config selection has 20 passing tests. Exact coverage could not be measured because `cargo llvm-cov` was unavailable locally; parser, HTTP, structured output, retry, protocol limits, native process recovery, unchanged tokens, failure cleanup, and concurrent recovery are directly exercised.

The live Codex auth-file SHA-256 was identical before and after both smokes, and `codex login status` remained authenticated.

## Commit attribution

Every commit is attributed to `Gabriel Scharb <gabrielscharb@gmail.com>`.

## Release impact

- [x] **Minor** — additive: new flag/subcommand/MCP tool/config key, new agent harness or provider
- [ ] **Major (breaking)** — on-disk format, removed/renamed surface, breaking MCP schema change — called out in "What changed" above

## CHANGELOG (merge gate)

- [x] I added a `CHANGELOG.md` `[Unreleased]` entry — **required** for any user-facing change: new flag / env var / endpoint / MCP tool / marker key, changed behaviour, or an observable bug fix. (Exempt only for internal refactors, dead-code removal, and test-only churn.)
- [x] Any changed **default** (flag behaviour, config default, env var, response shape) is called out explicitly in "What changed" above.

## Notes for reviewers

The main review-sensitive boundary is credential and account ownership:

- The provider follows the auth file selected by `CODEX_HOME`; it does not introduce an independent ai-memory account selector.
- The file is reloaded before every operation so an account change is observed.
- `CodexAuth` contains only resolved paths.
- The minimal auth parser declares only `access_token` and `account_id`.
- Secrets are wrapped in `SecretString` and excluded from debug and error output.
- ai-memory never writes the Codex auth file.
- Renewal remains entirely owned by `codex app-server`.

This is inspired by ai-usagebar's reuse of official Codex CLI authentication, especially its multiple-account use case, but intentionally does not copy its direct refresh-token implementation.

Recovery accepts only the expected JSON-RPC response IDs, ignores notifications without IDs, bounds stdout/stderr, caps recovery at 30 seconds, and terminates the child process on every exit path.

The initial version supports file-backed Codex auth. Keyring and ephemeral credential stores are intentionally unsupported. Docker requires the Codex binary and credential file to exist inside the same environment.

Please pay particular attention to account-switch behavior, fallback-chain/native-auth interaction, the shared Responses transport extraction, Windows child-process cleanup, and the guarantee that `openai-oauth` semantics remain unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -501,6 +501,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   loss). A per-path async lock now serializes same-path writers while
   different paths still proceed concurrently; batches acquire their paths in a
   fixed global order so they cannot deadlock.
+
 ## [2.0.3] - 2026-09-04
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Added an independent `codex` LLM provider that follows the Codex CLI account
+  selected by `CODEX_HOME`, reloads its read-only `auth.json` credentials before
+  each operation, and delegates expired-token recovery to
+  `codex app-server --stdio` (#716).
+
 ## [2.1.2] - 2026-09-11
 
 ### Changed
@@ -495,7 +501,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   loss). A per-path async lock now serializes same-path writers while
   different paths still proceed concurrently; batches acquire their paths in a
   fixed global order so they cannot deadlock.
-
 ## [2.0.3] - 2026-09-04
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -291,7 +291,8 @@ The full model is in [`docs/security.md`](docs/security.md),
 
 Optional. Everything works with zero LLM calls; adding a provider
 upgrades session summaries and enables semantic search. Anthropic,
-OpenAI (incl. OAuth/Codex), GitHub Copilot, Gemini, OpenCode (Go and Zen), and
+OpenAI (including OAuth), Codex CLI credential reuse, GitHub Copilot, Gemini,
+OpenCode (Go and Zen), and
 any OpenAI-compatible endpoint (Ollama, LM Studio, vLLM) are supported
 for consolidation; OpenAI, Voyage, Gemini, and keyless OpenAI-compatible
 endpoints for embeddings. Configuration lives in
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -1909,6 +1909,9 @@ pub struct LlmTestArgs {
     /// Prompt to send.
     #[arg(long)]
     pub prompt: String,
+    /// Request a small JSON-schema response instead of plain text.
+    #[arg(long)]
+    pub structured: bool,
     /// Base URL override (required for openai-compat).
     #[arg(long)]
     pub base_url: Option<String>,
@@ -3346,6 +3349,28 @@ mod tests {
         assert!(matches!(args.provider, LlmProviderChoice::AnthropicOauth));
         assert_eq!(args.model, "claude-sonnet-4-6");
         assert_eq!(args.prompt, "ping");
+    }
+
+    #[test]
+    fn llm_test_codex_structured_parses() {
+        let cli = Cli::try_parse_from([
+            "ai-memory",
+            "llm-test",
+            "--provider",
+            "codex",
+            "--model",
+            "gpt-5.6-luna",
+            "--prompt",
+            "ping",
+            "--structured",
+        ])
+        .unwrap();
+
+        let Command::LlmTest(args) = cli.command else {
+            panic!("expected llm-test command");
+        };
+        assert!(matches!(args.provider, LlmProviderChoice::Codex));
+        assert!(args.structured);
     }
 
     #[test]
--- a/crates/ai-memory-cli/src/cli.rs
+++ b/crates/ai-memory-cli/src/cli.rs
@@ -1696,6 +1696,8 @@ pub enum LlmProviderChoice {
     OpenaiCompat,
     /// OpenAI ChatGPT/Codex OAuth backend.
     OpenaiOauth,
+    /// Reuse Codex CLI authentication and delegated refresh.
+    Codex,
     /// GitHub Copilot Chat backend.
     Copilot,
     /// OpenCode cloud API (Go by default; AI_MEMORY_LLM_BASE_URL selects Zen).
--- a/crates/ai-memory-cli/src/commands/llm_test.rs
+++ b/crates/ai-memory-cli/src/commands/llm_test.rs
@@ -88,6 +88,14 @@ mod tests {
     }
 
     #[test]
+    fn codex_choice_maps_to_runtime_provider() {
+        assert_eq!(
+            ProviderChoice::from(LlmProviderChoice::Codex),
+            ProviderChoice::Codex
+        );
+    }
+
+    #[test]
     fn llm_test_exercises_pipeline_sampling_compatibility() {
         let request = representative_request("diagnostic".into());
 
--- a/crates/ai-memory-cli/src/commands/llm_test.rs
+++ b/crates/ai-memory-cli/src/commands/llm_test.rs
@@ -69,6 +69,7 @@ impl From<LlmProviderChoice> for ProviderChoice {
             LlmProviderChoice::Gemini => Self::Gemini,
             LlmProviderChoice::OpenaiCompat => Self::OpenAiCompat,
             LlmProviderChoice::OpenaiOauth => Self::OpenAiOAuth,
+            LlmProviderChoice::Codex => Self::Codex,
             LlmProviderChoice::Copilot => Self::Copilot,
             LlmProviderChoice::Opencode => Self::OpenCode,
         }
--- a/crates/ai-memory-cli/src/commands/llm_test.rs
+++ b/crates/ai-memory-cli/src/commands/llm_test.rs
@@ -36,10 +36,24 @@ pub async fn run(config: &Config, args: LlmTestArgs) -> Result<()> {
         model = client.model(),
         "sending prompt",
     );
-    let resp = client
-        .complete(representative_request(args.prompt))
-        .await
-        .context("calling provider")?;
+    let request = representative_request(args.prompt);
+    if args.structured {
+        let value = client
+            .complete_structured_raw(
+                request,
+                serde_json::json!({
+                    "type": "object",
+                    "properties": {"answer": {"type": "string"}},
+                    "required": ["answer"]
+                }),
+            )
+            .await
+            .context("calling provider for structured output")?;
+        println!("--- model: {} ---", client.model());
+        println!("{}", serde_json::to_string_pretty(&value)?);
+        return Ok(());
+    }
+    let resp = client.complete(request).await.context("calling provider")?;
 
     println!("--- model: {} ---", resp.model);
     if let Some(u) = resp.usage {
--- a/crates/ai-memory-cli/src/config.rs
+++ b/crates/ai-memory-cli/src/config.rs
@@ -2810,6 +2810,48 @@ mod tests {
     }
 
     #[test]
+    fn codex_provider_uses_codex_home_auth_and_default_model() {
+        let tmp = TempDir::new().unwrap();
+        let codex_home = tmp.path().join("custom-codex-home");
+        let cfg = Config {
+            llm_provider: Some("codex".into()),
+            runtime_env: RuntimeEnv {
+                codex_home: Some(codex_home.clone()),
+                codex_executable: Some(PathBuf::from("codex-custom")),
+                ..RuntimeEnv::default()
+            },
+            ..Config::default()
+        };
+
+        let provider = cfg.llm_provider_config().unwrap().unwrap();
+        let auth = provider.auth.require_codex_auth().unwrap();
+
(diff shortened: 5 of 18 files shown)


abhisheksharma2411 (commented):
Reviewed the credential handling closely, since "reuse another tool's auth file" is where a provider integration usually goes wrong. **It holds, and it's tested rather than asserted** — which is the part I want to call out specifically.

## The security claims check out

> Materialized only `access_token` and `account_id`; the provider never reads refresh or ID tokens into its credential types

Verified three ways rather than taken from the description:

- `CodexAuth` carries **only paths** — `auth_file` and `executable`. No secret lives in the resolved-auth type at all.
- The materialized credential struct has exactly `access_token` and `account_id` (`codex.rs:41-42`, `:47-48`).
- Grepping every added line in the PR for `refresh_token|id_token` returns **only** a test fixture, setting both to the literal `"must-not-be-materialized"`. A negative claim with a fixture that would fail if it stopped being true is worth more than the sentence in the description.

`impl Debug for CodexCredentials` printing `CodexCredentials([REDACTED])` is the other half — the failure mode for this kind of integration is a token in a log line, and the type can't produce one.

On "never writes": `codex.rs` touches the auth file through `std::fs::read` only; the `write_json_line` calls target the subprocess stdin, not the file. Reloading before every operation so account changes made in Codex are observed — without ai-memory owning an account selector — is the right division of responsibility.

## One thing that isn't the feature

`crates/ai-memory-wiki/` is in the diff:

```
crates/ai-memory-wiki/src/watcher.rs | 55 ++++++++++++-----
crates/ai-memory-wiki/src/wiki.rs    | 24 +++++++-----
```

That's a performance change to the purged-session check from #701 — gating the `purged_sessions` round trip on `session_id_for_page(&page_path)` so an ordinary edit to `index.md` never queries. `session_id_for_page` doesn't exist on `main`, so it's genuinely introduced here, and the merge-base **is** current main so it isn't stale-base drift.

It looks like a good change on its own terms — the reasoning in the new comment about amortising over a directory versus a single file per event is sound, and "fails open, deliberately" is the right call for a watcher. But it has nothing to do with a Codex provider, and it changes behaviour on a path (#701, purge correctness) that someone reviewing an LLM-provider PR will not be looking at.

Worth splitting so it gets read on its own merits. Same suggestion I left on #709, which carries an unrelated core module and a schema migration — if the two of you are branching off a shared tree, that may be where both are coming from.

## Question

`docs/testing/codex-provider.tdd.md` is new and the diff is large enough that I did not read the 917-line `codex.rs` line by line. Is the delegated-refresh path (spawning the Codex executable) covered by `fake_codex.rs` for the failure cases — refresh returning non-zero, or the auth file changing underneath a request — or only the happy path? That's the bit I'd want pinned before it ships, and it's the one thing I could not establish from the diff alone.

gabrielscharb:
Thanks for the careful review, @abhisheksharma2411 — especially for independently checking the credential boundary.

I pinned the delegated-recovery cases more explicitly in `76cfc1b6` / `11334763`:

- the Rust fake app-server now has an `exit-nonzero` mode that exits with status 17 before replying;
- the recovery test verifies that this is rejected as a sanitized premature-exit error;
- the focused test passed, as did the full `ai-memory-llm` suite (234 passed, 2 ignored) and workspace Clippy with `-D warnings`.

The auth-file-changing-underneath-a-request case was already covered by `provider_reloads_rotated_token_and_retries_only_once`: the first 401 rotates `auth.json`, the second request uses the reloaded token, and the configured executable is `must-not-run`, so delegated recovery cannot be invoked in that path. The second-401 test separately pins the one-retry ceiling.

I also rechecked the `ai-memory-wiki` concern against the exact review pair (`b8068766...5598d4eb`). That path has no diff, and `session_id_for_page` is already present in base commit `b8068766` (v2.1.2), via #701. The current PR files API likewise lists no `crates/ai-memory-wiki` files, so there is no wiki change in #717 to split. It may have been a transient comparison/cache issue in the displayed diff.

akitaonrails:
Merged onto `release/2.3` for the 2.3.0 feature line (commit `f81ecf97`) — a new LLM provider is additive, so it rides the feature train rather than a 2.2.x patch. Audited and verified: no new dependency; env reads stay in `Config::load` and the typed `ProviderAuth::codex` resolves before provider construction (invariants #1/#14); the credential boundary is narrow (only `access_token` + `account_id` materialized, never written, refresh delegated to `codex app-server --stdio`); and the shared Responses-transport extraction left `openai-oauth` behavior intact (its full regression suite passes). Full workspace test + clippy `-D warnings` green on the merged branch, including your codex recovery/concurrency tests. GitHub can't auto-mark this merged because the PR targets `main` and the commit landed on `release/2.3`; closing manually. Not released yet — it ships when 2.3.0 is cut. Thanks — clean, security-conscious work.

abhisheksharma2411:
You're right and I was wrong — **the wiki finding was my error, not a display glitch**, and I'd rather name the cause than let it stand as "transient".

My local `origin/main` was three weeks stale (`0e47a7cf`, from #706). I ran `git grep session_id_for_page origin/main`, got nothing, and concluded the symbol was introduced here. After actually fetching:

```
$ git grep -c session_id_for_page b8068766 -- crates/ai-memory-wiki/
b8068766:crates/ai-memory-wiki/src/watcher.rs:1
b8068766:crates/ai-memory-wiki/src/wiki.rs:2

$ gh api /repos/akitaonrails/ai-memory/pulls/717/files | filter crates/ai-memory-wiki
0 files
```

Present in your base, and #717 touches zero files under that crate. Exactly as you said. Sorry for the detour — you spent time disproving a claim I should have verified before making.

The lesson is mine to fix: compare against a *freshly fetched* base ref, never a local one that happens to be lying around. This is the third time the same stale-ref mistake has bitten me in a week; the other two I caught before publishing, and this one I didn't.

**On the substance — the recovery coverage is exactly what I was asking for.** The `exit-nonzero` mode exiting 17 before replying, verified as a sanitized premature-exit error, is the failure case that matters: a delegated refresh that dies mid-handshake is where a provider integration leaks a raw subprocess error into a user-facing message.

And thank you for pointing at `provider_reloads_rotated_token_and_retries_only_once` for the other half — I'd missed that `must-not-run` as the configured executable is what makes it a *proof* rather than a scenario. The test can't accidentally pass by invoking delegated recovery, because invoking it would fail. That's the same trick as your `"must-not-be-materialized"` fixture, and both are better than an assertion that the thing didn't happen.

Congratulations on the merge onto `release/2.3`.

Issue #580, opened 2026-09-01
Generated TypeScript integrations do not spool: lifecycle events are dropped when the server is unreachable


aguirreSL:
**Version**

`ai-memory 1.30.0` (client and server). I re-checked `main` today and the code paths below are unchanged.

**What happened**

I worked a full day with the ai-memory server unreachable: the laptop was off the LAN where the server lives, so every delivery attempt got `ConnectionRefused`. Same machine, same repository, same time window, two agents running:

- **Claude Code**: events were spooled to `hook-spool/` and delivered on reconnect the next morning, then consolidated normally. Nothing lost.
- **OpenCode**: nothing reached the server, then or later. The sessions survived only inside OpenCode's own local SQLite (`opencode.db`), which is how I reconstructed them by hand afterwards.

Two full sessions of real work were lost to capture. There was nothing for `hook-drain` to drain, because nothing was ever written to disk for OpenCode.

**What I expected**

OpenCode lifecycle capture to survive an offline window the way Claude Code's does, and drain when the server comes back.

**Steps to reproduce**

1. Install hooks for both agents pointing at a server URL that is unreachable from the current network: `ai-memory install-hooks --agent claude-code --apply` and `ai-memory install-hooks --agent open-code --apply`.
2. Disconnect from the network where the server lives, then run a real session in each agent, in the same repository.
3. Reconnect and run `ai-memory hook-drain`.

The Claude Code session shows up on the server. The OpenCode session does not, and never will: its events were discarded at the moment each POST failed.

**Environment**

- OS: macOS (Darwin 25.5.0)
- Docker? Server yes, client no
- Agent: Claude Code and OpenCode, both on the same machine and repository
- Transport: http

**Relevant logs**

From the client log on the offline day (server address redacted):

```
2026-08-31T14:57:52Z ERROR rmcp::transport::worker: worker quit with fatal:
  ConnectError("tcp connect error", <server>:49374,
  Os { code: 61, kind: ConnectionRefused, message: "Connection refused" })
```

After reconnecting, `ai-memory status` showed the spool draining, but it only ever held the Claude Code events. No OpenCode event was present in `hook-spool/` at any point.

---

**Mechanism**

In `crates/ai-memory-cli/src/commands/install_hooks.rs`, the OpenCode plugin template's `drainHookQueue`:

- `hookQueue` is an in-memory array, and `HOOK_QUEUE_MAX = 100` drops the oldest entry when it is full.
[...]

PR #720, opened 2026-09-12
fix(hooks): spool an undelivered event in the POSIX script bundle

### LINKED ISSUE #719: Shell hook bundle drops lifecycle events when the server is unreachable: deliberate boundary for the docker path, or the gap #580 closed elsewhere?
**Version**

`ai-memory 2.2.0`. Every line below is cited against `2dc9e334` (the `v2.2.0`
tag on `main`), not from memory.

## The question

`#580` established that a capture path which POSTs and forgets is a defect:
*"Generated TypeScript integrations do not spool: lifecycle events are dropped
when the server is unreachable"*. You fixed it for all four generated
integrations in `6498ac57` (merged `73715b6f`).

The POSIX shell hook bundle has the same shape and did not get the same fix.
I do not know whether that is a deliberate boundary for the docker path or a
gap that was simply out of scope at the time, so I am filing it as a question
rather than a defect claim.

## Mechanism

`hooks/_lib.sh:459` posts and discards the result:

```sh
ai_memory_post_hook() {
    ...
        curl -s --max-time 0.5 -X POST "$1" \        # _lib.sh:462, :469, :474
```

and every one of the 80 call sites swallows the outcome —
`hooks/claude-code/post-tool-use.sh:20`:

```sh
    | ai_memory_post_hook "$SERVER/hook?event=post-tool-use&agent=claude-code${QS}" >/dev/null 2>&1 || true
```

Nothing in `hooks/` touches disk: `grep -rniE "spool|drain" hooks/` returns
nothing at `2dc9e334`. An unreachable server or a 5xx therefore costs the
event, silently, with no local record for `hook-drain` to find later.

## Why this path in particular

It is the one the documented docker install wires. `bin/ai-memory:575` forces
the script platform:

```sh
export AI_MEMORY_HOOK_PLATFORM="${AI_MEMORY_HOOK_PLATFORM:-posix}"
```

and `render_shared.rs:1302` says why — *"defaults to the shell command because
setup-agent copies scripts, not a host-local native binary"*, with
`render_shared.rs:1195` giving the same reason for the wrapper: *"the host has
no local binary"*.

So a reader who follows the docker quick start gets the one capture path with
no durability, and nothing at the point of choice says so. The property is
documented three times as a virtue of the paths that have it — most explicitly
at `docs/install.md:1445`, for the generated integrations: *"so a day of laptop
work off the server's network is captured, not silently dropped"* — and never
as a caveat on the one that lacks it.

## The two readings

**Deliberate boundary.** Invariant 5 declares hooks fire-and-forget and
bounded, and a fire-and-forget policy accepts bounded loss by definition. The
docker path may be understood to trade durability for not requiring a host
binary, with the native install documented as the answer for anyone who needs
delivery guarantees.

**Unclosed gap.** `#580`'s fix shows a non-native path can hold the contract:
the TypeScript integrations do not execute the binary either, and they spool by
writing the file directly. Your closing comment on `#580` also lists the
deliverers of those entries as *"`ai-memory hook-drain` and the shell hooks'
piggyback drain"* — if that phrase meant this bundle, the expectation and the
code do not match, since the bundle has no drain.

## Why the obvious fix does not apply

"Make the docker hooks call `ai-memory hook`" fails for the reason the code
already gives: there is no host-local binary on that install. It is also the
reason you gave in `#580` for not shelling out from the TypeScript adapters —
*"the integrations keep working where the binary isn't installed next to the
agent"*. The same constraint points at the same solution: write the spool entry
directly, in the CLI's on-disk contract.

## Reproduction

Following the docker quick start, with hooks installed through the wrapper:

1. `ai-memory install-hooks --agent claude-code --apply` (through
   `bin/ai-memory`, so the rendered commands are the `.sh` bundle).
2. Stop the server: `docker stop ai-memory`.
3. Run a session with a few tool calls.
4. Start the server and run `ai-memory hook-drain`.

Nothing from step 3 arrives, then or later. The same sequence on a native
install delivers every event. Step 2 is not contrived: `ai-memory upgrade`
recreates the container, so every upgrade is one of these windows.

## Scope

I am claiming only what I checked at `2dc9e334`:

- The POSIX bundle (`hooks/*.sh`, `hooks/_lib.sh`): no disk write on failure.
- The PowerShell bundle (`hooks/lib/ai-memory-hook.ps1`): same shape,
  `Invoke-WebRequest -TimeoutSec 3` with no spool. I did not trace its failure
  path as carefully, so I am flagging it rather than claiming it.
- I did not measure how often this costs an event in practice; on a loopback
  server the window is narrow, and the case that matters is the remote-server
  deployment `#580` was reported from.

If the answer is "deliberate", the useful outcome is a line in `docs/install.md`
at the point where a user picks the docker path, and this can be closed. If it is
"gap", a PR follows this issue with the fix written against `main`, taking the
same three choices `#580` made: direct spool writes, an `ingest_key` minted once
at spool time, and 4xx treated as a permanent rejection.

felipe-NR:
## What changed

The POSIX shell hook bundle stops dropping an undelivered lifecycle event.
`ai_memory_post_hook` now reads the status code and, on an unreachable server or
a 5xx, writes the event to `<data_dir>/hook-spool/` in the exact on-disk
contract `ai-memory hook-drain` reads: same filename shape
(`<13-digit ms>-<pid>-<16 hex>.json`), same `SpoolEntry` JSON, 0600 file inside
a 0700 dir, tmp+rename. A 4xx stays a permanent rejection and is not retried.

**Changed default, called out explicitly:** the second commit lowers the script
POST's hard timeout from 500 ms to **200 ms**, the figure invariant 5 documents
in `AGENTS.md` and `docs/ARCHITECTURE.md`. The handoff GET keeps its 1 s.

## Why

Closes #719.

`#580` established that a capture path which POSTs and forgets is a defect, and
fixed it for all four generated TypeScript integrations. The POSIX shell bundle
has the same shape and did not get the same fix — and it is the path the
documented docker install wires (`bin/ai-memory:575` forcing
`AI_MEMORY_HOOK_PLATFORM=posix`; `render_shared.rs:1302`, *"because setup-agent
copies scripts, not a host-local native binary"*).

#719 asks whether that is a deliberate boundary for the docker path or an
unclosed gap. **This PR assumes the second answer.** If the first is right,
close it without ceremony — the useful outcome is then a line in
`docs/install.md` where a user picks the docker path, and I will send that
instead.

The fix follows the three choices `#580` made, for the same reasons:

- **Direct spool writes, not shelling out to `ai-memory hook`** — the docker
  host has no local binary, which is why this path renders scripts at all.
- **`ingest_key` minted once at spool time** and baked into the URL, so this
  bundle's drain and a concurrent `hook-drain` cannot double-ingest.
- **4xx delivered-and-done; 5xx or unreachable spools.**

A delivery that succeeds proves the server is reachable, so it flushes the
backlog behind itself — detached from the hook's process so the agent never
waits, and skipped entirely when nothing is queued, which is every call on a
healthy install. The per-tool-call hot path gains one directory probe.

`ai_memory_json_field` reads a spooled field back, scanning left to right and
undoing the escapes `ai_memory_json_string` emits. An entry carrying a `\uXXXX`
escape came from a richer serializer (the native binary), so it is declined
rather than mangled and left for `ai-memory hook-drain` — the same boundary the
TypeScript drain uses for a file it cannot parse.

### On the timeout commit

The two commits are separable and the second is the smaller claim. `_lib.sh`
called 500 ms "the project-wide hook latency budget" in its own comment while
both invariant statements said `≤200 ms`, so code and documentation disagreed;
`git log -S "max-time 0.2"` finds no hook path that ever used 200 ms, so the
figure dates from the M10 documentation pass (`733126c8`) and the
implementation settled elsewhere. This takes the documented figure as the
intended one. **If 500 ms is what you intend, drop that commit** and the right
fix is to correct the two invariant lines instead; the spool commit stands
either way.

Ordering matters between them: a tighter ceiling times out sooner against a
remote server, and that is only safe because a missed window now spools instead
of dropping. Measured on a loopback install, a real `POST /hook` round trip runs
about 0.3 ms, roughly 600x under the budget, so a local install is unaffected.

## Test plan

- [x] `cargo test --workspace --all-targets` passes (exit 0; `cargo tf` not run,
      nextest is not installed on this host — `cargo test -p ai-memory-cli
      --all-targets` is separately green at 117 passed)
- [x] Manual test: both directions of the spool contract against a live server
      pair. The native `hook-drain` binary consumes a shell-written entry —
      parsing the URL with its shell-minted `ingest_key` and unescaping the body
      back into a JSON object for `/hook/batch` — and retires the file on ack;
      the shell drain delivers a natively-written entry and retires it. Also
      verified 4xx does not spool, 5xx does, and that a spooled entry replays to
      the URL it was bound to rather than the URL of the later success.
- [x] `sh tests/hooks/test_lib.sh`: 64 passed, 0 failed, including nine new
      cases (millisecond helper shape, round-trip of a body carrying quote /
      backslash / newline / tab, `ingest_key` minted, 0600 mode, filename shape,
      `\uXXXX` declined, and `post_hook` spooling against an unreachable port).
- [x] `sh -n` on every `hooks/**/*.sh`.

## Commit attribution

## Release impact

- [ ] **Minor**
- [ ] **Major (breaking)**

`main` sits at `v2.2.0` with an empty `[Unreleased]`, so a `### Fixed`-only
section can cut 2.2.1 straight from trunk without waiting on the 2.3 train.

## CHANGELOG (merge gate)

- [x] `CHANGELOG.md` `[Unreleased]` entry added under `### Fixed`, one bullet
      per commit.
- [x] The changed default (500 ms → 200 ms) is called out in "What changed".

## Notes for reviewers

**The constraint I am most aware of breaking.** `hooks/_lib.sh` opens by asking
that changes stay "byte-trivial," because every supported agent sources this one
file. This adds roughly sixty lines. The mitigation is that the happy path is
untouched: nothing new executes until a delivery has already failed, and the
drain probe is a single directory check that returns immediately when the spool
is empty. If you would rather have this as a separate sourced file, say so and I
will restructure it.

**One incidental robustness fix.** `ai_memory_auth_header_file` returns a false
status when no auth-header file exists, which aborted callers under `set -e` —
the property the function's own comment claims. The two call sites this PR
touches now absorb it; I left the third (`ai_memory_get_handoff`) alone rather
than widen the diff.

**Deliberately not included.** The PowerShell bundle has the same gap
(`Invoke-WebRequest -TimeoutSec 3`, no spool). I left it out to keep this
reviewable and am happy to follow up with the parity change.

**One thing worth your eye in #719.** Your comment closing `#580` describes the
deliverers of those spooled entries as "`ai-memory hook-drain` and the shell
hooks' piggyback drain." I could not find a drain in this bundle at `2dc9e334`
(`grep -rniE "spool|drain" hooks/` returns nothing), so I read that phrase as
meaning the native `ai-memory hook` commands an agent's config invokes. If it
meant this bundle, that is the gap this PR closes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- The POSIX shell hook bundle no longer drops a lifecycle event when the
+  server is unreachable or answers 5xx. A failed delivery is written to
+  `<data_dir>/hook-spool/` in the same on-disk contract `ai-memory hook-drain`
+  reads (same filename shape, same `SpoolEntry` JSON, same 0600/0700 modes,
+  tmp+rename), with an `ingest_key` minted once at spool time so a shell drain
+  and a concurrent `hook-drain` cannot double-ingest; the backlog is flushed
+  behind the next delivery that succeeds, detached from the hook so the agent
+  never waits. A 4xx stays a permanent rejection and is not retried. This is
+  the durability the generated TypeScript integrations got in #580, for the
+  path the docker deploy installs. The PowerShell bundle is unchanged (#NNN).
+
 ## [2.2.0] - 2026-09-12
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   behind the next delivery that succeeds, detached from the hook so the agent
   never waits. A 4xx stays a permanent rejection and is not retried. This is
   the durability the generated TypeScript integrations got in #580, for the
-  path the docker deploy installs. The PowerShell bundle is unchanged (#NNN).
+  path the docker deploy installs. The PowerShell bundle is unchanged (#719).
+- The POSIX shell hook bundle's `POST /hook` now hard-timeouts at 200 ms
+  rather than 500 ms, which is the budget invariant 5 documents for a script
+  hook. A real loopback round trip runs about 0.3 ms, so a local install is
+  unaffected; against a remote server the tighter ceiling is safe only because
+  a missed window now spools instead of dropping (above). The handoff GET keeps
+  its 1 s: it is fed synchronously to the agent's context, and truncating an
+  almost-ready handoff costs more than it saves (#719).
 
 ## [2.2.0] - 2026-09-12
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -593,7 +593,12 @@ agent host, then use that native executable to run
 `install-hooks --agent claude-code --apply`. Even if the script fallback is
 retained, the server still strips any raw field on receipt before persistence.
 
-Native `ai-memory hook --event ...` commands spool events locally. Session start
+Native `ai-memory hook --event ...` commands spool events locally. The POSIX
+shell bundle spools too, but only on failure: it POSTs first and writes the
+event to the same `<data_dir>/hook-spool/` contract when the server is
+unreachable or answers 5xx, then flushes the backlog behind the next delivery
+that succeeds. A 4xx is a permanent rejection and is not retried. (The
+PowerShell bundle still drops an undelivered event.) Session start
 does a short bounded cleanup drain before fetching a handoff; cancellation-prone
 boundary events (`stop`, `pre-compact`, and `session-end`) start a detached
 `hook-drain` helper so delivery does not depend on one shutdown hook surviving.
@@ -865,7 +870,8 @@ successful calls; it reuses the post-tool-use handler), `Stop`,
 native `ai-memory hook --event … --agent kimi-code` commands on local installs
 (local spool plus batched delivery, capture-policy v1 enforced); the staged
 script bundle under `~/.local/share/ai-memory/hooks/kimi-code/` is the
-compatibility fallback (fire-and-forget POSTs to `/hook`). A pending handoff
+compatibility fallback (POSTs to `/hook`, spooling a failed delivery for a
+later drain, without capture-policy v1 enforcement). A pending handoff
 is injected at `UserPromptSubmit` through the hook's stdout, which Kimi Code
 appends to the model context as a user message before the turn; Kimi Code
 fires `SessionStart` but discards that hook's stdout, so hooks installed by
--- a/hooks/_lib.sh
+++ b/hooks/_lib.sh
@@ -446,7 +446,7 @@ ai_memory_clear_session_id() {
 
 # POST stdin to "$1" as JSON. Adds an
 # `Authorization: Bearer` header when `AI_MEMORY_AUTH_TOKEN` is set.
-# The 0.5s timeout matches the project-wide hook latency budget
+# The 0.2s timeout is invariant 5's budget for a script hook
 # (never block the agent), and the trailing `|| true` makes the
 # function safe to call from `set -e` scripts. An undelivered event
 # (unreachable server or 5xx) is spooled for a later drain instead of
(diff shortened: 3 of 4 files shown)


felipe-NR:
While investigating the Cursor duplication reported in #721, I found a separate duplication window in the POSIX shell bundle changed by #720.

The initial POST is sent without an `ingest_key`. If the server persists the event but the client times out or loses the response, the fallback writes the same event to the spool and assigns a new `ingest_key`. The replay can then be accepted as a new event.

This is unrelated to the native Cursor double-hook path reported in #721 and addressed by #722. I am preparing a separate PR that assigns the key before the initial POST, preserves it during spooling and replay, and adds a regression test for an accepted request whose response is lost.

PR #722, opened 2026-09-13
fix(hooks): drop Cursor's duplicate event from Claude Code hooks (fixes #721)

### LINKED ISSUE #721: Cursor: every lifecycle event is stored twice when Claude Code hooks are also installed (2.2.1)
**Version**
Output of `ai-memory --version`: `ai-memory 2.2.1` — native macOS aarch64 release tarball, `serve --transport http` under a LaunchAgent, loopback, `auth=false`.

**What happened**
Every lifecycle event from a Cursor IDE session is stored twice in the same session: two observations 5–9 ms apart with identical kind/title/body. A one-prompt session (session-start, user-prompt, one tool call, stop) produced 12 observations for 6 events.

Session attribution itself is correct on 2.2.1: the session landed under the repository's project with the right `cwd` through the native `workspace_roots` handling from 2.1.0, so the `cwd`-injecting shim we ran before is gone. Thanks for that.

Claude Code and Codex sessions on the same server the same day show no doubling, so this is Cursor-specific.

Observed in SQLite (`sessions.agent_kind = 'cursor'`, one session):

```
kind           created_at (µs)
session-start  1789307351504835
session-start  1789307351513979
pre-tool-use   1789307396663452
pre-tool-use   1789307396671038
post-tool-use  1789307396676696
post-tool-use  1789307396681259
```

`user-prompt` and `stop` were doubled the same way. `extension` is empty on every row.

Hook wiring at the time:

- `~/.cursor/hooks.json`: exactly one native `ai-memory … hook --event <e> --agent cursor --project-strategy repo-root` entry per event (`sessionStart`, `beforeSubmitPrompt`, `stop`, `preToolUse`, `postToolUse`, `postToolUseFailure`, `sessionEnd`, `preCompact`), written by `install-hooks --agent cursor --apply` on 2.2.1 and verified with `jq`. Plus one unrelated third-party `sessionStart` entry.
- `~/.claude/settings.json`: the standard native `--agent claude-code` entries from `install-hooks --agent claude-code --apply`.
- No project-level `.cursor/hooks.json` or `.claude/settings*.json` hooks in the repository that was open.

**Hypothesis**
The 2.1.0 changelog says the Cursor CLI also runs the hook commands declared in Claude Code's `~/.claude/settings.json`, and that payloads carrying `cursor_version` are now attributed to `cursor` regardless of the declared `--agent`. With both Cursor's own hooks and Claude Code's hooks installed on one machine, each Cursor event reaches the server twice: once through `~/.cursor/hooks.json` (`agent=cursor`) and once through `~/.claude/settings.json` (`agent=claude-code`, re-attributed to cursor). Both copies now land in the same cursor session. Before 2.1.0 the second copy became a separate mislabelled `claude-code` session, which is what we saw on 1.32.1, so the re-attribution turned a mislabelling into a duplication.

I have not captured the two hook invocations directly. The evidence is the timing: two writes 5–9 ms apart, for every event, only for Cursor.

**What I expected**
One observation per Cursor event, however many hook surfaces Cursor fires.

**Suggestions** (any one would do)
- Server side: when a `cursor_version`-marked payload arrives through a hook that declared `agent=claude-code`, and a native cursor hook delivered the same session/event within a short window, drop the re-attributed copy. Or more generally, dedupe identical `(session, event, payload hash)` within N ms.
- Or document how to run only one path, for example a flag on the `--agent claude-code` hook command to ignore Cursor-marked payloads when Cursor's own hooks are installed.

**Steps to reproduce**
1. `install-hooks --agent claude-code --apply` and `install-hooks --agent cursor --apply` on the same machine (native macOS install, 2.2.1).
2. Open a repository folder in Cursor IDE 3.20.17, send one prompt that triggers a tool call, let it stop.
3. `sqlite3 <data-dir>/db/memory.sqlite "SELECT o.kind, o.created_at FROM observations o JOIN sessions s ON s.id = o.session_id WHERE s.agent_kind = 'cursor' ORDER BY o.created_at DESC LIMIT 12;"`

**Environment**
- OS: macOS 26.6.2, Apple Silicon
- Docker? no, native release tarball
- Agent: Cursor IDE 3.20.17; Claude Code and Codex hooks are installed on the same host
- Transport: http (loopback, no auth)

**Relevant logs**
`serve` logged nothing for these events at INFO or WARN, and no collision/rejection warnings. The only nearby line is a `handoff fetch failed error=invalid state: startup receiver id belongs to a different scope, agent, or operator` WARN from a separate session-start that Cursor's launcher window fired before a folder was open. That one landed in the default `scratch` project, as expected.

**Related, minor**
`install-hooks --agent cursor --apply` only recognises entries that invoke the `ai-memory` binary as its own. Our pre-2.1.0 `hooks.json` called a small shell shim that injected `cwd` from `workspace_roots`; the 2.2.1 re-apply added the eight native entries beside the eight shim entries instead of replacing them, which would also have doubled every event. A warning when an existing entry mentions `ai-memory` but is not the binary, or a note in `docs/macos.md`, would catch that.

Gaalbu:
Fixes #721.

## Problem
Cursor runs the hook commands from `~/.claude/settings.json` as well as `~/.cursor/hooks.json`. Since 2.1.0 the `--agent claude-code` copy is re-attributed to `cursor` via `cursor_version`, so on a host with both installs every Cursor event is stored twice in the same session.

## Fix
- `hook`: a `--agent claude-code` invocation whose payload carries `cursor_version` is skipped (success response, nothing spooled) when `~/.cursor/hooks.json` already registers an ai-memory `--agent cursor` hook. The file check only runs for Cursor-marked payloads, so plain Claude Code events never touch disk.
- Without Cursor's own hooks the Claude Code path keeps capturing Cursor sessions exactly as before; missing/malformed `hooks.json` counts as not installed.
- `install_hooks`: `cursor_native_hooks_installed_in` reuses `is_ai_memory_hook_entry` and accepts exec-form and command-string entries (bounded read).
- Exported `agent_from_payload` from `ai-memory-hooks`; CHANGELOG entry under Unreleased/Fixed.

Scope: the native `ai-memory hook` command. The shell-shim note at the end of the issue is not addressed here.

## Tests
- New unit tests for the drop decision and the `hooks.json` detection.
- `cargo test -p ai-memory-cli --lib hook` (357 passed), `cargo clippy -p ai-memory-cli -p ai-memory-hooks --all-targets -D warnings`, `cargo fmt --check`, `git diff --check`.


felipe-NR:
I found a separate duplication window
https://github.com/akitaonrails/ai-memory/pull/720#issuecomment-5666830368

Gaalbu:
> I found a separate duplication window [#720 (comment)](https://github.com/akitaonrails/ai-memory/pull/720#issuecomment-5666830368)

Thanks for catching this. It is a real duplication window introduced by #720 

However, this PR only fixes the native Cursor/Claude Code double-hook path, so I'm keeping the scopes separate. The shell retry case is being addressed in #730 !

Gaalbu:
Thanks! :rocket:

PR #723, opened 2026-09-13
fix(install-hooks): warn about Cursor hook wrappers that mention ai-memory (#721)

Gaalbu:
Addresses the "Related, minor" part of #721 (the main duplication is #722).

## Problem
`install-hooks --agent cursor --apply` only replaces entries it recognizes as ai-memory's (the binary's exec form or legacy command strings). A wrapper such as the pre-2.1.0 shim that injected `cwd` from `workspace_roots` is kept beside the eight native entries added on re-apply, so every event is captured twice with no signal.

## Fix
After writing `~/.cursor/hooks.json`, apply scans it and prints a `# warning:` for each entry (flat or nested) whose command mentions `ai-memory`/`ai_memory` but is not recognized by `is_ai_memory_hook_entry`, naming the event and command. It does not remove anything, since the entry could be legitimate, and it never fails the apply. CHANGELOG entry under Unreleased/Fixed.

## Tests
- New unit test `unrecognized_ai_memory_hook_entries_flags_wrappers_only` (native entry ignored, third-party ignored, flat and nested wrappers flagged).
- `cargo test -p ai-memory-cli --lib install_hooks` (153 passed), `cargo clippy -p ai-memory-cli --all-targets -D warnings`, `cargo fmt`, `git diff --check`.

Note: this branch and #722 both add a `### Fixed` block under Unreleased in CHANGELOG.md, so whichever merges second needs a trivial conflict resolution.


Gaalbu:
Thanks! :rocket:

PR #728, opened 2026-09-14
fix(hooks): read a spool entry's fields in linear time (#727)

### LINKED ISSUE #727: Shell hook drain saturates the machine after an outage: `ai_memory_json_field` is quadratic in value length and every hook kicks an unlocked detached drain (v2.2.1, spool from #719)
**Version:** `ai-memory 2.2.1`. Every line is cited against `74d2d31e` (the `v2.2.1` tag on `main`); the installed `hooks/_lib.sh` is byte-identical to that file.
**Platform:** macOS 15 (Darwin 25.5.0), Apple Silicon, system `awk`
**Setup:** server in Docker (OrbStack) on `127.0.0.1:49374`, `http` transport, POSIX shell hook bundle installed for Claude Code (not the native binary hooks)

## Summary

After the server was unreachable for a while and then came back, the shell hook bundle saturated the machine. Three samples, in order:

```
# during the incident
Load Avg: 374.05, 362.67, 315.82      CPU: 90.8% user, 9.91% sys, 0.0% idle
418  /bin/sh .../hooks/claude-code/{pre,post}-tool-use.sh
208  awk -v key ...                    (oldest 28 min)
  0  other awk processes on the machine

# after pkill -f 'ai-memory/hooks'
  6  hook shells
227  awk -v key ...                    (217 now with PPID 1, oldest 33 min)

# after pkill -f 'awk -v key'
Load Avg: 14.87                        CPU: 57.46% idle
  0  awk -v key
```

The hook count was still climbing on its own between the first two samples, with the server already healthy — 350 at one sample, 418 a few minutes later.

The spool held 213 entries / 3.4 MB, and one entry was **2.2 MB** — a single tool result from a `grep` over a large repository. Removing that one entry emptied the backlog within minutes; the remaining 212 entries drained without trouble.

Nothing else on the machine changed across those samples.

The drain is doing real work here, not spinning on an error. It is just quadratic, and it is started once per hook invocation with nothing serialising it.

## Mechanism

**1. The spool has no size cap.** `ai_memory_spool_event` (`hooks/_lib.sh:586`) writes whatever the agent sent, so a large tool result becomes a multi-megabyte entry.

**2. Every successful delivery kicks a detached drain.** On a 2xx, `ai_memory_post_hook` calls `ai_memory_kick_drain` (`_lib.sh:488`), which backgrounds a full pass, fully detached (`_lib.sh:709`):

```sh
(ai_memory_drain_spool 64 >/dev/null 2>&1 &) 2>/dev/null || true
```

There is no lock and no single-flight guard, so each hook invocation starts another pass over the same files — two per tool call, with pre- and post-tool-use.

**3. Each entry is parsed three times.** `url`, `body` and `token` are three separate full passes of `ai_memory_json_field` (`_lib.sh:674`, `:676`, `:677`).

**4. `ai_memory_json_field` is quadratic in the value length.** `_lib.sh:628` unescapes one character at a time into an accumulator (`_lib.sh:652`):

```awk
out = out c
i += 1
```

Each append reallocates and copies a growing string.

**5. The loop feeds itself.** The delivery POST is capped at 200 ms (`_lib.sh:468`, `:476`), and anything that is not 2xx/4xx spools (`_lib.sh:490`). Once the machine is saturated, deliveries stop making that deadline, so each hook both fails to deliver *and* adds another entry to the backlog it is about to make someone else parse. The spool is not only filled by a full outage; a server that is merely slow for 200 ms has the same effect.

I read steps 2, 3 and 5 from the source; I did not capture the individual drain spawns. The process count climbing on its own with the server already healthy is what I observed, and it is consistent with a new detached pass per tool call.

The comment at `_lib.sh:700` is accurate: *"a no-op when nothing is queued — which is every call on a healthy install"*. That is why this stays invisible until an outage fills the spool, and why the spool added for #719 is the path that exposes it.

## Measurements

Synthetic single-line spool entries, same machine, current implementation against the proposal below:

| body size | current | proposed |
|---|---|---|
| 50 KB | 0.24 s | 0.016 s |
| 100 KB | 0.87 s | 0.024 s |
| 200 KB | 4.12 s | 0.043 s |
| 400 KB | 16.71 s | 0.083 s |
| 2250 KB | **529.5 s** | 0.367 s |

Doubling the size roughly quadruples the current implementation's time and roughly doubles the proposal's. The 2250 KB row is measured, not extrapolated: **one pass over the entry that was sitting in my spool costs 8 min 49 s of CPU**, and steps 2 and 3 multiply that by the agent's tool rate, so passes accumulate faster than they retire.

## Repro

1. Install the POSIX shell hooks for Claude Code, server reachable.
2. Stop the server.
3. In an agent session, run one tool whose result is a few MB — `grep -rn` over a large repository will do. Confirm a multi-MB entry under `hook-spool/`.
4. Start the server again.
5. Keep using the agent normally. `ps -Ao args= | grep -c 'awk -v key'` climbs with every tool call and the machine saturates.

## Impact

An outage of any length, combined with one large tool result, turns every subsequent tool call into a multi-minute CPU job. The agent itself never blocks — the drain is detached — so there is no feedback signal until the machine is unusable. On this machine it reached a load average of 374 with zero idle.

## Proposed fix

Make `ai_memory_json_field` linear: find the end of the string with one regex pass, then undo the escapes with `gsub`.

```awk
ai_memory_json_field() {
    awk -v key="$1" '
        { text = text (NR > 1 ? "\n" : "") $0 }
        END {
            needle = "\"" key "\":\""
            start = index(text, needle)
            if (start == 0) exit 1
            rest = substr(text, start + length(needle))
            if (!match(rest, /^(\\.|[^"\\])*/)) exit 1
            out = substr(rest, RSTART, RLENGTH)
            if (substr(rest, RLENGTH + 1, 1) != "\"") exit 1
            gsub(/\\\\/, "\001", out)
            if (out ~ /\\[^ntr"\/]/) exit 1          # \uXXXX and unknown escapes: leave to hook-drain
            gsub(/\\n/, "\n", out)
            gsub(/\\t/, "\t", out)
            gsub(/\\r/, "\r", out)
            gsub(/\\"/, "\"", out)
            gsub(/\\\//, "/", out)
            gsub(/\001/, "\\", out)
            printf "%s", out
            exit 0
        }
    ' "$2"
}
```

`^(\\.|[^"\\])*` is the JSON string grammar, so it stops at the first unescaped quote — the left-to-right property the current comment calls for. `\001` holds escaped backslashes aside so they are not re-read as the start of another escape.

Behaviour is unchanged. I ran both implementations over 14 cases — plain text, `\n`, `\t`, `\r`, `\/`, escaped quotes, escaped backslashes, a trailing backslash, an empty body, a value containing the literal `"body":"`, a missing key, a key that is not first, and the `url` and `token` fields — plus the two that must fail, a `\uXXXX` escape and an unknown escape. Output and exit code are identical on all 14.

Three smaller ones, each independent of the above:

- **Cap what gets spooled.** `ai_memory_spool_event` could refuse or truncate above some size. #249 already established that observation bodies are meant to be bounded; the spool is the one path where an unbounded body reaches a shell parser.
- **Single-flight the drain.** A lock file around `ai_memory_drain_spool` keeps one pass running instead of one per tool call. Without it, any per-entry cost is multiplied by the agent's tool rate.
- **Parse each entry once** instead of three times (`_lib.sh:674-677`).

## Operational note

`pkill -f 'ai-memory/hooks'` does not stop this. It matches the shell scripts by path, but the `awk` children do not carry that path in their argv: they are re-parented to `launchd` and keep running. `pkill -f 'awk -v key'` is what releases the CPU.

munizeverton:
## What changed

`ai_memory_json_field` reads a spool entry's field without ever building the decoded value in memory.

Before this PR the value was rebuilt one character at a time (`out = out c`), so each append reallocated and copied a growing string. The first version of this PR bounded the value with one regex pass over the JSON string grammar and unescaped it with `gsub` over whole segments — but it still appended, once per escaped-backslash pair, into an accumulator that grows to the whole value. That is the same quadratic with a coarser step, and it is invisible on a payload that has no `\\` pairs in it, which is exactly what my first timing table measured. Credit to @samirhvbr for catching it and for the drop-in patch, and to @abhisheksharma2411 for the differential fuzz that pinned the behaviour.

Now nothing accumulates: each segment goes straight to stdout. Validation is a pass of its own, before any printing, so a value that declines halfway prints nothing at all — `ai_memory_drain_spool` reads the field through `|| continue`, and a partial read must never look like a whole one.

Removed as well: `if (!match(rest, /^(\\.|[^"\\])*/)) exit 1`, which both reviewers pointed out is unreachable — the pattern accepts the empty string, so `match` always returns 1. The terminator check on the following line is what rejects an unterminated value, and the comment now says so, so that it does not get "simplified" away later.

Observable behaviour is unchanged: same bytes, same exit codes, including the deliberate `exit 1` on a `\uXXXX` or unknown escape that leaves the entry to `ai-memory hook-drain`, and the raw control byte in a value (`ai_memory_json_string` does not escape those) still round-trips untouched.

No default, flag, config key or on-disk contract changes.

## Timings

Two payload shapes, because the difference between them is the whole story. **grep output** is `grep -rn` over C source, where every literal backslash — the `\n` inside a string, a regex, a Windows path — becomes an escaped pair once JSON-encoded; the 2.2 MB entry in #727 was exactly that. **plain text** has no pairs at all, and is what my original table measured.

2.2 MB grep-shaped entry, 104k escaped pairs, one read of one field:

| awk | before this PR | first version | now |
|---|---|---|---|
| one-true-awk 20200816 (what macOS ships) | 732 s (12 min) | 30.6 s | **5.5 s** |
| BusyBox awk (busybox:latest) | — | 40 s | **8 s** |
| mawk 1.3.4 | — | 9 s | **1 s** |
| gawk 5.3.2 | — | 10 s | **1 s** |

Scaling on one-true-awk, same shape:

| entry | before this PR | first version | now |
|---|---|---|---|
| 200 KB | 4.9 s | 0.20 s | 0.08 s |
| 400 KB | 22.5 s | 0.72 s | 0.21 s |
| 800 KB | 77.4 s | 2.68 s | 0.61 s |
| 2250 KB | 732 s (12 min) | 30.6 s | 5.5 s |

Plain text with no pairs, for contrast: 2250 KB reads in 1.4 s under the first version and 1.6 s now. The accumulator was never the cost on that shape, which is why the original table showed a clean line and the incident payload did not.

### One residual, measured and deliberately not fixed here

With the accumulator gone, the remaining super-linear term is `split()` itself, on one-true-awk and BusyBox awk. Isolating it: with the value length held at 2.2 MB, split is linear in the segment count (10k → 0.35 s, 50k → 1.41 s, 100k → 2.77 s); with the segment count held at 50k, it is linear in the length (500 KB → 0.34 s, 1 MB → 0.69 s, 2 MB → 1.13 s). A grep-shaped payload scales both at once, so doubling it roughly quadruples that phase. Cumulative phase cost on one-true-awk:

| entry | match + terminator | + split | + validate | + unescape and print |
|---|---|---|---|---|
| 400 KB | 0.02 s | 0.12 s | 0.14 s | 0.22 s |
| 800 KB | 0.04 s | 0.42 s | 0.46 s | 0.67 s |
| 1600 KB | 0.08 s | 1.66 s | 1.60 s | 2.24 s |
| 2250 KB | 0.12 s | 3.32 s | 3.36 s | 4.12 s |

mawk and gawk do not show it. Removing it means not splitting the whole value at all, which is a different shape of change to a parser that three people have now checked byte for byte — I would rather land the measured improvement and open that separately than fold a rewrite into a fix.

## This does not close #727

Changed the closing keyword: #727 lists four mechanisms that compound, and this PR removes one of them. Still open there:

- **no spool size cap**, so a multi-megabyte entry is written in the first place;
- **the unlocked detached drain** — `ai_memory_kick_drain` starts a pass behind every 2xx with no single-flight guard, so passes multiply any per-entry cost;
- **three parses per entry** (`url`, `body`, `token`).

## Test plan

- [x] Differential against the first version of this PR: **28 cases × 3 fields × 4 awks = 336 comparisons**, byte-identical stdout and identical exit codes throughout. The same 336 outputs also agree across the four awks. Cases: plain text, each of `\n \t \r \/ \"`, an escaped backslash leading / trailing / doubled, `\\n` and `\\"` ambiguities, a value of nothing but backslashes, a Windows path inside a nested JSON body, a raw `\001` byte in the value, an empty value, a value containing the literal `"body":"`, a key that is not first, a missing key, an unterminated value, a lone trailing backslash, a `\uXXXX` escape, an unknown escape, and a grep-shaped line. The five that must decline produce exit 1 and **zero bytes** on stdout.
- [x] `sh tests/hooks/test_lib.sh` — 62 passed, 2 failed, identical before and after the change. The two failures are pre-existing on this machine and unrelated: `url_encode encodes UTF-8 per byte` (macOS `printf` sign-extends the byte) and `spool_event mints an ingest_key` (a `case` inside `$( )` that bash 3.2 in `sh` mode will not parse). `json_field declines a \u escape` and `spooled body round-trips` pass.
- [ ] `cargo fmt --all -- --check` — not run: no Rust toolchain on this machine. The diff touches no Rust.
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — not run, same reason
- [ ] `cargo test --workspace --all-targets` — not run, same reason

I could not run the Rust gates, so CI is the first thing to check on this branch. As of now the branch reports no checks at all — if the workflows are waiting on maintainer approval for a fork PR, that approval is what would confirm this.

## Commit attribution

## Release impact

      new agent harness or provider

## CHANGELOG (merge gate)

- [x] No default changes in this PR

## Notes for reviewers

`^(\\.|[^"\\])*` is the JSON string grammar, so the match ends at the first quote that is not escaped — the left-to-right property the old comment called for, now enforced by the regex. `match` cannot fail on it, so the terminator check on the next line is the real validation and the comment says not to drop it.

The unescaping splits on the escaped-backslash pairs first (`split(..., /\\\\/)`) and rejoins the segments with a single backslash. That ordering is what makes the remaining `gsub` calls safe: no segment can contain an escaped backslash, so inside a segment every backslash starts a real escape, and `\\n` cannot be misread as a newline. I first wrote this with a sentinel byte holding the escaped backslashes aside, then dropped it — `ai_memory_json_string` escapes only `\\`, `"`, tab, CR and newline, so a raw control byte can reach the value, and any sentinel would have corrupted it. @samirhvbr confirmed that independently with a byte-level comparison; the case matrix above keeps a raw `\001` for exactly that reason.

The two-pass structure is load-bearing, not style: validating inside the printing loop would let a `\uXXXX` found halfway through emit half a value before exiting 1.


abhisheksharma2411 (commented):
An awk rewrite with "observable behaviour is unchanged" attached is exactly the claim worth testing rather than reading, so I extracted both implementations and ran them against each other.

## Differential: 413 inputs, zero divergences

Thirteen hand-picked edge cases plus 400 fuzzed values over an alphabet built to stress the escape handling (`"`, `\`, `\\`, `\n`, `\t`, `\r`, `\/`, a raw `\x01`, and bare `/`):

```
cases: 413   divergences: 0
```

Identical stdout **and** identical exit codes throughout, including the ones that decide behaviour rather than output:

- `\uXXXX` → exit 1 on both, so the entry still falls through to `ai-memory hook-drain`
- an unknown escape (`\q`) → exit 1 on both
- a trailing lone backslash → exit 1 on both
- a raw control byte in the value → round-trips untouched on both
- empty value → empty output, exit 0

That last one is where I expected trouble and didn't find any. The split-on-`\\\\`-first trick is what earns it: because no segment can contain an escaped backslash, every backslash inside a segment starts a real escape, so the five `gsub` calls can't collide with each other and no placeholder byte is needed. The comment says this; the fuzz agrees with the comment.

## Performance: the quadratic shape is visible

Same two implementations, values built from repeated `\n`-bearing chunks:

```
100 KB   old 0.67s    new 0.03s
400 KB   old 10.35s   new 0.07s
```

4× the input costs the old path ~15× the time and the new path ~2.3×. That's the O(n²) → O(n) change showing up directly, and it's the reason your 2.2 MB entry cost minutes: the curve, not the constant.

My 400 KB figure is 10.35s against your 15.0s — different machine, same conclusion.

## One note

`if (!match(rest, /^(\\.|[^"\\])*/)) exit 1` can't actually fail: the pattern matches the empty string, so `match` always succeeds and the guard is unreachable. Harmless, and arguably good defensive style, but the real check is the line after it — `substr(...) != "\""` — which is what rejects an unterminated value. Worth knowing they aren't two independent guards, in case someone later "simplifies" by dropping the second one.

Nice fix. The `gsub`-over-segments approach is the non-obvious part and the comment carries it.

samirhvbr:
Not a maintainer — I went through this while reading the open hook PRs, and found something worth flagging before it merges. Short version: your design call against the `\001` placeholder is right, and I can confirm it with a test; but the replacement for it is still quadratic, and on the platform #727 was reported from it is quadratic enough to leave that issue open.

## Your rejection of the placeholder is correct

The issue's proposal holds escaped backslashes aside in `\001`, and your comment says that corrupts a raw control byte that `ai_memory_json_string` does not escape. That is exactly what happens. Feeding a spool entry whose `body` contains a raw `0x01`:

```
PR #728:            61 01 62     (a, 0x01, b — round-trips)
issue #727 sketch:  61 5c 62     (a, backslash, b — corrupted)
```

So the segment split is the right mechanism and the placeholder is not. Everything below is about the last line of the loop only.

## `out = out "\\" seg[i]` is the same quadratic, with a coarser step

The old code appended once per character. This appends once per escaped-backslash pair, into an accumulator that grows to the full value. The cost is the same shape — it just needs `\\` pairs in the value instead of any characters at all, so it is ~16× cheaper per unit and identical in asymptotics.

That is why the benchmark reads as linear: the fixture has no `\\` pairs in it. The entry from the incident does. #727's 2.2 MB entry was "a single tool result from a `grep` over a large repository", and every literal backslash in source code — `\n` in a string, a regex, a Windows path — is a `\\` pair once the payload is JSON-encoded. My fixture is `grep -rn` output over C-ish source, 2.2 MB, 48,000 pairs:

| awk | #728 | proposed below |
|---|---|---|
| **original-awk 20250116** (the lineage macOS ships as `/usr/bin/awk`) | **44.2 s** | **0.28 s** |
| BusyBox awk 1.37 | 53.4 s | 1.44 s |
| mawk 1.3.4 (Debian/Ubuntu default) | 2.42 s | 0.17 s |
| gawk 5.2.1 | 1.56 s | 0.54 s |

Scaling on original-awk, same fixture at 200/400/800/2250 KB: **0.28 → 1.29 → 4.77 → 44.2 s**. Four times the cost per doubling; that is the quadratic, not a large constant.

The same 2.2 MB with no backslashes in it: 0.19 s under #728 and 0.19 s under the patch — identical, which is the number the PR description reports and the reason this did not surface.

Why it matters for closing #727: the reporter is on macOS system awk, and their mechanism 5 is that passes accumulate faster than they retire. At 44 s per entry per pass, with a detached pass started behind every successful delivery and each entry read three times, that loop still closes. 529 s → 44 s is a real improvement, but it is not the 529 s → 0.29 s the description claims on that platform.

## The patch

Do not accumulate at all — the function's only job is to print. Validate in one pass, stream in the second:

```diff
     parts = split(substr(rest, RSTART, RLENGTH), seg, /\\\\/)
+    for (i = 1; i <= parts; i++)
+        if (seg[i] ~ /\\[^ntr"\/]/) exit 1
     for (i = 1; i <= parts; i++) {
-        if (seg[i] ~ /\\[^ntr"\/]/) exit 1
         gsub(/\\n/, "\n", seg[i])
         gsub(/\\t/, "\t", seg[i])
         gsub(/\\r/, "\r", seg[i])
         gsub(/\\"/, "\"", seg[i])
         gsub(/\\\//, "/", seg[i])
-        out = (i == 1 ? seg[i] : out "\\" seg[i])
+        printf "%s%s", (i == 1 ? "" : "\\"), seg[i]
     }
-    printf "%s", out
     exit 0
```

The validation has to be its own first pass, not merged into the printing one. Otherwise a `\uXXXX` found halfway through a value would exit 1 with half the value already on stdout, and `ai_memory_drain_spool`'s `|| continue` depends on a declined read producing nothing usable. Split this way the all-or-nothing contract is exactly what it is today.

Verified as a drop-in: identical bytes on stdout **and** identical exit codes against your version, on 21 cases × the 3 fields the drain reads (`url`, `body`, `token`) × gawk, mawk, original-awk and BusyBox awk — 252 comparisons, no differences. The cases cover your control-byte round-trip, plain text, each of `\n \t \r \/ \"`, escaped backslashes leading/trailing/doubled, `\\n` and `\\\"` ambiguities, an empty value, a value containing the literal `"body":"`, nested-JSON bodies with `C:\\\\Users` paths, a missing key, a raw newline in the value, and the two that must decline — `\uXXXX` and an unknown escape.

Nit while you are in there: `if (!match(rest, /^(\\.|[^"\\])*/)) exit 1` is dead. The pattern can match empty, so `match` always returns 1 and the branch is unreachable. Inherited from the issue's sketch; harmless, but the following line is what actually validates the terminator.

## Two things about #727 this does not cover

Worth being explicit so the issue does not get closed on this PR alone. #727 lists four mechanisms and this fixes the fourth. Still open:

- **No spool size cap**, so a multi-megabyte entry is written in the first place.
- **The unlocked detached drain** — `ai_memory_kick_drain` starts a pass behind every 2xx with no single-flight guard, two per tool call. This is the multiplier on any per-entry cost, and it is what turned a slow parse into load average 374.
- **Three parses per entry** (`url`, `body`, `token`).

## Unrelated, for context

- Your red `cargo-deny` / `cargo-audit` jobs are not yours — that is `RUSTSEC-2026-0285` against `main`'s lockfile, filed as #731. I opened #734 with the lockfile refresh.
- `hooks/_lib.sh` currently has no CI coverage at all: `tests/hooks/test_lib.sh` runs from no workflow. #736 adds that job across the three awks above, which would have covered this file. The BusyBox column is also its own bug, unrelated to this PR — #733.

Happy to be wrong on any of this; the fixtures and the case matrix are mechanical to reproduce if you want to check the numbers yourself.

munizeverton:
Both of you were right, and the second one changes the number this PR was asking to be merged on. Pushed as `4e801f6`, with the description rewritten around measurements that actually cover the incident payload.

## @samirhvbr — confirmed, independently

I rebuilt your fixture shape (`grep -rn` over C source, where every literal backslash is an escaped pair once JSON-encoded) and timed the real `ai_memory_json_field` through the shell rather than an extracted copy. macOS 15, `/usr/bin/awk` 20200816, 2250 KB / 104k pairs:

| entry | before this PR | as I first wrote it | with your patch |
|---|---|---|---|
| 200 KB | 4.9 s | 0.20 s | 0.08 s |
| 400 KB | 22.5 s | 0.72 s | 0.21 s |
| 800 KB | 77.4 s | 2.68 s | 0.61 s |
| 2250 KB | **732 s** | **30.6 s** | **5.5 s** |

Same payload on the other three: BusyBox awk 40 s → 8 s, mawk 1.3.4 9 s → 1 s, gawk 5.3.2 10 s → 1 s.

And the control that shows why I did not see it: the same 2250 KB with **no** pairs at all reads in 1.4 s before your patch and 1.6 s after. The accumulator was never the cost on that shape, which is exactly the shape my first table measured. That is the whole error, and it is a good lesson — a benchmark fixture that avoids the reported payload will confirm anything.

Your two-pass structure is in as you wrote it, including the reason: validating inside the printing loop would let a `\uXXXX` found halfway through emit half a value before exiting 1, and `ai_memory_drain_spool` reads the field through `|| continue`.

### A residual, which I am leaving in and declaring

`0.21 → 0.61 → 2.26 → 5.5` is not linear either, and publishing "linear" twice on the strength of one fixture is the mistake I had just made, so I profiled the phases before believing it. Cumulative, one-true-awk, grep-shaped:

| entry | match + terminator | + split | + validate | + unescape and print |
|---|---|---|---|---|
| 400 KB | 0.02 s | 0.12 s | 0.14 s | 0.22 s |
| 800 KB | 0.04 s | 0.42 s | 0.46 s | 0.67 s |
| 1600 KB | 0.08 s | 1.66 s | 1.60 s | 2.24 s |
| 2250 KB | 0.12 s | 3.32 s | 3.36 s | 4.12 s |

It is `split()` itself. Isolated on synthetic input: with the value length held at 2.2 MB, split is linear in the segment count (10k → 0.35 s, 50k → 1.41 s, 100k → 2.77 s); with the segment count held at 50k, it is linear in the length (500 KB → 0.34 s, 1 MB → 0.69 s, 2 MB → 1.13 s). A grep-shaped payload scales both at once, so doubling it roughly quadruples that phase. mawk and gawk do not show it; BusyBox does.

That means the mechanism this PR introduced — split on the escaped-backslash pairs — carries its own `O(length × segments)` on the two awks that matter most here. I left it: removing it means not splitting the whole value, which is a different shape of change to a parser three people have now checked byte for byte. Better as its own PR than folded into this one.

One thing worth checking on your side: you measured 0.28 s at 2.2 MB with 48k pairs on original-awk **20250116**; I get 5.5 s with 104k pairs on **20200816**. Halving the pairs accounts for part of that, not all of it. If 20250116 really does 100k pairs in well under a second, then the residual is "the awk macOS ships is five years old" rather than "stop splitting", and that is a different follow-up.

## @abhisheksharma2411 — thank you for the fuzz

Your 413-case differential is what made it safe to change this again today. I re-ran the same kind of comparison against the previous commit, widened to every awk I could get: **28 cases × 3 fields (`url`, `body`, `token`) × 4 awks = 336 comparisons**, byte-identical stdout and identical exit codes throughout, and identical across the four awks as well. The cases cover the raw `\001` round-trip, `\\n` and `\\"` ambiguities, a value of nothing but backslashes, a Windows path in a nested body, an empty value, a value containing the literal `"body":"`, a key that is not first, and the five that must decline — `\uXXXX`, an unknown escape, an unterminated value, a lone trailing backslash, a missing key — each producing exit 1 and **zero bytes**, which is the part that matters now that printing streams.

`sh tests/hooks/test_lib.sh`: 62 passed, 2 failed, identical before and after. Both failures are pre-existing on this machine and unrelated (`url_encode encodes UTF-8 per byte`, and a `case` inside `$( )` that bash 3.2 will not parse).

The dead `match` guard you both flagged is gone. The terminator check on the next line is now visibly the only validation, with a comment saying not to drop it — that was your point about someone "simplifying" later, and it seemed cheaper to remove the decoy than to explain it.

## Two housekeeping notes

I changed `Fixes #727` to a plain reference. #727 lists four mechanisms that compound and this PR removes one; the size cap at spool time, the single-flight lock around `ai_memory_kick_drain`, and the three parses per entry are all still open there, and merging this should not close it.

On CI: this branch reports no checks at all, not red ones — `gh pr checks` comes back empty and so does the status rollup. If fork PRs here need a maintainer to approve workflow runs, that approval is what would confirm the Rust gates I cannot run locally. Thanks for the #731/#734 and #736/#733 pointers either way; a CI job for `tests/hooks/test_lib.sh` would have caught the first version of this, since the fixture problem was mine but the quadratic was visible to any timing test with backslashes in it.

abhisheksharma2411:
Re-ran against `4e801f6` rather than trusting the summary. **Confirmed on all three counts**, and I can give you the timing independently, which is the number CI isn't producing for you.

### The quadratic, measured here

macOS, `/usr/bin/awk`, your grep-over-C-source shape (every literal backslash an escaped pair once encoded), `main` vs this branch, same fixture, same invocation:

| entry | `main` | this PR | output |
|---|---|---|---|
| 9 KB | 0.03 s | 0.02 s | byte-identical, 7999 B |
| 90 KB | 0.47 s | 0.04 s | byte-identical, 79999 B |
| 360 KB | **7.07 s** | **0.13 s** | byte-identical, 319999 B |

`main` goes 15× slower for 4× the input; this branch goes 3.25× for the same 4×. That's the quadratic and its removal, from a third machine, and the output matches exactly at every size — so the speedup isn't bought with a behaviour change.

### The correctness pass

19/19 on my differential — 14 round-trips through `ai_memory_json_string` (escaped pairs, `\"`, a raw `\001`, a Windows path, a value that is nothing but backslashes, a literal `"body":"` inside the value, empty, trailing backslash, CJK/emoji) and 5 declines (`\uXXXX`, unknown escape, unterminated, lone trailing backslash, missing key). Every decline is exit 1 with **zero bytes**, which is the property that matters now that printing streams — `|| continue` in `ai_memory_drain_spool` can't tell a partial read from a whole one otherwise.

Identical 19/19 on `main`, which is the right result: this is a performance change, not a semantics change.

### Confirmed the dead guard is gone

`match(rest, ...)` is still there because it has to be — it's what computes `RSTART`/`RLENGTH`. What's gone is the *redundant* return-value test; the terminator check on the next line is now visibly the only validation, with the comment saying not to drop it. That's the shape I was arguing for.

### One trap worth recording, because it cost me twenty minutes

My first fixture declined on **both** branches and I nearly reported "no quadratic difference." The cause was mine: `json.dumps` defaults to `", "` / `": "` separators, so the file said `"body": "…"` while the needle is `"body":"` with no space. `index()` returned 0, exit 1, 0.03 s — a fast *rejection* that looks exactly like a fast parse if you only time it.

Not a defect in your code — real spool entries come from `ai_memory_json_string` and have no spaces. But anyone hand-rolling a fixture for the CI job you're asking for will hit it, so it's worth a line in the test helper. It's also an argument for the job asserting **output bytes**, not just exit status and wall time.

### On CI

You're right that this branch shows no checks at all rather than red ones. Same thing on my hyperswitch PRs — fork PRs sit at `action_required` until a maintainer approves the workflow run. Nothing you can do from your side; it needs @akitaonrails to click it.

Issue #731, opened 2026-09-14
Update rustls to 0.23.45 for RUSTSEC-2026-0285


felipe-NR:
**Version**
Upstream `main` at `74d2d31ebd8cca656c49f31563fac53e0b61c5cf` (`v2.2.1`).

**What happened**
The current lockfile resolves the direct `rustls` dependency to 0.23.40. The newly published `RUSTSEC-2026-0285` advisory affects rustls 0.23.13 through 0.23.44, causing dependency security gates to fail on PRs that inherit the current `main` lockfile.

This is visible on #730 even though that PR changes only the POSIX hook bundle:

- [`ci / cargo-audit`](https://github.com/akitaonrails/ai-memory/actions/runs/34871373065/job/104067797970) failed after 23 seconds.
- [`ci / cargo-deny`](https://github.com/akitaonrails/ai-memory/actions/runs/34871373065/job/104067797896) failed after 41 seconds.

Both jobs report `RUSTSEC-2026-0285` against `rustls 0.23.40`. Any open PR rerun against the current advisory database can encounter the same failures while `main` retains this lockfile resolution.

The published advisory says rustls can accept TLS 1.3 handshake messages at the wrong encryption level when they follow a key-changing message in the same record. It rates the issue Moderate and identifies 0.23.45 as the patched release:

https://github.com/rustls/rustls/security/advisories/GHSA-2mjx-qc3c-rqvc

`Cargo.toml` already permits compatible 0.23 patch releases, so the fix requires a lockfile refresh rather than a manifest or public-surface change. The same resolution updates `rustls-webpki` from 0.103.13 to 0.103.15.

**What I expected**
`cargo audit` and `cargo deny check` should pass without a known advisory in a direct TLS dependency, including on PRs unrelated to dependency changes.

**Steps to reproduce**
1. Check out `74d2d31ebd8cca656c49f31563fac53e0b61c5cf`.
2. Refresh the RustSec advisory database.
3. Run `cargo audit` or `cargo deny check`.
4. Observe `RUSTSEC-2026-0285` against `rustls 0.23.40`.

**Proposed fix**
Run `cargo update -p rustls --precise 0.23.45`, review the lockfile update, and rerun the full contribution gates. No advisory ignore is needed.

**Environment**
- OS: Linux and GitHub Actions Ubuntu runners
- Docker: no
- Agent: not applicable
- Transport: not applicable

**Relevant logs**
[...]

PR #737, opened 2026-09-14
fix(hooks): JSON escaping was a no-op on BusyBox awk

### LINKED ISSUE #733: POSIX hook bundle destroys every spooled event on BusyBox awk: `ai_memory_json_string` escapes nothing (2.2.1)
**Version**

`ai-memory 2.2.1`. Every claim below is cited against `74d2d31ebd8cca656c49f31563fac53e0b61c5cf` (the `v2.2.1` tag on `main`).

**What happened**

On a host whose `awk` is BusyBox awk, `ai_memory_json_string` (`hooks/_lib.sh:517`) escapes nothing at all. Every event the POSIX shell hook bundle spools is written as structurally invalid JSON, and both drains then destroy it instead of delivering it. `session-start.sh` is broken the same way, so a handoff is never injected.

The cause is one point of awk semantics: BusyBox awk applies backslash processing to a `gsub` **replacement** string; gawk, mawk and one-true-awk pass the replacement through literally. Probing the two replacements the function actually uses:

```
$ printf 'X' | "$awk" '{ gsub(/X/, "\\\\"); print }' | od -An -c
$ printf 'X' | "$awk" '{ gsub(/X/, "\\\""); print }' | od -An -c

gawk           gsub(/X/,"\\\\") ->  \  \     gsub(/X/,"\\\"") ->  \  "
mawk           gsub(/X/,"\\\\") ->  \  \     gsub(/X/,"\\\"") ->  \  "
one-true-awk   gsub(/X/,"\\\\") ->  \  \     gsub(/X/,"\\\"") ->  \  "
busybox awk    gsub(/X/,"\\\\") ->  \        gsub(/X/,"\\\"") ->  "
```

BusyBox awk collapses `\\` to `\` and `\"` to `"`, so `gsub(/\\/, "\\\\")` and `gsub(/"/, "\\\"")` are both no-ops there. A backslash stays a backslash and a quote stays a bare quote.

**Consequence 1 — every spooled event is destroyed**

`ai_memory_spool_event` serializes `url`, `body` and `token` with that function (`hooks/_lib.sh:604-611`). Running the real function from the v2.2.1 tree with an ordinary Claude Code payload, `AI_MEMORY_AUTH_TOKEN` unset:

```
{"url":"http://127.0.0.1:49374/hook?event=post-tool-use&agent=claude-code&ingest_key=sh...","body":"{"session_id":"abc","hook_event_name":"PostToolUse","tool_response":"31 passed"}","created_ms":1789414102126,"auth_mode":"none","attempts":0}
```

The quotes inside `body` are bare, so the object ends after `{`. The same payload under gawk produces the correct `"{\"session_id\":\"abc\",...}"`.

Both drains then lose the event, by different routes:

- **`ai-memory hook-drain`.** Deserializing that file with the real `SpoolEntry` and `serde_json`: `expected ',' or '}' at line 1 column 116`. `load_live_entry` (`crates/ai-memory-cli/src/commands/hook_spool.rs:986`) treats an unparseable spool file as a wedge risk and deletes it — `remove_file`, `result.dropped += 1`. The event is gone with no retry.
- **The shell drain.** `ai_memory_json_field` stops at the first bare quote, so it returns `body` = `{` with exit 0 and POSTs one byte. The server rejects it, and `ai_memory_drain_spool` retires any `4*` as a permanent rejection (`hooks/_lib.sh:693`). The event is gone too.

So on BusyBox awk the spool added in #719/#720 silently discards exactly what it exists to preserve. The comment at `hooks/_lib.sh:534` states the file is written "in the same on-disk contract `ai-memory hook-drain` reads (same filenames, same `SpoolEntry` JSON…)"; that contract does not hold there.

**Consequence 2 — handoffs never inject**

`hooks/claude-code/session-start.sh:38-39` wraps the handoff in `hookSpecificOutput.additionalContext` through the same function, so its stdout is invalid JSON for any handoff containing a quote — which is every handoff. The agent discards the output. This is the same failure #732 reports, reached by an ordinary quote instead of an ANSI byte: #732 is the control-character half of this gap on a gawk-family awk, and this is the quote/backslash half on BusyBox awk. One function, one fix site.

**The repo's own suite already catches it, and CI never runs it**

`tests/hooks/test_lib.sh` on `74d2d31e`, unmodified, selecting each `awk` through a PATH shim:

```
gawk           64 passed, 0 failed
mawk           64 passed, 0 failed
one-true-awk   64 passed, 0 failed
busybox awk    62 passed, 2 failed
                 FAIL  json_string escapes text
                 FAIL  spooled body round-trips
```

The two failures name the bug precisely. No workflow in `.github/workflows/` runs this suite — `tests/hooks/test_lib.sh` appears only in `AGENTS.md:305` as a manual shell-level check — so nothing in the gate can see it, and the shell bundle is the one shipped surface no Rust job reaches.

**What I expected**

Either the bundle works on BusyBox awk, or `install-hooks` refuses to stage it there. What it should not do is install cleanly, look healthy, and delete every queued event.

I am not assuming BusyBox awk is a supported target: it is not mentioned anywhere in the repo, the runtime image is `debian:bookworm-slim`, and "not supported" is a legitimate answer. But the current failure is silent data loss on a host that looks fine, which argues for detecting it either way. A probe in `install-hooks` that runs the two `gsub` forms above and warns would fit the pattern #723 already uses for unrecognized Cursor wrappers.

If the bundle is meant to work there, the fix has to avoid backslashes in `gsub` replacements altogether, because no single replacement string is correct on both families — the same constraint governs whatever escaping #732 gets, which is why I am filing this next to it rather than inside it.

**Steps to reproduce**

1. On a host with BusyBox awk first in `PATH` (Alpine, or a shim: `printf '#!/bin/sh\nexec busybox awk "$@"\n' > bin/awk`), check out `74d2d31e`.
2. `sh tests/hooks/test_lib.sh` → `62 passed, 2 failed`.
3. Or directly against the escaper:
   ```sh
   . hooks/_lib.sh
   printf '{"a":"b"}' | ai_memory_json_string
   # busybox awk: "{"a":"b"}"   (invalid)
   # gawk:        "{\"a\":\"b\"}"  (correct)
   ```
4. Or end to end: with `AI_MEMORY_DATA_DIR` set to a scratch dir, call `ai_memory_spool_event <url> '{"session_id":"abc"}'`, then feed the resulting `hook-spool/*.json` to `serde_json::from_slice::<SpoolEntry>` — it fails, and `load_live_entry` deletes the file.

**Environment**

- OS: reproduced on Debian 13 with `busybox awk` (BusyBox v1.37) selected through a PATH shim; the same applies to any Alpine-based host or dev container. Contrasted against gawk 5.2.1, mawk 1.3.4, and one-true-awk `awk version 20250116` (the family macOS ships as `/usr/bin/awk`), all three of which are correct.
- Docker: not required for the repro; relevant to the container path, which installs this bundle.
- Agent: any agent using the POSIX script bundle. Quoted above for Claude Code.
- Transport: http hook endpoint.

**Relevant logs**

None server-side, and that is part of the report: the native drain's deletion is counted as `dropped` and the shell drain's loss looks like an ordinary 4xx rejection, so neither path logs anything that identifies the cause.

abhisheksharma2411:
Fixes #733.

## Reproduced, and it's wider than reported

BusyBox awk applies backslash processing to a gsub **replacement**; the others pass it through literally. All four escapes in `ai_memory_json_string` carry a backslash in the replacement, so on BusyBox all four are no-ops:

```
                    one-true-awk   busybox
gsub(/X/, "\\\\")        \\           \
gsub(/X/, "\\\"")        \"           "
gsub(/X/, "\\t")         \t           t
gsub(/X/, "\\r")         \r           r
```

The report covers the first two. **The last two are the same root cause and fail more quietly**: they don't break the framing, they corrupt content. A tab becomes the letter `t` and a carriage return becomes `r`, inside a value that still parses — so a payload with no quotes or backslashes gets silently mangled rather than loudly destroyed, and nothing downstream can tell.

End to end with the real function, input `a\b"c<TAB>d<CR><LF>e`:

| tree | awk | result |
|---|---|---|
| `main` | one-true-awk | VALID `'a\b"c\td\ne'` |
| `main` | **busybox** | **INVALID** — `"a\\b"ctdr\ne"`, JSON ends at char 5 |
| this PR | one-true-awk | VALID `'a\b"c\td\ne'` |
| this PR | **busybox** | **VALID** `'a\b"c\td\ne'` |

Rows 1 and 3 are byte-identical, so this is a repair, not a behaviour change for anyone currently working. `ctdr` in row 2 is the tab and CR corruption visible in the wild.

Reproduced against `busybox:latest` in Docker rather than reasoned about, since the whole bug is one implementation's undocumented divergence.

## The fix

Probe the behaviour once in `BEGIN` and pre-double the replacements where they'll be halved:

```awk
probe = "X"; gsub(/X/, "\\\\", probe)
if (length(probe) == 1) { BS = "\\\\\\\\"; ... }   # busybox
else                    { BS = "\\\\";     ... }   # everyone else
```

Detection rather than `uname` or a version string, because the thing that varies is the awk, not the OS — a BusyBox userland on Linux and a GNU userland on the same kernel need different answers.

## Why not split/join

The obvious portable alternative is to split on each metacharacter and concatenate the parts back, avoiding replacement strings entirely. That reintroduces exactly the quadratic cost #727 reported and #728 just removed from the decode side: it appends once per occurrence, and a JSON body is mostly quotes. This keeps `gsub`'s single linear pass.

## Not covered

`ai_memory_json_field` (the decode side #728 rewrote) uses `gsub` replacements of `"\n"`, `"\t"`, `"\r"`, `"\""` and `"/"` — worth a separate look for the same divergence, but it's a different function and #728 is still open, so I've left it alone rather than collide.


abhisheksharma2411:
The two red checks here are not from this change. `cargo-deny` and `cargo-audit` both report **RUSTSEC-2026-0285** against `rustls 0.23.40` — that's #731, and #734 already has the bump to 0.23.45 open.

This PR touches only `hooks/_lib.sh`; `Cargo.lock` is unchanged, so it inherits `main`'s lockfile resolution like every other open PR right now. A rerun after #734 lands should clear both.

Issue #735, opened 2026-09-14
tools/list schemas keep `type: [T, "null"]` for optional fields, which breaks Gemini 3 function calling through strict gateways


JaimeGonzalezVallejo:
**Version**

`ai-memory --version`: 2.0.3 (docker image `akitaonrails/ai-memory:latest`). I checked the v2.2.1 source and the served schema path looks unchanged there.

**What happened**

With a Gemini 3 model (tested `google/gemini-3.8-flash` through the Command Code gateway, which routes to Vertex), every model call fails with a 400 as soon as the ai-memory tools are registered. Vertex says:

```
memory_auto_improve functionDeclaration parameters.min_confidence schema specified other fields alongside any_of. When using any_of, it must be the only field set.
```

The MCP schemas use Draft 2020-12 type arrays for optional fields. `min_confidence` in `memory_auto_improve` comes back from `tools/list` as:

```json
{"description": "Override the proposal confidence floor for this run.", "type": ["number", "null"], "format": "float", "default": null}
```

I counted 73 of these unions across the 18 tools. My harness passes `inputSchema` through unchanged, and the gateway's JSON Schema to Gemini conversion turns each union into `any_of` while leaving the sibling fields (`description`, `format`, `default`) on the same node. Vertex validates function declarations strictly on Gemini 3 and rejects that shape, so the whole request dies. Non-Gemini providers accept the type-array form fine, which is why this only shows up on Gemini 3.

My reading of the source: #148 added `normalize_nullable_types` in `crates/ai-memory-llm/src/gemini.rs`, but that only runs on outbound LLM request schemas. The `tools/list` inputSchema (schemars output in `crates/ai-memory-mcp`) is served as-is.

**What I expected**

The schemas served over `tools/list` to go through the same nullable normalization as the outbound path. Every affected parameter is optional, so `type: ["number","null"]` and a plain `type: "number"` behave the same as far as I can tell, and the single-type form survives any gateway.

**Steps to reproduce**

1. Run ai-memory 2.0.3 with `--transport http`.
2. `tools/list` and look at any nullable parameter (`min_confidence` in `memory_auto_improve` is an easy one).
3. Register the tools with a Gemini 3 model behind a gateway that converts `input_schema` to function declarations, then send any prompt.

**Environment**

- OS: Linux x86_64
- Docker? yes, image `akitaonrails/ai-memory:latest` (2.0.3)
- Agent: pi (my extension registers the MCP tools verbatim)
- Transport: http

[...]

PR #738, opened 2026-09-14
feat(install-mcp): `--flavor` pins the schema dialect for any client

abhisheksharma2411:
Refs #735.

The server already serves the Gemini dialect — `?flavor=gemini`, or `gemini_safe_schemas` server-wide. What is missing is any installer path to it.

`install-mcp` picks a dialect only for the two clients whose upstream is fixed:

| Client | Marker written | Why it can be inferred |
|---|---|---|
| Kimi Code | `?flavor=moonshot` | always Moonshot |
| Kiro CLI | `?flavor=bedrock` | always Bedrock |
| everything else | *(none)* | — |

A client that fronts several models has no answer in that table. #735 is the shape: Command Code routed to Vertex, every model call 400ing at `tools/list` with the exact error `docs/mcp-install.md` already quotes. The fix existed; there was no way to install it.

```bash
ai-memory install-mcp --client command-code --flavor gemini --apply
```

### Why not just map `command-code` → `gemini`

Because it would be wrong for the other half of that client's users. Command Code and OpenCode front several models — the Gemini dialect is needed on Vertex and needed by nobody else, and pinning it per-client would narrow the advertised schema for models that never had the problem. The client's name doesn't carry the answer, so the operator supplies it.

`vertex` is accepted as an alias, matching the server's own `restricted_schema_flavor`.

### Three things that had to move with it

**Flavor resolution moved to one place.** `flavored_url_for(args, url)` — explicit `--flavor`, else the client default, else nothing — applied once in `build_mcp_entry` instead of inside the Kimi and Kiro arms. So `--flavor gemini` on Kimi Code *replaces* `moonshot` rather than stacking a second pair. The server maxes over every pair, so a stacked URL would have worked, but the installed config would name a dialect the operator didn't pick.

There's no way to weaken a client this way: every flavor is at least as permissive as any default, so an override can only relax the advertised schema.

**`mcp-bridge` keeps the bare URL.** The marker is a `tools/list` concern and belongs on the URL a *client* fetches schemas from. The bridge is our own process and re-derives its endpoint; handing it a flavored URL would push the marker to the server on every bridged request.

**`uninstall` had to stop keying off the client.** `mcp_url_candidates` built its list from the client's built-in default, which was complete while only Kimi and Kiro could ever be flavored. Now that any client can be, an entry pinned by hand would survive removal. It rebuilds from `FLAVOR_MARKERS` — every marker the installer can write — for every client. Candidates that don't appear in the file match nothing, so the broadening costs nothing.

This changed an existing test's expectations, which is worth flagging rather than burying: `mcp_url_candidates_adds_client_schema_flavors` asserted exact per-client candidate lists and is now `mcp_url_candidates_covers_every_installable_flavor`. The old name encoded the old contract. Both defaults are still asserted, plus no-duplicate-on-already-flavored, plus the Command Code case the broadening exists for.

### Verification

```
cargo fmt --all -- --check                   clean
cargo clippy --workspace --all-targets -D warnings   clean
cargo test --workspace --all-targets         exit 0, no failures
cargo test -p ai-memory-cli --lib            887 passed (was 882)
```

Each new guard was mutation-tested rather than trusted on a green run:

| Mutation | Fails |
|---|---|
| explicit `--flavor` ignored | `..._pins_a_client_the_installer_cannot_infer`, `..._replaces_the_clients_built_in_default` |
| client default wins over explicit | `..._replaces_the_clients_built_in_default` |
| bridge gets the flavored URL | `the_session_aware_bridge_keeps_the_unflavored_url` |
| `gemini` dropped from `FLAVOR_MARKERS` | `every_flavor_marker_is_one_the_server_recognizes`, `mcp_url_candidates_covers_every_installable_flavor` |
| restored | 887 pass |

**One honest limit:** `every_flavor_marker_is_one_the_server_recognizes` checks the CLI's markers against `FLAVOR_MARKERS`, not against the server's `restricted_schema_flavor` — that's a different crate and the function is private there. So it catches CLI-side drift (a new flavor the uninstaller doesn't know) but not a rename on the server. If you'd rather that be pinned for real, the smallest version is making `restricted_schema_flavor`'s accepted set `pub(crate)` and asserting across the two; say the word and I'll add it.

CHANGELOG under `## [Unreleased] ### Added`, and `docs/mcp-install.md`'s dialect table gains the `--flavor` invocation beneath it.


abhisheksharma2411:
The two red checks here — `cargo-deny` and `cargo-audit` — are not from this change. Both report **RUSTSEC-2026-0285** against `rustls 0.23.40`, which is #731, and #734 already has the bump to 0.23.45 open.

This PR adds no dependencies (`Cargo.lock` is untouched), so it inherits `main`'s lockfile resolution like every other open PR. Once #734 lands a rerun here should clear both.

Everything this branch can be judged on is green: `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, and `cargo test --workspace --all-targets` (exit 0, no failures — 887 in `ai-memory-cli` alone, up from 882).

akitaonrails:
Merged into `release/2.3` in 294b733e (preserving your commits), where 2.3 features accumulate — this is additive (a new `--flavor` option), so it ships in 2.3.0 rather than the 2.2.x patch line. It fully resolves #735, including the Gemini/Vertex case. Thanks for the clean, well-scoped implementation (centralized `flavored_url_for`, `FLAVOR_MARKERS` so uninstall still matches pinned entries, docs + changelog). GitHub won't auto-mark this merged because the PR base is `main`; closing manually.

PR #744, opened 2026-09-15
docs: add EU/enterprise-readiness reference docs (data handling, SSO, air-gapped install)

murilojrpereiras:
## Summary

This adds the documentation artifacts a security/legal reviewer typically needs before approving a dev tool for internal use, aimed at corporates (especially in the EU/Germany) that run vendor-security and GDPR-adjacent review before a tool can touch a workstation. No behavior change — every claim in the new docs is sourced from what's already true in this repo's `SECURITY.md`, `docs/install.md`, and `docs/local-embeddings.md`, just consolidated into the shape a reviewer expects:

- `DATA_HANDLING.md` — single page answering "what does ai-memory collect, where does it live, what (if anything) leaves the machine." Covers the local-by-default storage model, the two existing opt-in paths that send data externally (`capture_assistant`, `AI_MEMORY_RERANKER=llm`), a GDPR "is this personal data" framing, a processor-relationship statement (self-hosted = no processor relationship; a configured cloud LLM provider becomes *your* processor under *your* agreement with them), and deletion/retention.
- `docs/sso.md` — consolidates the existing `auth login oidc-device` flow into an SSO answer for IT-security teams, with an explicit "what this is not" section so it doesn't overclaim what OIDC covers (it authenticates the developer to the CLI/hooks, not the server's own admin auth ladder).
- `docs/airgapped-install.md` — offline install path: the build is already self-contained (bundled SQLite, vendored libgit2, no network fetch during `cargo build`), checksummed release binaries, and the existing offline path for local embedding models from `docs/local-embeddings.md`.
- `SECURITY.md` — one paragraph added to the reporting section: a fallback contact path for reporters whose org requires a channel independent of a GitHub account.
- `README.md` — links the three new docs from the Security section and the Docs table.
- `CHANGELOG.md` — `[Unreleased] ### Added` entry per `CONTRIBUTING.md`'s merge-gate rule.

Not included: an SBOM/license-manifest CI step and anything touching release signing — those change the release pipeline's trust model rather than just documenting existing behavior, so I'm not bundling them here; happy to open a separate discussion/issue for that if there's interest.

## Test plan

- [x] Docs-only change; no code paths touched.
- [x] `git diff --check` clean (no whitespace issues).
- [x] Every factual claim cites an existing section of `SECURITY.md` / `docs/install.md` / `docs/local-embeddings.md` by name, so it can be verified against the current repo state.
- [x] New docs cross-linked from `README.md`'s Security section and Docs table.
- [x] CHANGELOG entry added under `[Unreleased] ### Added` per `CONTRIBUTING.md`.


murilojrpereiras:
Hi @akitaonrails — first-time contributor here. This is a docs-only PR (no code paths touched), consolidating facts already in `SECURITY.md`/`docs/install.md`/`docs/local-embeddings.md` for teams that need a data-flow/SSO/offline-install answer before approving a dev tool internally. Happy to split, trim, or adjust tone/scope however fits the project's conventions — just let me know. Rebased cleanly onto `main` after the 2.2.2 cut, CI just needs a first-run approval. Thanks for maintaining this project.

abhisheksharma2411 (commented):
Checked this the way a security reviewer would use it — every factual claim against the code rather than against `SECURITY.md` — because a doc like this is worse than no doc if one line is wrong. It largely holds up, and I found one structural thing I'd change before it ships.

**What I verified and it's correct:** the local-by-default storage model, "no analytics, crash reporting, or usage telemetry", no encryption at rest in v1, and that both enumerated paths really are off by default and really do require the double opt-in you describe (`capture_assistant` server-side *and* `install-hooks --capture-assistant`).

### The heading undersells the biggest egress path

The section is titled **"The two opt-in paths that send data externally"**, and a reviewer skimming for the answer will read "two" as the complete list. But the embedding provider is a third, and it's the one that moves the most sensitive content.

It *is* in the doc — I want to be fair, I initially thought it wasn't. It's in the preamble:

> …unless one of the two opt-in features below is explicitly enabled, **or you point ai-memory's own LLM/embedding-provider config at a cloud API for its core functionality** (the same way any AI coding tool needs a model to call).

That's a subordinate clause, and the "same way any AI coding tool needs a model" framing normalises it. Compare what each path actually transmits:

| path | what leaves the host |
|---|---|
| `capture_assistant` | the assistant's final-turn text |
| `AI_MEMORY_RERANKER=llm` | each live query + bounded titles/snippets |
| `embedding_provider = openai\|voyage\|google\|openai-compat` | **the text of every stored page** |

The third is the broadest by a distance. `crates/ai-memory-llm/src/embedding.rs` is explicit — `OpenAiEmbedder` hits OpenAI's `/v1/embeddings`, `VoyageEmbedder` hits Voyage's, `GoogleEmbedder` uses Gemini `embedContent`. And it isn't only new writes: `crates/ai-memory-consolidate/src/embed.rs` is the backfill that embeds pages lacking a current vector, with `reembed` exposed through `POST /admin/embed` and `ai-memory embed --force`. So switching a running install from local to a cloud provider ships the **existing corpus** to that provider, not just what's written afterwards.

Your default-safe claim is solid and worth keeping — `config.default.toml` says it plainly: *"in-process local embeddings (all-MiniLM-L6-v2, 384-dim), no API key, no data egress"*. That's a genuinely strong position and the doc should lead with it.

I'd just promote embeddings to a numbered item — "three opt-in paths", with the default-local sentence attached. A German works-council or vendor-security reviewer will ask "does our source material reach OpenAI?" and the honest answer depends entirely on `embedding_provider`, which right now they have to infer from a clause about how "any AI coding tool" works.

### Smaller

- **"the two opt-in paths… Both are off by default"** stays accurate for those two; it's only the heading's implied completeness that's the issue.
- The Art. 28 processor framing at line 72 already says "cloud LLM **or embedding** provider", so the legal section is consistent with three paths while the technical section says two. Aligning the heading would remove that internal mismatch.
- The sanitizer caveat ("best-effort, not a guarantee") is the right call and I'd resist any pressure to soften it. Worth noting the sanitizer sits on the two enumerated paths — I didn't find it on the embedding path, so if that's right it's another reason to list embeddings explicitly rather than by implication.

Genuinely useful PR — this is the artifact that gets a tool approved, and most projects never write it. The above is about making one claim harder to misread, not about whether it should land.

(Reviewer note — not a maintainer, no approval implied.)

murilojrpereiras:
@abhisheksharma2411 thanks — that's a real gap, not a nitpick. Fixed: the embedding provider is now a numbered item (promoted ahead of capture/reranking, since it's the broadest path), with the backfill-on-switch behavior (\`POST /admin/embed\`, \`ai-memory embed --force\`) called out explicitly, and the heading/count updated from "two" to "three". Also noted the local-provider default (\`config.default.toml\`: in-process MiniLM, no key, no egress) up front so that's not buried either.

I checked \`crates/ai-memory-llm/src/embedding.rs\` and \`crates/ai-memory-consolidate/src/embed.rs\` for a sanitizer call on the embed path and didn't find one, so I made that explicit too rather than leaving it implied.

Rebased onto latest main.

abhisheksharma2411 (commented):
This is a genuinely useful document, and the discipline of "restates facts that already live in `SECURITY.md`, cited by section" is the right way to build it. I spot-checked several claims against the tree and they hold up — the embedding-provider section in particular is better than the source, because calling out that switching `local` → cloud triggers a **backfill of the existing corpus** is exactly the thing a reviewer would otherwise miss, and it isn't spelled out that plainly anywhere else.

**One omission, and it's the one that matters most for the audience this page is written for.**

`SECURITY.md` opens by scoping the whole threat model (L14–17):

> ai-memory is a **single-tenant workstation/homelab service**. It supports multiple attributed users, but every authenticated user belongs to the same trust domain and **can read the same project memory**.

and states it again as an explicit non-goal (L134–136):

> **Tenant isolation and per-user ACLs.** Database users provide attribution, not private memory. There are **no per-user or per-project ACLs**; run separate servers/data directories for users who must not see one another's data.

Grepping the new page:

```
"single-tenant"          0 mentions
"per-user ACL"           0 mentions
"Per-project isolation"  1 mention
```

So the one phrase that survives the consolidation is the one a security reviewer will read as a tenancy boundary — and the two statements that say it explicitly isn't one didn't make it across.

**To be fair: the citation itself is accurate.** `SECURITY.md`'s "Per-project isolation" bullet is about namespacing and purge scoping ("a purge operation for project A cannot delete files that also belong to project B"), and the new page cites it in exactly that context — scoped deletion. Nothing here is false.

The problem is the page's own framing: *"reviewers who need a single place to answer..."*. A reviewer who reads only this page, as invited to, sees per-project isolation and scoped purge, and never learns that on a shared server every authenticated account can read every project. For a GDPR-adjacent review that is the load-bearing fact — it decides whether one server can hold two clients' work at all.

Suggest lifting the single-tenant sentence verbatim into the "Default: fully local" section, and adding the "no per-user or per-project ACLs — run separate servers" line next to the purge paragraph so the two travel together the way they do in the source.

Context, and a disclosure: this is live rather than theoretical — #708 reproduces two ordinary users reading each other's projects, and @akitaonrails confirmed it from the schema. I have a docs PR open on the same gap (#746), so I have an interest here; take the suggestion on its merits and ignore the overlap, or fold it in and I'll close mine.

Also worth noting the PR is currently `CONFLICTING` against `main`.

Issue #745, opened 2026-09-15
Server-spawning tests in ai-memory-cli flake under parallel execution


luisfnicolau:
Running `cargo test -p ai-memory-cli --test suite --no-fail-fast` on macOS gives a different failure set each run, and every failing test passes in isolation.

**Run A** (default parallelism): 110 passed, 8 failed —
`autoscope_env::{default_mode_is_per_actor, single_mode_is_still_selectable_via_env, per_actor_mode_via_env, session_ttl_zero_clamps_and_starts}`,
`shutdown_signals::{stdio_server_exits_on_sigint, http_server_exits_on_sigterm, stdio_server_exits_on_sigterm_after_initialize}`,
`serve_shutdown::slow::stdio_serve_exits_when_interrupted`.

**Run B** (`--test-threads=1`): 117 passed, 1 failed — `autoscope_env::default_mode_is_per_actor`.

**Run C** (that same test alone): passes, 2.70s.

### What I ruled out

- **Not the startup budget from #709.** That patch is applied locally here, so `STARTUP_TIMEOUT` is 45s rather than 8s, and these still fail.
- **Not the default bind port.** My first hypothesis was contention on `DEFAULT_BIND` (`127.0.0.1:49374`), but it doesn't survive checking: `autoscope_env` and `shutdown_signals` both already request `127.0.0.1:0`. The `bind=127.0.0.1:49374` that appears in the startup log of every spawned server is `bind = %config.bind` printed unconditionally from config, not evidence of a listener.

I could not identify the cause, so I'm reporting the shape rather than guessing at it. The common factor is that all of these spawn a real server process and wait on it. If CI runs this suite serially, or on a quieter machine, it may well be masked there.

### Environment

macOS. `release/2.3` plus local patches that touch none of these tests' behaviour — the `config.rs`/`serve.rs` changes are an added config field and a startup check, both inert when unset.

Happy to run any instrumented variant you want on this machine, since it reproduces here readily.

abhisheksharma2411:
Tried to reproduce this and **couldn't**, on `main` at `42abd158` (v2.2.2). Reporting that as data rather than as a verdict — a flake nobody else can trigger is still real, and knowing where it doesn't fire narrows it.

```
cargo test -p ai-memory-cli --test suite --no-fail-fast
  → 118 passed; 0 failed; 1 ignored          (26.77s, default parallelism)

… -- --test-threads=16                        (deliberately over-subscribed)
[...]

Issue #699, opened 2026-09-10
ai-memory serve does not exit on ctrl+c


barenko:
**Version**
Output of `ai-memory --version`: ai-memory 2.1.1

**What happened**
serve does not exit on ctrl+c

**What I expected**
terminate the execution of `ai-memory serve` after type `ctrl+c`

**Steps to reproduce**

```sh
ai-memory serve
2026-09-10T07:24:49.477727Z  INFO ai_memory_cli: ai-memory starting version="2.1.1" server_url=http://127.0.0.1:49374 data_dir=/data bind=127.0.0.1:49374
2026-09-10T07:24:49.477877Z  INFO ai_memory_cli::commands::serve: single-instance serve lock held lock=/data/.serve.lock
2026-09-10T07:24:49.480525Z  INFO refinery_core::traits: current version: 60
2026-09-10T07:24:49.480549Z  INFO refinery_core::traits::sync: no migrations to apply
2026-09-10T07:24:49.480554Z  INFO refinery_core::traits::sync: preparing to apply 0 migrations: Map {
    iter: Iter(
        [],
    ),
}
2026-09-10T07:24:49.704414Z  INFO ai_memory_cli::commands::serve: embedder enabled provider="local" model="all-MiniLM-L6-v2" dim=384
2026-09-10T07:24:49.705223Z  INFO ai_memory_cli::commands::serve: starting wiki watcher root=/data/wiki workspace=default project=scratch
2026-09-10T07:24:49.707066Z  INFO ai_memory_cli::commands::serve: active-project isolation mode mode=PerActor session_ttl_secs=3600 max_entries=4096
2026-09-10T07:24:49.711286Z  INFO ai_memory_cli::commands::serve: AI_MEMORY_LLM_PROVIDER unset; memory_consolidate disabled, PreCompact falls back to rule-based checkpoint, lint runs rule-based only
2026-09-10T07:24:49.711321Z  INFO ai_memory_cli::commands::serve: auto-improve scheduler enabled but no LLM provider is configured; job not started
2026-09-10T07:24:49.711327Z  INFO ai_memory_cli::commands::serve: scheduled maintenance started jobs=4
2026-09-10T07:24:49.711333Z  INFO ai_memory_cli::commands::serve: MCP server ready on stdio (Ctrl-C to stop)
^C^C^C^C^C^C^C^C^C^C^C^C
^C^C^C^C^C
[...]

PR #750, opened 2026-09-17
test(serve): make the stdio-shutdown spawn hermetic like its neighbours

samirhvbr:
Refs #745 (one contributing factor, not the whole flake).

`serve_shutdown::slow::stdio_serve_exits_when_interrupted` is the only
server-spawning test in `ai-memory-cli`'s suite that sets neither
`AI_MEMORY_EMBEDDING_PROVIDER=none` nor `RUST_LOG`. `autoscope_env` and
`shutdown_signals` both set both, and label the first one "Hermetic:
the 2.0 default would otherwise start a background model download on
every spawned server."

Consequences of the gap:

- **Embedder.** With the var unset, `Config::embedder_config()` falls
  back to the 2.0 `local` default and `configure_embedder` spawns a
  ~87 MB HuggingFace download on every run of this test — plus reqwest
  DNS work on the blocking pool inside the very window this test
  measures (how fast the process exits after SIGINT). That contention
  is imposed on the whole parallel suite, which is why it shows up in
  #745's failure sets.
- **Log level.** The test waits 60s for a `"ready on stdio"` stderr
  line. An ambient `RUST_LOG=warn` in the runner's environment deletes
  that line, so an environment setting becomes a one-minute timeout
  failure with no hint as to why.

Both are one-line pins. Test-only change: the property under test
(#699 — a signal handler installed before the `initialize` handshake,
observed even when the inherited disposition is `SIG_IGN`) is untouched,
and reverting the `serve` fix still fails this test.

No CHANGELOG entry: test-only churn is exempt per CONTRIBUTING.

Verified: `cargo test -p ai-memory-cli --test suite serve_shutdown` passes
(1.45s), `cargo fmt --all -- --check` clean.


samirhvbr:
Heads-up on the red `test (ubuntu-latest)`: it is **not** this diff, which
touches only `crates/ai-memory-cli/tests/suite/serve_shutdown.rs`. The
failure is in `ai-memory-llm`:

```
thread 'codex::tests::rust_fake_codex_exercises_recovery_and_defensive_failures'
panicked at crates/ai-memory-llm/src/codex.rs:834:13:
mode=exit-nonzero, error=auth: Could not write to Codex recovery process;
run `codex login status` and authenticate Codex again if needed
```

That is the same test e9627e4 de-flaked yesterday, flaking down a third
path the fix did not cover — the child's exit beating the parent's
request write (EPIPE) rather than either EOF observer. Fixed in #751.

Every other check here is green, and the suite passes locally. I do not
have rerun rights on this repo, so this one will stay red until it is
retriggered or rebased past #751.

Issue #748, opened 2026-09-17
export-okf refuses to export any project with a captured event ledger (log-YYYY-MM.md flagged non-conformant, same root cause as #669)


Kalile24:
## Summary

`ai-memory export-okf` refuses to export any project that has ever had an
observation captured, because the monthly event ledger (`log-YYYY-MM.md`)
the server itself writes has no frontmatter and is treated as a
non-OKF-conformant page:

```
server returned 422 Unprocessable Entity: {"error":"page log-2026-08.md is not OKF-conformant; run the server once to migrate before exporting"}
```

This makes `export-okf` unusable for any real project — it only succeeds on
a project with no ledger page at all (e.g. one with zero captured
observations).

## Root cause

Same underlying classifier as #669: `okf::is_conformant`
(`crates/ai-memory-core/src/okf.rs`) requires a non-empty `type` in
frontmatter, and `log-YYYY-MM.md` is written with no frontmatter at all
(`## [<timestamp>] <event> | ...` from the first line). #669 fixed the
*symptom* of this classifier firing on every `serve` boot (re-archiving the
whole data dir), but the classifier itself was not taught to exempt the
ledger page. `export-okf`'s pre-export validation appears to call the same
(or an equivalent) conformance check per-page and aborts the whole export on
the first non-conformant page it finds, which for any actively-used project
is always the current month's ledger.

## Environment

- ai-memory 2.1.1, reproduced again after `ai-memory upgrade` to **2.3.0**
  (container recreated, data intact, error unchanged)
- Docker install via the official wrapper script (`docker.io/akitaonrails/ai-memory:latest`)
- Linux x86_64
- Data dir is the named volume `ai-memory-data`, wiki already OKF v0.2 migrated
  (`wiki format: OKF v0.2 (migrated)` per `ai-memory status`)

## Minimal reproduction

```bash
ai-memory init --data-dir /tmp/okf-export-repro
mkdir -p /tmp/okf-export-repro/wiki/00000000-0000-0000-0000-000000000000/proj
cat > /tmp/okf-export-repro/wiki/.../proj/_meta.md <<'META'
---
project: proj
type: Scope Manifest
---
META
# start the server against this data dir, let it capture at least one event
# so it writes wiki/.../proj/log-YYYY-MM.md with no frontmatter, then:
ai-memory export-okf --project proj --workspace default -o /tmp/proj-bundle.tar.gz
[...]

Issue #669, opened 2026-09-07
`serve` archives the whole data dir on every start: the frontmatter-less `log-YYYY-MM.md` ledger is flagged as pre-OKF 1.x


edsontmarques:
## Summary

On a **fresh 2.1.0 install**, every `ai-memory serve` start writes a full
`ai-memory-backup-okf-v0.2-<ts>.tar.gz` (~81 MB here), even though the next log line
reports `no migrations to apply`. Nothing prunes the archives, so they accumulate one
per start.

The trigger is ai-memory's **own** monthly event ledger. `nonconformant_files()` treats
any `*.md` without a non-empty `type` in its frontmatter as a leftover pre-OKF 1.x file,
and `log-YYYY-MM.md` is written with **no frontmatter at all**. So from the moment the
first observation creates a ledger, `snapshot_before_db_migration()` concludes "this is a
1.x store mid-upgrade" on every boot, forever.

`docs/MIGRATION-2.0.md` documents the opposite:

> The migration is idempotent: restarting the server re-runs nothing and
> never takes a second backup. Fresh installs skip all of it, including
> the backup.

## Environment

- ai-memory **2.1.0**, built from source: `cargo install --locked --git https://github.com/akitaonrails/ai-memory --tag v2.1.0 ai-memory-cli`
- Linux x86_64, no Docker; `serve --transport http --bind 127.0.0.1:49374` as a systemd **user** service
- Data dir created the same day by `ai-memory init` — **never a 1.x directory**
- No LLM provider; local embeddings enabled automatically (2.0 default)

## Root cause

`crates/ai-memory-wiki/src/migrations/m2026_09_okf_conformance.rs`:

`nonconformant_files()` (~L192) walks the wiki and flags every `*.md` except `index.md`
whose frontmatter fails `okf::is_conformant` — which requires only a non-empty `type`
(`crates/ai-memory-core/src/okf.rs` L175):

```rust
} else if ft.is_file()
    && path.extension().is_some_and(|e| e == "md")
    && path.file_name().is_none_or(|n| n != "index.md")
{
```

`snapshot_before_db_migration()` (~L70) then uses "is that list empty?" as its
1.x-detection gate:

```rust
let wiki_root = data_dir.join("wiki");
if nonconformant_files(&wiki_root)?.is_empty() {
    // No pre-OKF wiki files → not a 1.x store being upgraded.
    return Ok(None);
}
```

The monthly event ledger the server itself writes has no frontmatter — it starts directly
[...]

PR #752, opened 2026-09-17
fix(export-okf): exclude the raw event ledger instead of failing the export

Kalile24:
## What changed

`export-okf` no longer refuses to export a project because of its own raw hook
event ledger (`log-YYYY-MM.md`). Before: any project with a captured event
ledger — i.e. any actively used project — failed the whole export with
`page log-YYYY-MM.md is not OKF-conformant; run the server once to migrate
before exporting`. After: the ledger is excluded from the exported bundle
the same way `log.md` already was, and the export succeeds.

## Why

The ledger is frontmatter-less by design (`ai-memory-wiki::ledger`'s own
doc comment). `nonconformant_files()` (the OKF migration scan) and the
watcher's indexer already content-gate this exact shape for #660 and #669,
but `export-okf`'s own validation in `ai-memory-mcp::admin` never got the
same exemption — it calls `okf::is_conformant` on every `.md` file and
`bail!`s on the first miss, which for a real project is always the current
month's ledger. Filed as #748 with a minimal repro.

Fix composes the same two primitives the watcher and the migration already
use (`is_log_ledger_filename` + `opens_with_log_ledger`), promoted from
`pub(crate)` to `pub` in `ai-memory-wiki::ledger` so `ai-memory-mcp` can
reuse them across the crate boundary instead of re-deriving the same
content-gating logic a third time.

## Test plan

- [x] `cargo tf` passes — **with one caveat**: `ai-memory-wiki
      git::tests::concurrent_appends_do_not_fail_a_commit` times out on
      this machine. I reproduced the identical 120s timeout on a clean
      `origin/main` worktree with zero changes applied, so it's pre-existing
      and unrelated to this PR (not caused by touching `ledger.rs`/`lib.rs`
      visibility). Every other test passes: 3305/3306 workspace-wide,
      including the full `ai-memory-wiki` (185/185) and `ai-memory-mcp`
      (413/413) suites and the 5 `export_okf_*` tests specifically.
- [x] Manual test: `cargo build -p ai-memory-cli`, then ran the fixed
      binary's `serve` against a read-only copy of a real 12-project data
      dir (the one from #748, months of real usage, not a fixture). Before
      this branch, `export-okf` returned 422 for 11 of the 12 projects with
      exactly the reported error; on this branch all 12 return 200, and
      `tar -tzf` on the resulting bundles confirms no `log-*.md` made it in
      while real pages (decisions/, gotchas/, sessions/, …) did.
- Added two regression tests in `crates/ai-memory-mcp/src/admin.rs`:
  `export_okf_skips_the_raw_ledger_by_content` (the bug itself: a project
  with a real `log-YYYY-MM.md` ledger now exports 200 OK and the ledger is
  absent from the bundle) and
  `export_okf_still_validates_a_prose_page_named_like_a_ledger` (control:
  content-gating, not filename-only — matches the existing #660/#669
  pattern's own test coverage).

## Commit attribution

## Release impact

## CHANGELOG (merge gate)

- [x] No default changed — this only makes a previously-failing export
      succeed; the bundle's other contents are unchanged.

## Notes for reviewers

- Kept the fix minimal and consistent with the existing pattern rather than
  introducing a new shared "is this a reserved page" helper: `watcher.rs`
  and the migration module each already compose
  `is_log_ledger_filename(...) && opens_with_log_ledger(...)` locally rather
  than sharing one combined function, so `admin.rs` does the same instead of
  refactoring the other two call sites in a bug-fix PR.
- One behavior note: `log.md` (the non-rotated ledger name) was previously
  excluded from the export by filename alone; it's now content-gated like
  everywhere else in the codebase, matching the `#660`/`#669` precedent (a
  page merely *named* `log.md` but holding prose now correctly participates
  in the export instead of being silently dropped). I don't think this
  needs a CHANGELOG callout since it only makes previously-dropped content
  ship when it's real content, but happy to add one if you'd rather be
  explicit about it.


akitaonrails:
Thanks for this — closing as a duplicate of #749, which fixed the same bug (#748) and has now merged to `main`. The two fixes were equivalent; #749 landed with fully green CI, a regression test that also covers the migration-stamped ledger shape (#669), and a `docs/okf.md` correction, so we took that one. Really appreciate the report and the fix.

PR #753, opened 2026-09-17
test(cli): opt non-watcher serve spawns out of the wiki watcher

viniciusdsandrade:
## What changed

CLI tests that spawn a real `ai-memory serve` only to observe startup logs or shutdown-on-signal now pass the shipped `--no-watcher` flag. The child's captured stderr is asserted to contain `watcher disabled by --no-watcher` and not `starting wiki watcher`. Original mode / SIGINT / SIGTERM / stdio-interrupt assertions are unchanged. The production default is unchanged: the watcher still starts unless opted out.

## Why

#745 flakes under parallel `ai-memory-cli` suite runs on the reporter's macOS. #750 made `serve_shutdown` hermetic (embedder + `RUST_LOG`). The remaining owner-named hypothesis is machine-global watcher exhaustion (FSEvents/inotify instances scale with concurrent live servers). None of `autoscope_env`, `shutdown_signals`, or `serve_shutdown` exercises wiki file watching, but every child still installed a watcher before logging the needle they wait on.

This is that next step. I could not reproduce the original parallel flake here, so this does not claim it is gone — only that these children no longer take a machine-global watcher they do not need.

Refs #745.

## Test plan

- [x] `cargo tf` (or `cargo test --workspace --all-targets`) passes
- [x] Manual test: `SSL_CERT_FILE=/etc/ssl/cert.pem cargo test -p ai-memory-cli --test suite --no-fail-fast` → 118 passed, 0 failed, 1 ignored (29.59s, default parallelism). Confirmed `autoscope_env::*`, `shutdown_signals::*`, and `serve_shutdown::slow::stdio_serve_exits_when_interrupted` are in the passing set.

(`cargo tf` was not re-run for the whole workspace; the change is test-only in `ai-memory-cli`. The command above is the same suite and harness the #745 report used.)

## Commit attribution

Author: Vinícius Andrade ([@viniciusdsandrade](https://github.com/viniciusdsandrade)).

## Release impact

- [ ] **Minor** — additive: new flag/subcommand/MCP tool/config key, new agent harness or provider
- [ ] **Major (breaking)** — on-disk format, removed/renamed surface, breaking MCP schema change — called out in "What changed" above

## CHANGELOG (merge gate)

- [ ] I added a `CHANGELOG.md` `[Unreleased]` entry — **required** for any user-facing change: new flag / env var / endpoint / MCP tool / marker key, changed behaviour, or an observable bug fix. (Exempt only for internal refactors, dead-code removal, and test-only churn.)
- [ ] Any changed **default** (flag behaviour, config default, env var, response shape) is called out explicitly in "What changed" above.

Test-only: CHANGELOG exempt per CONTRIBUTING. No default changed.

## Notes for reviewers

`--no-watcher` is already shipped (`ServeArgs::no_watcher` / `start_watcher`). This PR only passes it in the three spawn helpers whose tests do not exercise watching, including `spawn_and_wait_for_exit` (fail-fast invalid config never reaches the watcher, but every serve spawn in that file is then hermetic). Tests that actually watch wiki files are untouched.

The disable-line assertion drives the real `CARGO_BIN_EXE_ai-memory` binary and the child's real stderr; it is not a re-implemented watcher check.


abhisheksharma2411 (commented):
The change is right and the honesty about what it does *not* claim is the right call. I went after the thing nobody has done yet — @akitaonrails asked for the exhaustion hypothesis to be **confirmed**, and three of us have now failed to reproduce the flake, so I measured the per-server cost directly instead.

`ai-memory serve --transport http --bind 127.0.0.1:0`, embedder pinned to `none`, sampled 5s after start:

```
                     total fds   threads   REG   KQUEUE
with watcher (today)     26         20      16      3
--no-watcher (this PR)   21         17      11      3
                        ----       ----    ----   ----
watcher costs             +5        +3      +5      0
```

**Two things fall out, and one of them refines the hypothesis rather than confirming it.**

**1. The watcher costs no kqueue descriptors on macOS.** Identical at 3 either way. So "FSEvents instances scaling with concurrent servers" isn't visible as a per-watcher kernel-queue handle here — whatever the FSEvents backend allocates, it isn't showing up as a descriptor this measurement can see. The cost that *is* there is 5 regular-file descriptors and 3 threads, and the regular files are overwhelmingly SQLite handles (`memory.sqlite`, `-wal`, `-shm`), i.e. the watcher brings its own store connection along.

**2. The limit is the variable, and it's worth asking the reporter for.** My `ulimit -n` is `1048576`, which is why fd pressure could never show here — and it may be why neither @samirhvbr nor I can reproduce. **macOS ships a default `ulimit -n` of 256.** At 26 fds per live server, ~10 concurrent servers is ~260 descriptors plus the harness's own — straight into a 256 limit. At 21 it's ~210, roughly 20% of headroom bought back by exactly this PR.

So the useful next question for #745 isn't another run on another machine, it's one line from the reporter:

```
ulimit -n
```

If that comes back 256 (or anything under ~1024) the hypothesis has a concrete mechanism and this PR is a real mitigation rather than a speculative one. If it comes back large, fd pressure is ruled out and the remaining suspects are the 3 threads per watcher or something else entirely.

**On the change itself:** asserting both the positive (`watcher disabled by --no-watcher`) and the negative (`starting wiki watcher` absent) is the right shape — the positive alone would still pass if the flag were renamed and silently ignored. Scoping it to the three suites that only observe startup logs or shutdown signals is correct; none of them touches wiki file watching, so nothing under test is weakened, and leaving the production default alone means this can't mask a real watcher regression.

One small thing: it would be worth saying in the PR body that `--no-watcher` is a **shipped user-facing flag**, not a test-only escape hatch — it already exists at `cli.rs:2465` with `serve.rs:2116` honouring it. That makes this "use the supported opt-out" rather than "add a test hook," which is a stronger argument for merging it regardless of whether #745 turns out to be watcher-related at all.

PR #754, opened 2026-09-17
fix(mcp): let memory_consolidate default to the latest completed session

caminhar-deus:
## What changed

`memory_consolidate` no longer requires `session_id`, and a caller-input mistake
no longer comes back as an opaque internal error.

Before:

- `{}` → `-32602 failed to deserialize parameters: missing field 'session_id'`
- `{"session_id": null}` → `-32602 failed to deserialize parameters: invalid type: null, expected a string`
- `{"session_id": ""}` → `-32603 malformed record in store: invalid uuid: invalid length: expected length 32 for simple format, found 0`
- `{"project": "<name that does not exist>"}` on any scoped tool → `-32603 project '<name>' not found in workspace '<ws>'`

After:

- An omitted `session_id` — or an explicit `null` — resolves the **latest
  completed session of the resolved project** via the existing
  `latest_completed_session_for_project` helper, the default
  `memory_auto_improve` and `memory_read_session_observations` already use. A
  project with no completed session fails as `no completed session in <scope>;
  pass session_id to consolidate a specific session`.
- A blank `session_id` (`""`, or whitespace only) is the same request as an
  omitted one.
- A malformed id answers `-32602 invalid params`, the code
  `memory_auto_improve` and `memory_read_session_observations` already answer
  with for the same argument. The message is unchanged.
- Scope-resolution failures over MCP map `is_bad_request()`/`is_not_found()` to
  `invalid_params` — the same 400/404 split the web route already applies in
  `scope_error_response`. Only `WriterRequired` (create-on-write resolution with
  no writer handle) and `Store` failures stay internal. Messages are unchanged.
- A consolidation LLM call that fails on a transient provider error (`429`, any
  `5xx`, a transport timeout or connect failure) is retried twice, two seconds
  apart, before the failure is reported — the bounded policy `bootstrap`
  already applies to its chunks. Deterministic failures (auth, schema, a
  malformed-request `4xx`, unparseable or truncated output) still fail on the
  first attempt, and a failed attempt writes no page.

An explicit, valid id keeps its previous behaviour.

## Why

`memory_consolidate` was the only tool in the cluster with no natural default:
`memory_auto_improve` and `memory_read_session_observations` already fall back to
the latest completed session when the id is omitted, and there is no CLI
equivalent (`ai-memory consolidate` is not a subcommand). The strict schema made
the agent discover the requirement by failing.

The error codes were the second half of the same report. An agent that followed
the nearest `.ai-memory.toml` and passed the project name it declares got a
generic internal error when the store spelled it differently, which reads as a
server fault rather than "fix your scope". A blank id reached the store's UUID
parser instead of the documented default. And a `503` window during
consolidation made the operator re-issue the whole call by hand, several times a
day, for something a two-second retry absorbs.

## Test plan

- [x] `git diff --check`
- [x] `cargo clippy -p ai-memory-mcp -p ai-memory-consolidate --all-targets -- -D warnings`
- [x] `cargo test -p ai-memory-consolidate -p ai-memory-mcp -p ai-memory-store` —
      `ai-memory-consolidate` 221 passed (4 ignored), `ai-memory-mcp` 431
      passed, `ai-memory-store` 452 passed (1 ignored); 0 failed
- [x] Manual, over MCP Streamable HTTP (`stateful=false`) against a locally
      built image of this commit:
      - `{}`, `{"session_id": null}`, `{"session_id": ""}` and
        `{"session_id": "   "}` with `dry_run: true` all resolve the same
        `sessions/<id>.md` page
      - `{"session_id": "not-a-uuid"}` → `-32602` `malformed record in store:
        invalid uuid: …` (was `-32603`)
      - `memory_status` with a project that does not exist → `-32602` `project
        '…' not found in workspace '…'` (was `-32603`); with the project that
        does exist → counts

Tests added, in the style of the sibling tools: in `ai-memory-mcp`,
`memory_consolidate_treats_a_blank_session_id_as_omitted`,
`memory_consolidate_rejects_a_malformed_session_id_as_invalid_params`,
`memory_status_reports_an_unknown_project_as_invalid_params` and
`scope_error_keeps_internal_failures_internal` (the last pins that
`WriterRequired`/`Store` stay internal); in `ai-memory-consolidate`,
`a_transient_provider_failure_is_retried`,
`a_transient_provider_failure_gives_up_after_the_budget` and
`a_deterministic_provider_failure_is_not_retried_and_writes_no_page`.

## Commit attribution

- [x] Every commit is authored and committed by
      `Caminhar <o_caminhar_com_Deus@proton.me>` (the "Update branch" merge
      commit is GitHub's own)

## Release impact

## CHANGELOG (merge gate)

- [x] `CHANGELOG.md` `[Unreleased]`: one entry under `### Changed` (the retry)
      and two under `### Fixed` (the error classification) — no changed default

## Notes for reviewers

The two commits are independent: the first adds the omitted-id default, the
second fixes what that default left behind (blank id, error codes) and adds the
provider retry. The retry lives in `ai-memory-consolidate` rather than in the MCP
handler so the SessionEnd and `serve` consolidation paths get it too, and it
reuses `LlmError::is_transient()` — the classification `bootstrap` already
trusts, which deliberately excludes `Serde`/`UnexpectedShape`, so a truncated
provider response is still reported rather than retried.


akitaonrails:
Audited — the fix is correct: `memory_consolidate` was the only tool in its cluster requiring an explicit `session_id` (its siblings `memory_auto_improve` and `memory_read_session_observations` already default), so omitting it failed deserialization before the handler ran. The implementation defaults to the latest **completed** session for the project (`ended_at IS NOT NULL`, scoped, no owner filter — the right precedent for a rerunnable write op, consistent with invariant #16), with clean errors for the empty case; tests + docs are solid. Because relaxing a required MCP parameter to optional is an **additive** change to the input surface, it lands as a **minor (2.4.0)** per our semver rule rather than a 2.3.x patch — I've retargeted the base to `release/2.4`. It should merge cleanly there (release/2.4 already contains main). Minor nit for follow-up: a blocking-reject path elsewhere returns 500 vs 403 — not relevant here. Thanks!

akitaonrails:
Correction to my note above: after retargeting to `release/2.4`, GitHub shows a merge conflict — it's just the CHANGELOG `[Unreleased]` section overlapping with another fix that landed there. No rush, since this is parked for the 2.4.0 cut rather than merging now: a quick rebase on `release/2.4` (moving your `### Changed` entry under the current `[Unreleased]`) will clear it when we prepare 2.4.0. The code itself is unaffected.

abhisheksharma2411 (commented):
Three separable improvements in one PR and all three are the right call — the default-session change removes a caller-input trap, and splitting `-32602` from `-32603` means an operator's malformed scope stops looking like a server bug. The retry piece is the one I looked at hardest, since a retry policy is easy to get subtly wrong.

**The classifier is well-drawn.** I expected `429` to have been left out of `is_transient` and it isn't:

```rust
Self::Provider { status, .. } => *status == 429 || (500..=599).contains(status),
Self::Http(e) => e.is_timeout() || e.is_connect(),
_ => false,
```

Auth, schema, malformed `4xx`, deserialization and unexpected-shape all stay deterministic, which is the half people usually get wrong by retrying everything. Refusing tenacity-style escalation and pointing at the cognee #2840 lesson is a good constraint to have written down next to the constant rather than in a commit message.

**One thing worth naming, since `429` is in the set: the retry is blind to `Retry-After`.** `LlmError::Provider` carries only `{ status, body }` — no headers — and `grep -rni "retry.after" crates/ai-memory-llm/` finds nothing, so a `429` is retried at a fixed 2s twice regardless of what the provider asked for. For the `5xx`/timeout cases a fixed short delay is exactly right. For `429` it's the case least likely to benefit: a provider that says `Retry-After: 60` will answer the 2s and 4s retries with two more `429`s, burning both attempts and adding load to an endpoint that just asked for less.

That isn't a regression — today's behaviour is a single attempt, so three is strictly better — and I would *not* suggest honouring a 60s `Retry-After` here, because that's the escalating-wait shape the comment deliberately rejects. The cheap version is the opposite: when the provider states a delay **longer than the bounded budget**, stop early rather than spend two attempts that cannot succeed, and report the wait it asked for so the operator knows to re-issue later. That needs the header plumbed onto `Provider`, which is a bigger change than this PR — worth an issue rather than scope creep.

**On bundling:** the three changes are independent, and the CHANGELOG describes each on its own terms, which is what makes it reviewable. The one that will surprise people is the default-session change — `memory_consolidate` without `session_id` used to fail deserialization and now does real work on the latest completed session. Aligning it with `memory_auto_improve` and `memory_read_session_observations` is the right consistency argument, and `dry_run=true` gives a preflight, so the surprise is bounded.

`no completed session in <scope>` is a much better failure than `missing field session_id` for the empty-project case.

PR #756, opened 2026-09-18
fix(cli): apply message send admission and post-commit observers

szsip239:
> Draft: holding review while the broader event-driven result-delivery path is developed and validated locally. No additional implementation will be pushed before local end-to-end validation.

## What changed

`ai-memory message send` previously inserted a message through `/admin/messages/send` without invoking configured `message_send` admission or observer webhooks. The MCP send path already invoked them. The CLI/admin path now uses the same recipient-scoped order: authorize, insert, notify observers only after successful commit.

With no matching webhook configured, behavior is unchanged. A configured blocking reject policy now applies to CLI sends too. Nonblocking observers remain asynchronous and use the existing bounded dispatcher. The existing authenticated skip-header policy is preserved.

## Why

An event-driven mailbox consumer configured for `message_send` received notifications for MCP sends but silently missed CLI sends. This fixes that path inconsistency without adding an endpoint, config key, model call, agent integration, or scheduling mechanism.

The existing notification is best-effort scope metadata, not a durable outbox or processing ACK. This PR does not add raw Stop-capture notifications or alter message pop/cancel.

## Test plan

- [x] `git diff --check`
- [x] `env -u CODEX_HOME cargo test --workspace --all-targets`
- [x] `cargo deny check`
- [x] New regression tests: an observer reads the committed message while its delayed response does not block send; a rejecting decider leaves the inbox empty; a full inbox returns 409 without notifying an observer.
- [x] Before the fix, the observer and rejection tests failed; all three pass after it.
- [x] Isolated released-binary comparison: an identical nonblocking webhook received an MCP send event but no CLI send event.

The full suite must run without an inherited `CODEX_HOME`: the existing native-session discovery test otherwise looks in the caller's real harness store instead of its temporary fixture. Test builds use debug info/incremental disabled to bound local build storage.

## Commit attribution

- [x] Verified every commit uses the author's GitHub noreply identity.

## Release impact

- [x] **Patch** — existing-path behavior fix, no new surface.
- [ ] **Minor**
- [ ] **Major (breaking)**

## CHANGELOG

- [x] Added an `[Unreleased]` fix entry and updated messaging documentation.
- [x] The configured-policy behavior change is called out above.


akitaonrails:
Audited — the fix is correct: it routes CLI/admin `message send` through the same admission gate + post-commit observers the MCP path already enforces (the base admin handler went straight to `insert_message`, silently skipping a blocking admission webhook meant to refuse cross-project text). Tests are sound. Since you've marked it draft and are holding for local end-to-end validation, we'll leave it for you to mark ready — it merges to `main` (2.3.x) as-is when you are. One optional nit for when you pick it back up: a blocking-reject refusal currently returns 500; 403 would be more accurate (tests only assert non-success, so not a blocker).

akitaonrails:
Audited (read-only). The fix is complete and well-tested: `handle_send_message` now authorizes recipient-scope admission *before* insert and dispatches observers only on success, mirroring the MCP path exactly; invariant #6 is preserved (the admin path already scrubs body/subject via the sanitizer — no bypass added); the three test cases (observer sees committed row, blocking reject leaves inbox empty, full-inbox 409 doesn't notify) all hold. Respecting your draft-hold while the broader event-driven delivery path is developed — it's mergeable in substance whenever you lift the hold.

PR #777, opened 2026-09-18
fix(config): defer missing llm_fallbacks credentials to serve (#762)

### LINKED ISSUE #762: CLI commands hard-fail when llm_fallbacks api_key_env is absent from the invoking shell (breaks read-only status)
# CLI commands hard-fail when `llm_fallbacks` `api_key_env` is absent from the invoking shell (breaks read-only `status`)

## Summary

After adding an `[[llm_fallbacks]]` profile to `config.toml`, **every** CLI invocation that loads the config (including read-only diagnostics like `ai-memory status`) fails at startup validation if the credential env var named by `api_key_env` is not present in the **calling shell's** environment — even when the running server is perfectly healthy and was started with that env var injected by its service wrapper.

This effectively requires exporting provider API keys in every interactive shell, which works against the fallback design's stated goal of keeping credentials out of config and minimizing credential spread (`docs/llm-provider-fallback.md`).

## Environment

- ai-memory 2.2.1, Windows native install (`ai-memory.exe`), server running as a WinSW service over HTTP
- LLM chain: primary `opencode/mimo-v2.5`, fallback `openai-compat/mistralai/mistral-nemo` (OpenRouter)
- `config.toml`:

  ```toml
  llm_provider = "opencode"
  llm_model = "mimo-v2.5"

  [[llm_fallbacks]]
  provider = "openai-compat"
  model = "mistralai/mistral-nemo"
  base_url = "https://openrouter.ai/api/v1"
  api_key_env = "OPENROUTER_API_KEY"
  ```

- The server process has `OPENROUTER_API_KEY` injected via the service wrapper's env block → starts fine and reports the chain in status.

## Steps to reproduce

1. Configure the `[[llm_fallbacks]]` block above and start the server in an environment where `OPENROUTER_API_KEY` **is** set → startup succeeds, `ai-memory status` (run with the var set) shows:

   ```
   providers:
     llm:       opencode/mimo-v2.5 ok
     fallback chain:
       opencode/mimo-v2.5 (last selected)
       openai-compat/mistralai/mistral-nemo
   ```

2. Open a shell where `OPENROUTER_API_KEY` is **not** set and run:

   ```
   ai-memory status
   ```

## Actual

```
Error: llm_fallbacks[0].api_key_env=OPENROUTER_API_KEY is set but the environment variable is missing or empty
```

Non-zero exit, no status output at all. Same failure for other config-loading subcommands (e.g. `llm-test`).

## Expected

Read-only/diagnostic commands should not hard-fail because a *fallback* credential is missing from the invoking shell. The server itself is the component that needs the credential; the CLI is just asking for status. Reasonable behaviors:

- **Graceful degradation (preferred):** report the candidate as `credential missing` / `unavailable` in the health/fallback-chain output and still print status; or
- Enforce credential resolution only on the `serve` path, letting other subcommands skip fallback-profile validation; or
- At minimum, document loudly that every CLI call now requires the fallback env vars exported.

## Additional context

- The failure message is helpful and the fail-closed intent is sound **for the server** (a silently broken fallback is worse than no fallback). The problem is that the same strictness is applied to invocations that will never call the LLM.
- Workaround we adopted: persist `OPENROUTER_API_KEY` as a Windows user-level env var. That works but spreads the credential to every process the user runs — precisely the sprawl the `api_key_env` indirection seems meant to avoid.
- Possibly related implementation area: config validation resolving `api_key_env` eagerly at `Config::load()` time (`crates/ai-memory-llm`, fallback profile loader).

rafaelkenedy:
Fixes #762.

## The failure

`Config::load` resolved every `[[llm_fallbacks]]` credential eagerly, so every command that loads the config — including read-only `ai-memory status` and the other thin clients — exited with

```
Error: llm_fallbacks[0].api_key_env=OPENROUTER_API_KEY is set but the environment variable is missing or empty
```

whenever the variable was absent from the *invoking shell*, even with a healthy server that had the key injected by its service wrapper (WinSW in the report). Only the server ever builds the chain. The reporter's workaround — persisting the key as a user-level env var — spreads it to every process the operator runs, which is what `api_key_env` indirection is meant to avoid.

## The approach

Keep the fail-closed guarantee where it matters (`docs/llm-provider-fallback.md`: a fallback must not sit latent until the primary has an outage) and drop it where it doesn't:

- **`Config::load`** still reads the environment once (invariant #1) and still validates every profile — provider, model, base URL, provider construction — for every command. When an `api_key_env` variable is absent, the profile is validated with a discarded placeholder key, left out of `llm_fallback_configs`, and its error message recorded in a new `llm_fallback_unresolved` field instead of being raised.
- **`Config::require_llm_fallback_credentials()`** raises that message. `serve::run` calls it at startup, before anything else opens the data dir, so the server refuses to start exactly as before — whether or not a primary provider is configured.
- **`Config::llm_provider_chain()`** also refuses to build while a profile is unresolved. This is the guard that makes the deferral correct rather than merely convenient: the `serve` check fails fast, but the invariant — no chain with a fallback silently dropped — is held where the chain is built, so a future caller that constructs one outside `serve` startup cannot get a quietly shortened chain.

The error text is unchanged, so anything matching on it keeps working.

## Changes

| file | what |
|---|---|
| `crates/ai-memory-cli/src/config.rs` | deferred credential recording in `load`, `llm_fallback_unresolved`, `require_llm_fallback_credentials()`, guard in `llm_provider_chain()` |
| `crates/ai-memory-cli/src/commands/serve.rs` | startup check |
| `docs/llm-provider-fallback.md` | which failures stop every command vs. only `serve` |
| `CHANGELOG.md` | `### Fixed` entry (#762) |

## Tests

`load_rejects_a_missing_or_empty_api_key_env_value` asserted the old behavior and is replaced by:

- `a_missing_api_key_env_value_defers_to_serve_and_the_chain` — `load` succeeds, the profile never reaches `llm_fallback_configs`, and both `require_llm_fallback_credentials()` and `llm_provider_chain()` fail with the original message.
- `a_missing_api_key_env_value_does_not_hide_a_malformed_profile` — an `openai-compat` profile with no `base_url` still fails `load` when its key is also absent.
- `a_config_without_unresolved_fallbacks_passes_the_serve_check`.
- `every_api_key_provider_accepts_the_unresolved_placeholder` (review follow-up) — `anthropic`, `openai`, `gemini`, `opencode`, and `openai-compat` all load with the credential absent, record it as unresolved, and discard the placeholder-built config. Fails if a constructor ever starts checking key shape.

Manual repro on Windows with the config from the issue (`opencode` primary, `openai-compat`/OpenRouter fallback):

| scenario | result |
|---|---|
| `ai-memory status`, fallback var unset | runs (reports the local spool; no server was up) |
| `ai-memory serve`, fallback var unset | `Error: llm_fallbacks[0].api_key_env=... is set but the environment variable is missing or empty` |
| `ai-memory serve`, vars set | starts, chain built, LLM checkpointing enabled |

## Gates

`cargo fmt --all -- --check`, `git diff --check`, `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test -p ai-memory-cli --lib` 918 passed. The CLI integration suite and the full `cargo tf` tier were not run locally.


abhisheksharma2411 (commented):
Right diagnosis and the right place to draw the line — a credential the *server* needs has no business failing `ai-memory status` in an operator's shell. I checked the two things that could make a deferral unsafe, and both hold.

**1. The deferral is enforced twice, independently.** The PR body says the chain still fails rather than silently dropping a fallback, and it does — but not only at `serve`:

```rust
// serve.rs:808
config.require_llm_fallback_credentials()?;

// config.rs — llm_provider_chain()
// "A chain with a profile silently dropped would look healthy until
//  the primary has an outage."
if let Some(message) = self.llm_fallback_unresolved.first() { … }
```

That second guard is the one that matters. `serve` failing fast is convenient; the chain builder refusing is what makes it *correct*, because any future caller that constructs a chain without going through `serve` startup still cannot get a quietly-shortened chain. Worth calling out in the PR body — it's the difference between "we remembered to check at startup" and "the invariant is held where it's used."

**2. The placeholder never escapes `load`.** `UNRESOLVED_FALLBACK_KEY` was the thing I wanted to be sure about, since a stand-in credential that reaches a provider is a much worse bug than the one being fixed. It doesn't:

```rust
match missing {
    Some(message) => unresolved.push(message),      // profile NOT pushed
    None          => fallback_configs.push(provider_cfg),
}
```

The placeholder-bearing config is built purely to validate the rest of the profile, and only the resolved ones reach `llm_fallback_configs`. `grep -rn llm_fallback_configs crates/` finds no consumer outside `config.rs`, so there's no second path that could pick it up.

Constructing-and-discarding to validate shape is a nice touch — it means a malformed `base_url` still fails at load for every command, which is the half that *should* stay eager. The split you've drawn (profile shape eager, credential presence deferred) is the right seam.

**One question rather than a finding:** `build_provider` is called with `UNRESOLVED_FALLBACK_KEY` as the credential. If any provider constructor validates key *shape* — a prefix check, a length check — a profile whose real key would be fine could fail at load with a confusing error about a key the operator never set. Does any current provider do that, and is it worth pinning with a test for a provider that has an opinion about key format? The distinctive placeholder string would also be worth grepping the log paths for, on the same reasoning.

Docs updated alongside, and the CHANGELOG entry explains the operator-visible behaviour change rather than just naming the fix — that's the part someone upgrading actually needs.

rafaelkenedy:
Thanks for checking both of those — and agreed on the chain guard being the one that carries the invariant; I've reworded the PR body to say so.

On the key-shape question: no current provider validates it at construction. `AnthropicProvider::new`, `OpenAiProvider::new`, `GeminiProvider::new`, and `OpenCodeProvider::new` build the `reqwest` client and store the `SecretString`; the key is only read when a request's auth header is built. `OpenAiCompatProvider::new` already relies on the same property — it substitutes `"dummy"` when no key is given. The native-credential providers (OpenAI OAuth, Codex, Copilot, Anthropic OAuth) never receive the resolved key at all: `fallback_provider_auth` routes them to their own sources.

Rather than leave that as an observation, b927083 pins it: `every_api_key_provider_accepts_the_unresolved_placeholder` loads a fallback for each API-key provider (`anthropic`, `openai`, `gemini`, `opencode`, `openai-compat`) with the variable absent and asserts that `load` succeeds, records the credential as unresolved, and discards the placeholder-built config. If a constructor ever grows a prefix or length check, that test fails rather than an operator's `status`.

On logs: `UNRESOLVED_FALLBACK_KEY` is referenced only in `load`, the config built from it is dropped there, and `SecretString`'s `Debug` redacts, so it has no route into a log line or an error message. The error an operator sees names the variable (`llm_fallbacks[0].api_key_env=... is set but the environment variable is missing or empty`), never a key.

PR #802, opened 2026-09-20
Add native ai-memory upgrade for GitHub release installs

### LINKED ISSUE #801: Native `ai-memory upgrade` for GitHub release installs
**Problem this would solve**
`ai-memory upgrade` exists only on the Docker shell wrapper (`bin/ai-memory`): refresh wrapper → `docker pull` → re-stage hooks. Native release / source installs have no equivalent CLI path. Docs tell operators to replace the binary by hand, then re-run `install-hooks --apply`.

macOS now recommends the native binary (`docs/macos.md`), so the happy path has no one-command upgrade. Thin launchers that `exec` a release binary under e.g. `~/.local/lib/ai-memory/` hit the same gap (`upgrade` is an unrecognized subcommand on the Rust CLI).

**Proposed solution**
Add an `upgrade` subcommand to `ai-memory-cli` for **GitHub-release native installs**:

1. Detect install kind (writable release binary under a user prefix vs package-managed / Docker wrapper).
2. Fetch the matching latest (or pinned) release asset for OS/arch from GitHub Releases; verify `.sha256`.
3. Atomically replace the on-disk binary (download → rename → re-exec).
4. Re-stage hooks for agents already under `~/.local/share/ai-memory/hooks/<agent>/` (same idempotent path as today's wrapper).
5. Stay client-only: never claim to upgrade a remote/homelab server.

Keep the existing Docker wrapper `upgrade` behavior (or detect wrapper and delegate). Refuse with a clear message for AUR/Homebrew/system paths (`use the package manager`).

**Alternatives considered**
- Document-only curl/tar recipe (works today; easy to get wrong; no sha verify / hook refresh).
- Extend only `bin/ai-memory` bash upgrade for native (does not help users whose PATH points at the Rust binary).
- `cargo install --force` / mise (not the release-binary path; toolchain required).

**Additional context**
- Current docs split: README / `docs/install.md` — Docker → `ai-memory upgrade`; native → manual binary + `install-hooks --apply`.
- Related: release assets already publish `ai-memory-macos-aarch64.tar.gz` (+ sha256) and peers.
- Semver: additive CLI subcommand → minor.

rubenlr:
## Summary

GitHub-release installs on macOS/Linux had no one-command upgrade. Docs told operators to replace the binary by hand, then re-run `install-hooks --apply`. Only the Docker wrapper exposed `ai-memory upgrade`.

This PR adds a native `upgrade` subcommand to the Rust CLI. It downloads the matching release tarball, verifies the `.sha256` sidecar, atomically replaces the on-disk binary (and a sibling `hooks/` tree when present), then refreshes staged hooks. Package-managed paths, containers, and Windows self-replace are refused with a clear next step. The Docker wrapper path is unchanged. This updates the local client only, not a remote server.

Optional `AI_MEMORY_RELEASE_BASE_URL` / `release_base_url` points Releases at a mirror or hermetic fixture. Downloads refuse bodies over 128 MiB.

Closes #801

---

## Added
- Native `ai-memory upgrade` for writable Unix GitHub-release installs (Linux/macOS), with `--version` pin and `--force` re-download
- Checksum verification of release `.sha256` sidecars before binary replace
- Automatic `install-hooks --apply` refresh for staged agents after a successful native upgrade
- Optional `release_base_url` / `AI_MEMORY_RELEASE_BASE_URL` override for mirrors and hermetic tests
- Hard 128 MiB download size cap (Content-Length and streamed)

## Changed
- Docs (`install.md`, `macos.md`, README) now document native vs Docker `upgrade` paths, the release-base trust boundary, and the download size limit


meiazero:
Thanks for this. I built the branch locally (Rust 1.95, Linux x86_64) with Claude's help and found a few blockers I'd like to help fix. Each has a repro; outputs below are from actual runs unless marked otherwise.

### 1. On Linux, `upgrade` always refuses: the running binary can't be opened for write

`unwritable_refusal` calls `is_writable_file(exe)`, which opens the current executable with `write(true)`. Linux returns `ETXTBSY` for a file that is being executed, so the check fails for every install, before anything is downloaded. The rename-based replace only needs the parent directory to be writable, so the file-level check can go.

Repro:

```bash
cargo build --bin ai-memory
mkdir -p /tmp/am-802 && cp target/debug/ai-memory /tmp/am-802/
stat -c '%U %a' /tmp/am-802/ai-memory     # <me> 775
/tmp/am-802/ai-memory upgrade --version v2.4.0 --force
# Error: refusing to self-upgrade: /tmp/am-802/ai-memory is not writable by this user; install under ~/.local (or another user-owned prefix) and re-run
```

Root cause in isolation:

```bash
cp /bin/sleep /tmp/sleepcopy && (/tmp/sleepcopy 5 &)
python3 -c "open('/tmp/sleepcopy','r+b')"
# OSError: [Errno 26] Text file busy: '/tmp/sleepcopy'
```

### 2. Real release tarballs fail validation

`release.yml:107` builds archives with `tar -C "dist/$artifact" -czf "$artifact.tar.gz" .`, so every entry is prefixed with `./`. `validate_release_entry` requires all components to be `Component::Normal`; `./` parses as `Component::CurDir`, so extraction aborts on the first entry. `extract_and_verify_round_trip` passes only because it builds the tar with a bare `ai-memory` path. (Masked end to end by #1 on Linux.)

The shipped v2.4.0 archive:

```bash
d=$(mktemp -d)
gh release download v2.4.0 -R akitaonrails/ai-memory -p 'ai-memory-linux-x86_64.tar.gz' -D "$d"
tar tzf "$d/ai-memory-linux-x86_64.tar.gz" | head -3
# ./
# ./ai-memory
# ./LICENSE
```

Failing test (drop into `upgrade.rs` tests):

```rust
#[test]
fn extracts_archive_built_like_release_yml() -> Result<()> {
    let src = tempfile::tempdir()?;
    fs::write(src.path().join(BINARY_NAME), b"bin")?;
    fs::create_dir_all(src.path().join("hooks/claude-code"))?;
    fs::write(src.path().join("hooks/claude-code/x.sh"), b"#!/bin/sh")?;
    let out = tempfile::tempdir()?;
    let archive = out.path().join("a.tar.gz");
    // Same invocation as .github/workflows/release.yml:107
    let status = std::process::Command::new("tar")
        .arg("-C").arg(src.path()).arg("-czf").arg(&archive).arg(".")
        .status()?;
    assert!(status.success());
    let dest = tempfile::tempdir()?;
    extract_release_archive(&fs::read(&archive)?, dest.path())?;
    assert!(dest.path().join(BINARY_NAME).is_file());
    Ok(())
}
// Error: release archive contains unsafe path: ./
```

### 3. Linuxbrew isn't treated as package-managed

`package_managed_prefixes()` has `/opt/homebrew/` and `/usr/local/Cellar/`, but not `/home/linuxbrew/.linuxbrew/`, and the tap ships Linux bottles (`on_linux` in `Formula/ai-memory.rb`). The Linuxbrew Cellar is user-writable, so once #1 is fixed the writability check won't catch it either and `upgrade` overwrites the brew-owned binary.

Failing test (note `classify_install` can't be used here: on a CI box the path doesn't exist, so it's refused as unwritable and the test passes for the wrong reason; `classify_rejects_homebrew_prefix` has the same blind spot):

```rust
#[test]
fn linuxbrew_is_package_managed() {
    let exe = Path::new("/home/linuxbrew/.linuxbrew/Cellar/ai-memory/2.4.0/bin/ai-memory");
    assert!(package_managed_refusal(exe).is_some());
}
// assertion failed: package_managed_refusal(exe).is_some()
```

### 4. Container detection only checks `/.dockerenv`

`container_refusal()` misses Podman (`/run/.containerenv`) and ignores `AI_MEMORY_IN_CONTAINER`. `ai_memory_wiki::backup::running_in_container()` covers all three and is already used by this PR in `staged_hooks_root`.

Repro (not run yet; use an image whose glibc matches the build host). Today #1 masks it — the refusal comes from the writability check, with the wrong message; once #1 is fixed, the guard passes and the download starts:

```bash
cargo build --bin ai-memory
mkdir -p /tmp/am-802-ctr && cp target/debug/ai-memory /tmp/am-802-ctr/
podman run --rm -v /tmp/am-802-ctr:/opt/am docker.io/library/ubuntu:24.04 \
  /opt/am/ai-memory upgrade --version v2.4.0 --force
# expected: "refusing to self-upgrade inside a container ..."
# today:    "refusing to self-upgrade: ... is not writable by this user" (from #1)
# after #1: "→ downloading ai-memory-linux-x86_64.tar.gz (v2.4.0)"
```

rubenlr:
Addressed in 3287daf7644ecaca3ef77b978989a274a54ac6ab.

1. **Writability / ETXTBSY** — `unwritable_refusal` now only requires the parent directory writable; dropped the running-exe `OpenOptions::write` probe that Linux rejects with ETXTBSY.
2. **`./` release tar paths** — normalize by skipping `Component::CurDir`, allow empty/`./` as the archive root directory, keep rejecting `..`, absolute, and unexpected paths. Added `extracts_archive_built_like_release_yml` (`tar -C … -czf … .`).
3. **Linuxbrew** — added `/home/linuxbrew/.linuxbrew/` to package-managed prefixes; regression via `package_managed_refusal` (not `classify_install`).
4. **Container detect** — `container_refusal` now delegates to `ai_memory_wiki::backup::running_in_container()` (`AI_MEMORY_IN_CONTAINER`, `/.dockerenv`, `/run/.containerenv`).

Commit: `3287daf7`

akitaonrails:
Audited (read-only). The download→verify→replace trust path is **sound**: sha256 is verified *before* any rename, the release base + owner/repo are pinned to GitHub, the 128 MiB cap is enforced on Content-Length and the streamed body, zip-slip/symlink entries are rejected, and it correctly refuses Windows self-replace / containers / package-managed prefixes; `release_base_url` goes through `Config`/figment (one config path). Two things before it can merge: (1) I retargeted the base **main → release/2.5** (a new subcommand is additive/minor). (2) **[SHOULD-FIX]** the `release_base_url` override accepts `http://` verbatim — since the `.sha256` sidecar is fetched from the same base, an on-path attacker against a non-loopback `http://` mirror controls both archive and checksum; please require `https://` unless the host is loopback. Optional: a hermetic full-flow test against a local `release_base_url` server. Left as your draft — ping when you consider it ready and I'll re-check + land it (closes #801).

rubenlr:
Windows-native `ai-memory upgrade` is in (zip extract + rename-aside self-replace). Please add the `windows` label so `.github/workflows/windows.yml` runs the native e2e on this PR.

rubenlr:
Thank you, @meiazero, I've incorporated the fixes and improved my review skills. I guess now it's all good

PR #824, opened 2026-09-21
fix(dev): preserve pre-push hooks across worktree installs

everton-dgn:
## What changed

The pre-push installer now works from a linked worktree, including a subdirectory. It asks Git for the shared hook path instead of assuming `.git` is a directory. Reinstallation replaces the managed block where it was, so user commands before and after it keep their execution order.

The installer stops without replacing the hook when `core.hooksPath` is configured, the managed markers are incomplete or duplicated, or the existing hook cannot be read. It recognizes CRLF markers and uses binary awk I/O to preserve the surrounding lines on Windows. Failed rewrites retain their temporary file and report its path.

## Why

Follow-up to #823. Its hook fixtures exposed two existing installer problems: `mktemp` fails under a linked worktree's `.git` file, and stripping the old block before appending a new one moves commands that originally followed it.

This draft depends on #823 for the Git-environment isolation and test fixtures. Please merge #822 and #823 first. The [incremental diff](https://github.com/everton-dgn/ai-memory/compare/feat/external-lifecycle-companion...fix/pre-push-installation) contains four files.

## Test plan

- [x] `git diff --check`
- [x] `cargo test --workspace --all-targets` through the pre-push hook: 3,395 passed, 0 failed, 11 ignored on macOS
- [x] `cargo deny check` and the audit command used by CI
- [x] Changelog section and frozen-release checks
- [x] Executed installer regression tests in isolated repositories. Both original bugs fail against the parent implementation and pass with this change. The linked-worktree test uses `git hook run pre-push`; the reinstallation test checks execution order and propagation of a failing test command. Additional cases cover configured hook paths, malformed markers, CRLF replacement and an unreadable existing hook.

- [x] [Upstream CI](https://github.com/akitaonrails/ai-memory/actions/runs/35571360568): all 17 jobs passed.
- [x] [Native Windows CI](https://github.com/everton-dgn/ai-memory/actions/runs/35571404647): workspace tests, relay checks and the real-server smoke test passed.

## Commit attribution

- [x] Verified the name and email on every commit in this PR.

## Release impact

- [x] Patch: installer bug fixes
- [ ] Minor
- [ ] Major

## Changelog

- [x] Added an `[Unreleased]` entry.
- [x] The changed installer behavior is described above.

## Notes for reviewers

`core.hooksPath` can point at a hook directory shared by unrelated repositories, or disable hooks. The installer therefore refuses an explicit override without changing that setting. Existing hook-manager users can integrate the managed block through their manager.

The generated test block is unchanged from #823. This follow-up only changes installation and its tests and documentation.


akitaonrails:
Audited (read-only). The incremental fix (`scripts/install-git-hooks.sh` using `git rev-parse --git-path hooks/pre-push` + in-place awk block replacement, with core.hooksPath / incomplete-marker guards and CRLF handling) is well-built and correct. It's not mergeable standalone yet because the diff carries the unmerged #822/#823 stack — note #822 already landed on `release/2.5`, so once #823 resolves (or the installer fix is un-stacked onto current main), the incremental 4-file change is ready. Leaving as your draft.

abhisheksharma2411 (commented):
Hi @everton-dgn — this is careful work. The `core.hooksPath` refusal is the right call (I ran into the other behaviour today from the opposite direction: `git rev-parse --git-path hooks` silently resolves to a configured shared hooks dir, so "stop and tell the user" is the only honest option), and the CRLF-aware `BINMODE=3` awk with `ENVIRON` for the block is a nicer solution than I'd have reached for.

I ran the shell rather than reading it. One finding, one thing I checked and want to confirm is fine, and a scope note.

### `set -euo pipefail` now leaks into the user's own hook lines

The managed block still opens with a bare `set -euo pipefail`, and that's the one line the in-place rewrite changes the meaning of. Before this PR the block was always stripped and re-appended at the **end**, so nothing of the user's ran after it. Now the block is replaced *where it sits* — which is the feature — so a user hook with commands after the block inherits `-e`, `-u` and `-o pipefail` they never opted into.

Reproduced with your block shape:

```
user command BEFORE the block
managed block ran
user command AFTER the block: about to reference an unset var
hook.sh: line 8: SOME_UNSET_VAR: unbound variable
hook exit = 1
```

`${SOME_UNSET_VAR}` was an empty string in their hook yesterday. Today it aborts the push, and the error names a line in a file they didn't edit. That's the specific promise the PR makes — "preserving surrounding user commands and their order" — and the order is preserved while the semantics aren't.

You already have the subshell, so the fix is to move the `set` inside it:

```sh
# >>> ai-memory pre-push >>>
(
    set -euo pipefail
    git_local_env_vars=$(git rev-parse --local-env-vars)
    ...
)
# <<< ai-memory pre-push <<<
```

Both properties hold afterwards — I checked both directions rather than just the happy one:

```
# user commands after the block
user command AFTER: value=<empty, as before>
user command AFTER still running
hook exit = 0

# a failing test inside the subshell
running tests
hook exit = 1        ← push still blocked
```

The second one matters: the subshell inherits `-e` from inside itself, exits non-zero, and the parent's own `set -e` (or just the subshell being the last statement) still fails the hook. Moving the `set` doesn't weaken the gate.

One knock-on: your comment *"A plain assignment, so `set -e` still aborts if `git rev-parse` fails"* stays true, since the `set -e` is then two lines above it instead of outside the parens.

### The `$?`-in-the-`else`-branch pattern is fine

This one looked wrong to me on a read and isn't, so I'm recording it so nobody "fixes" it later:

```sh
if git config --get core.hooksPath >/dev/null; then
    ...
else
    config_status=$?
```

`$?` in the `else` branch does hold the condition's status — confirmed, `git config --get` on a missing key gives `1`, and the `-ne 1` guard correctly distinguishes "not set" from a real failure. Same for the `grep -q` / `read_status` pair. Worth a short comment saying that's deliberate, because it's exactly the shape a drive-by cleanup deletes.

### Scope

+7483/-43 across the relay companion, two workflows, README, CONTRIBUTING and the e2e smoke test, for a PR titled "preserve pre-push hooks across worktree installs". I understand it's stacked on #823 and the hook fixtures came from there — but if #823 lands first, rebasing this down to the installer, its test, and the docs would make it a much easier review, and the installer logic here deserves attention that a 7k-line diff won't get.

Also worth knowing: #860 from @rubenlr touches the same installer with an overlapping worktree fix, plus a `GIT_*` scrub that's orthogonal to yours. Only one of the two lands as-is. I've flagged it there too — @akitaonrails may want to pick an order.

Happy to re-check any revision against the leak case; the reproduction is a six-line file.

everton-dgn:
Thanks @abhisheksharma2411, the leak was real, I reproduced it. The `SSL_CERT_FILE` export had the same problem on macOS, since it also sat outside the subshell.

Moving only the `set` leaves one case open, though: user commands after the block and no `set -e` in the user's own hook. The subshell failure is then just a status, the trailing commands run, and the hook exits with whatever ran last. `pre_push_hook_reinstall_keeps_user_content_and_one_managed_block` covers that case and fails with only the `set` moved: Cargo returns 29, the trailing command runs and the hook exits 0. Your two checks hold when the block is last or the user hook has its own `set -e`.

The shorter `( ... ) || exit $?` doesn't work either. Bash ignores `set -e` inside a compound command on the left of `||`, even with the `set -e` inside it, so a failing `rev-parse` would fall through to Cargo.

So the pushed version keeps the subshell as a plain statement and checks its status right after:

```sh
(
    set -euo pipefail
    # SSL_CERT_FILE export, Git scrub and cargo, unchanged
)
ai_memory_pre_push_status=$?
if [ "$ai_memory_pre_push_status" -ne 0 ]; then
    exit "$ai_memory_pre_push_status"
fi
unset ai_memory_pre_push_status
```

If the user's hook has `set -e`, the parent stops at `)` anyway, with the same status. The environment test now compares shell options, `pipefail` and `SSL_CERT_FILE` before and after the block, and a new test covers a failing `git rev-parse --local-env-vars` with user commands after the block. Each check fails against the old block, against moving only `set`, or against `|| exit $?`. The `$?` branches got the comment you suggested. I also replaced `IFS=$' \t\n'` with `unset IFS`: Bash 3.2 expands `$'...'` inside that heredoc, so the installed line came out different between Bash versions.

On scope, agreed. I took the "un-stacked onto current main" route @akitaonrails mentioned. The installer commits now sit on `main` without the relay, the changelog entries are under `[Unreleased]`, and the PR targets `main`. #823 keeps the relay.

This PR also carries the Git scrub from #823 (`1cf2d727`), so it overlaps #860 on both halves. I left a note there so we can settle the order.

Issue #758, opened 2026-09-18
Native Windows support


natansalvadorligabo:
**Problem this would solve**

Native Windows is currently marked as Experimental, which can become a significant adoption barrier in corporate environments.

Many companies standardize development machines on managed Windows devices. In those environments, developers may not be allowed to enable WSL2, use Docker Desktop, change system security policies, or install alternative environments.

This makes Native Windows support more than a convenience. For some teams, it is a prerequisite for adopting tools like `ai-memory`.

I'm currently evaluating `ai-memory` for broader use inside a software company, and this is one of the main constraints for wider adoption.

**Proposed solution**

Continue maturing Native Windows support toward becoming a first-class supported platform.

Ideally, this would include:

* reliable native installation and upgrades;
* stable Claude Code / Codex hooks on Windows;
* Windows coverage in CI where practical;
* clear documentation for managed/corporate Windows environments;
* guidance for environments using App Control for Business, Smart App Control, or similar security policies.

It would also be useful to document what is currently missing for Native Windows to move from **Experimental** to **Supported**.

**Alternatives considered**

* WSL2;
* Docker-based environments;
* changing local Windows security policies;
* using another development machine or environment.

These are reasonable alternatives for personal machines, but they are not always available on corporate-managed devices.

For company-wide adoption, requiring one of these alternatives can become a deployment and governance constraint.

**Would this break existing installs?**

No.

This request is about improving and stabilizing the existing Native Windows path, without changing current defaults or removing existing installation options.

**Additional context**

There have already been Native Windows-related fixes and discussions in the project, so I imagine this need has appeared before.

I'm opening this mainly to reinforce the corporate use case.

In many companies, Windows is not an optional development environment — it is the standard one. Improving Native Windows support could substantially reduce the barrier to adopting `ai-memory` across development teams.

rafaelkenedy:
Field report in the format `docs/windows.md` asks for, since the corporate case in this issue rests on native Windows actually holding up.

**Environment:** Windows 11 Pro 26200, `x86_64-pc-windows-msvc`, rustc 1.95.0, `main` @ `1fe32bc`. No WSL2, no Docker Desktop — the constraint this issue is about.

## The Rust side is already there

```
cargo test --workspace --all-targets
→ 3324 passed, 0 failed, 11 ignored (15 test binaries), exit 0
```

[...]

PR #830, opened 2026-09-21
fix(tests): convert MSYS paths before handing them to powershell.exe (#758)

rafaelkenedy:
Part of #758 (Native Windows support).

## The bug

`tests/hooks/test_lib.sh` **aborts on native Windows**: 49 of 67 assertions, exit 1.

```
$ sh tests/hooks/test_lib.sh
...
  ok  explicit project pin beats repo-root
. : O termo '/h/Workspace/ai-memory-pr/hooks/lib/ai-memory-hook.ps1' não é
reconhecido como nome de cmdlet, função, arquivo de script ou programa operável.
$ echo $?
1
```

The PowerShell parity block dot-sources `hooks/lib/ai-memory-hook.ps1` and passes `-Cwd`, giving both as `$PWD` / `$TMP` paths:

```sh
PS_REPO=$($PSH -NoProfile -ExecutionPolicy Bypass -Command \
    ". '$PWD/hooks/lib/ai-memory-hook.ps1'; Get-AiMemoryRepoRootProject -Cwd '$REPO/crates/cli'")
```

Under Git Bash those are MSYS paths (`/h/...`, `/tmp/...`). `powershell.exe` resolves neither, so the dot-source raises `CommandNotFoundException`, the command exits non-zero, and `set -e` kills the run at that line — the remaining `url_encode`, `json_field` and `post_hook` assertions never execute.

## Why it survived

The live branch is taken only when `pwsh` or `powershell` is on `PATH`. The `hooks-shell` legs in `ci.yml` are green, so whichever branch they take works on Linux. Native Windows is the one platform where `PATH` always has PowerShell *and* the paths handed to it are always unusable — and `windows.yml` runs `cargo test` and nothing else, so no job ever ran this suite there.

Which is the part worth fixing beyond the one line: the half of the suite written for Windows was the half no job ran on Windows.

## The change

- `host_path()` converts through `cygpath`, which ships with Git for Windows and exists nowhere else. Where it is absent the path is already native and the helper is a passthrough, so the Linux and macOS legs are byte-for-byte unchanged.
- `windows.yml` gains `sh tests/hooks/test_lib.sh`. Git Bash provides `sh` on `windows-latest`, and the suite is seconds against the ~1000s the Rust tests already cost in that workflow — nightly, dispatched, or `windows`-labelled, so it adds nothing to PR feedback time.

## Verification

Windows 11 Pro 26200, `x86_64-pc-windows-msvc`, rustc 1.95.0, `main` @ `1fe32bc`:

| | before | after |
|---|---|---|
| `sh tests/hooks/test_lib.sh` | 49/67, **exit 1** | **67 passed, 0 failed**, exit 0 |
| `powershell repo-root helper resolves repo basename` | never reached | ok |

`cargo test --workspace --all-targets` on the same box: **3324 passed, 0 failed, 11 ignored** — no Windows-only `#[ignore]`, no Rust-side failures. That result is in the #758 thread as a field report.

No CHANGELOG entry: test and CI only, which CONTRIBUTING exempts. No Rust touched, so `fmt`/`clippy` are unaffected; `git diff --check` is clean and the workflow YAML parses.

Please add the `windows` label so the new step runs before merge rather than after.


abhisheksharma2411 (commented):
The diagnosis and the one-line insight behind it — *the half of the suite written for Windows was the half no job ran on Windows* — are both right, and the fix is minimal in the good way. `cygpath` genuinely does ship with Git for Windows and genuinely doesn't exist elsewhere, so the helper is a true passthrough on Linux and macOS and those legs stay byte-for-byte identical.

I checked the thing that would have made this an incomplete fix: whether any **other** path reaches PowerShell unconverted. It doesn't. `test_lib.sh` has exactly one live `$PSH` invocation (L316) and both of its paths now go through `host_path`; every other `ai-memory-hook.ps1` reference in the file is a static `grep` against the script's text, which never leaves the shell. So the conversion set is complete rather than just sufficient for the one failing assertion.

`cygpath -w` in single quotes is also the right call — PowerShell single-quoted strings are literal, so the backslashes in `H:\...` need no escaping. (`-m` would work too and some people prefer it for exactly that reason, but `-w` is correct here and arguably more honest about what it's producing.)

**One thing I'd reconsider: where the step lives.**

The PR's own finding is that this suite never ran on the platform it was written for. With the step where it is, that stays *mostly* true:

```yaml
if: >-
  github.event_name != 'pull_request' ||
  contains(github.event.pull_request.labels.*.name, 'windows')
```

so on pull requests the seconds-long shell suite is gated behind an opt-in label and sits behind `cargo test --workspace --all-targets` — the ~1000s job — in the same `steps:` list. Two consequences:

1. A PR that touches `hooks/` or `tests/hooks/` without the label gets no Windows hook coverage, which is the exact gap this PR is closing. It's closed for nightly and for labelled PRs, not for the ordinary case.
2. Because it runs after `cargo test`, a Rust failure means the hook suite never executes at all — so on the runs where it *does* run, it's the least likely step to report.

Both are fixed by making it its own job rather than a step:

```yaml
  hooks:
    name: hook bundle (windows-latest)
    runs-on: windows-latest
    steps:
      - uses: actions/checkout@...
      - name: hook bundle (Git Bash)
        shell: bash
        run: sh tests/hooks/test_lib.sh
```

No Rust toolchain, no `rust-cache` restore (the workflow header notes that restore is ~10x slower on Windows over many small files), so it's a checkout plus a few seconds. That's cheap enough to run on every pull request without the label gate — which is what would actually make the guarantee "this suite runs on Windows" rather than "this suite runs on Windows when someone remembers". If you'd rather keep the PR-time cost at exactly zero, `paths: ['hooks/**', 'tests/hooks/**']` on that job gets most of the value.

I'd understand keeping it as-is too — it's consistent with the workflow's stated philosophy and you've clearly thought about the cost model. But the reasoning in that header is about the ~1000s Rust job, and this step doesn't share those economics.

**On the verification:** 49/67 → 67/67 with exit 0, plus a clean `cargo test --workspace --all-targets` on the same box, is more than enough, and citing the field report in #758 rather than asserting it is appreciated. Agreed on no CHANGELOG entry.

Adding the `windows` label before merge is the right call so this lands proven rather than hoped.

rafaelkenedy:
Agreed, and done in a840556 — the suite is its own `hooks` job now, outside the label gate.

Your framing is what settled it: as a step it inherited `test`'s `if:`, so the gap this PR is about stayed open for exactly the case that matters, a PR touching `hooks/` that nobody labelled. "Runs on Windows when someone remembers" is not the guarantee I was claiming in the PR body. And the ordering point is the one I'd missed entirely — behind ~1000s of `cargo test`, a Rust failure means the shell suite never reports at all, so the steps were coupled in a direction with no upside.

The workflow header and the AGENTS.md bullet now both name `hooks` as the exception that runs on every pull request, since both described a workflow that was opt-in in full.

**One correction on the fallback you offered.** `paths:` is not a job-level key in GitHub Actions — it is a filter under `on.<event>`, so `paths: ['hooks/**', 'tests/hooks/**']` can only be attached to the workflow's `pull_request` trigger, where it would also gate the Rust job. Per-job path filtering needs `dorny/paths-filter` or a separate workflow file. That makes the cheap version of "most of the value" not actually available, which I think strengthens your primary suggestion rather than weakening it: an unfiltered job that costs a checkout and seconds is simpler than either workaround and gives the stronger guarantee.

Two smaller confirmations:

- Same completeness result on the conversion set. `grep -n 'ai-memory-hook.ps1' tests/hooks/test_lib.sh` is one live `$PSH` invocation and five static `grep`s against the script's text; only the former leaves the shell, and both of its paths go through `host_path`.
- On `-w` vs `-m`: agreed both work here. I kept `-w` because the value is only ever interpolated into a PowerShell single-quoted string, where backslashes are literal and nothing has to tolerate them — `-m` would be hedging against an escaping problem this call site does not have.

`sh tests/hooks/test_lib.sh` still 67 passed / 0 failed on the same box after the restructure. The `windows` label is still what would prove the new job in CI rather than in my terminal.

rafaelkenedy:
The `hooks` job you asked for caught a second Windows-only bug on its first run, which I think settles the question of whether it earned its place.

```
tests/hooks/test_lib.sh: line 317: /c/Program: No such file or directory
##[error]Process completed with exit code 127
```

`$PSH` was expanded unquoted — pre-existing, not something the first commit introduced — so an interpreter path with a space word-splits. Fixed in 2cd1884 by quoting it.

What makes this a good demonstration of your point is *why* neither of us hit it earlier. The block prefers `pwsh` and falls back to `powershell`. On a Linux runner `pwsh` is `/usr/bin/pwsh`; on my machine `pwsh` is not installed at all, so it took the fallback to the System32 path. Neither has a space. `windows-latest` ships PowerShell 7 under `C:\Program Files`, so it is the first host where the *preferred* branch resolves to a path that cannot survive word splitting — and the label gate meant nothing would have run there until someone labelled a PR.

So the failure mode was: a bug reachable only on Windows, in the branch Windows is the most likely host to take, in a suite no Windows job ran. My local run was green for a reason that had nothing to do with the code being correct.

I reproduced it locally before fixing rather than fixing blind — a `pwsh` shim at a spaced path, put ahead of `PATH` so the script's own `command -v pwsh` resolves to it:

| | exit | result |
|---|---|---|
| before | 127 | dies at L317, same error shape as the runner |
| after | 0 | 67 passed, 0 failed |

Both Windows fixes are now one-liners in the same block, which is a fair description of how much was actually wrong: the suite had simply never been executed on the platform half of it was written for.

abhisheksharma2411:
You're right about `paths:` and I was wrong — it's a filter under `on.<event>`, not a job-level key, so what I offered as the cheap fallback isn't expressible that way at all. Attaching it to the `pull_request` trigger would have gated the Rust job too, which is the opposite of what I was suggesting. Thanks for correcting it rather than quietly working around it; per-job path filtering needing `dorny/paths-filter` or a second workflow file is exactly the kind of thing I'd have repeated elsewhere.

And the first run catching `$PSH` unquoted is a better argument for the job than anything either of us wrote in the thread:

```
tests/hooks/test_lib.sh: line 317: /c/Program: No such file or directory
```

The part I find genuinely instructive is *why* it stayed hidden. The block prefers `pwsh` and falls back to `powershell`; on Linux `pwsh` is `/usr/bin/pwsh`, and on your machine it fell back to System32 — neither has a space. `windows-latest` ships PowerShell 7 under `C:\Program Files`, so it is the first host where the **preferred** branch resolves to a path that word-splits. A bug reachable only on Windows, in the branch Windows is most likely to take, in a suite no Windows job ran. My own "verified locally, green" would have been green for a reason unrelated to correctness.

That is the same shape as the defect the PR fixes, one level up, and it's a cleaner demonstration of the gap than the original MSYS-path bug was.

Nice touch reproducing it with a `pwsh` shim at a spaced path ahead of `PATH` rather than fixing blind — that makes the quoting fix a verified one rather than a plausible one.

Nothing further from me. Congrats on the merge.

